PR #4, opened 2017-05-22
th/wait-nm: implement waiting for activation/deactivation

thom311:
When activating/deactivation a connection with NetworkManager, implement the 'wait' option to wait for the operation to complete. This is what `nmcli connection up|down` does, depending on the `--wait` option.


thom311:
merged to master as abb8945079c0e2c4979d6bb4d2031cd408c82fd6

Issue #5, opened 2017-06-13
Use proper reserved IP and MAC addresses in documentation


pcahyna:
README.md currently uses the d6:06:b9:56:12:5d MAC address in an example. examples/inventory use the 52:54:00:44:9f:ba and 52:54:00:05:f5:b3 MAC addresses.
The examples under examples/ use the 192.168.174. network prefix.
Is there a special reason why those values were chosen? If not, it might be better to use the addresses which are reserved for use in documentation by IANA. For IP adresses those are the 192.0.2.0/24,  198.51.100.0/24 and 203.0.113.0/24 blocks (RFC 5735). For MAC adresses, this is the 00-00-5E-00-53-00 through 00-00-5E-00-53-FF block (RFC 7042).
Note that the firewall role uses example addresses as well, it would be good to agree on a common standard.

thom311:
I think for IPv4 addresses, it is a common scenario to configure private addresses from 192.168.0.0/16 (etc). Note that these private addresses are also mentioned by RFC 5735. They are not globally routed and local only. As such, they seem suitable to use in the example and more familar.

For MAC addresses, I agree.

pcahyna:
I still think that using the proper address range reserved for examples in examples has some advantages. Private addresses are not globally routed, sure, but one uses them very often. There is therefore some risk of conflict one forgets to edit the example playbook and run it unmodified. It also makes possible to determine whether the example has been modified to list the correct addresses or not. If one sees 192.168.174. in an inventory/playbook, there is no reliable way to determine whether it is the actual intended address or an example placeholder in need to be updated. If one uses the proper address reserved for examples, it becomes clear that it needs to be changed.
By the way, the firewall role uses 1.2.3.4 as an example in documentation. This does not seem right (as it is a globally routed address), but has the advantage of being even more obvious that it is an example which needs to be changed.

pcahyna:
I forgot that there is one IPv6 address used in the documentation as well. I propose to change it to the 2001:DB8: prefix (RFC 3849).

thom311:
Should be fixed now with commit 797dd44ef51055429d62cfbd1756cab8c892107c and commit 3225b3a89d67afcbab502f1bf5ac52f9786d8b4b .

Please reopen, if something is missing.

Thank you

PR #6, opened 2017-06-23
Use proper reserved MAC addresses for documentation, according to RFC 7042

pcahyna:
See issue #5 .


thom311:
Thanks.

Merged: https://github.com/linux-system-roles/network/commit/797dd44ef51055429d62cfbd1756cab8c892107c

PR #12, opened 2017-08-28
Replace yet another RFC1918 address in an example by a RFC5737 one.

pcahyna:


thom311:
merged as https://github.com/linux-system-roles/network/commit/a4ebfadded1c4e96e3a7b5e71f7cfc2da46ee1d6

PR #14, opened 2017-09-21
role: improve making "network_provider" configurable via host vars

thom311:
The role supports two providers: "nm" and "initscripts".

The provider is autodetected by loading one of the vars/*.yml files
(where the default is loaded as internal "network_provider_default" variable).
The user can still overwrite the provider, by explicitly setting the
"network_provider" variable.

Depending on the provider there is the list of packages that shall be
installed and the service to start. Selecting this was broken before.

This is now fixed and works like following:

The variables "network_service_name" and "network_packages" can be
specified by the user as host variables. But usually one wouldn't
want to do that. In that case, we look into the internal
"_network_provider_setup" dictionary, which uses the variables
"network_service_name_nm", "network_service_name_initscripts",
"network_packages_nm", "network_packages_initscripts".

These variables are initialized in "defaults/main.yml" as well,
but they could be overwritten via "vars/*.yml" files.

https://bugzilla.redhat.com/show_bug.cgi?id=1485074

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,2 +1,35 @@
-network_provider: "{{ network_provider_default }}"
-network_connections: []
+network_provider:     "{{ network_provider_default }}"
+network_connections:  []
+
+network_service_name_default_nm: NetworkManager
+network_packages_default_nm: [ NetworkManager ]
+
+network_service_name_default_initscripts: network
+network_packages_default_initscripts: []
+
+
+# The user can set host variables "network_provider",
+# "network_service_name" and "network_packages".
+#
+# Usually, he only wants to select the "network_provider"
+# (or not set it at all and let it be autodetected via the
+# internal variable "{{ network_provider_default }}". Hence,
+# depending on the "network_provider", a different set of
+# service-name and packages is used.
+#
+# That is done via the internal "_network_provider_setup" dictionary.
+# If the user doesn't explicitly set "network_service_name" or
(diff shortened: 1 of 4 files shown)


defaults/main.yml
```diff
@@ -1,2 +1,35 @@
...
+# That is done via the internal "_network_provider_setup" dictionary.
+# If the user doesn't explicitly set "network_service_name" or
+# "network_packages" (which he usually wouldn't), then the defaults
+# from "network_service_name_default_*" and "network_packages_default_*"
+# apply. These values are hard-coded here too, but they could also
+# be overwritten as hot variables or in vars/*.yml.
+_network_provider_setup:
+    nm:
+        service_name: "{{ network_service_name_default_nm }}"
+        packages:     "{{ network_packages_default_nm }}"
```
pcahyna: I think there is too much indirection here. Why not put simply NetworkManager here instead of introducing a new variable `network_service_name_default_nm` ? If one wants to customize service/packages, one would anyway set `network_packages` or `network_service_name`, not these `*_default_*` variables.

thom311: It might make sense to overwrite them.

On debian there could be a file `vars/Debian 7.yml` with

```
network_provider_default: nm
network_packages_default_nm: [ 'network-manager' ]
```

Currently the default for these variables is hard-coded in `default/main.yml`.

Issue #15, opened 2017-09-24
RFE: Allow user-defined attributes in network_connections structure


forsetti:
The network_connections struct is a good general data model for network connections, and could be used by more than just the linux-system-roles.network role.  The value of this struct would be greatly increased if other attributes could be added, beyond what is useful by this role.  However, adding user-defined attributes to network_connections that are not recognized by this role generates a failure.

Please consider tolerating user-defined attributes within the network_connections struct, without generating a failure in the linux-system-roles.network role

Example below.  Note addition of non-standard "libvirt_dev", for use by another role in the same playbook.  I use this to create a VM by generating a domain.xml with matching network devices which I supply to the libvirt module.  Then when the VM is running, I could apply the linux-system-roles.network role to the guest.  However, this does not work, because linux-system-roles.network generates a failure because of my addition "libvirt_dev" attribute.

host_vars/foo.example.com/main.yml:
---
network_connections:
- name: "eth0"
  type: "ethernet"
  mac: "12:34:56:78:9a:bc"
  ip:
    dhcp4: yes
  libvirt_dev: eno1.2
---

Playbook execution:
---
TASK [linux-system-roles.network : Configure networking connection profiles] ***************************************************************************************************
fatal: [foo.example.com]: FAILED! => {"changed": false, "failed": true, "msg": "configuration error: connections[0]: invalid key \"libvirt_dev\""}
---

If it is necessary to maintain a level of restrictiveness to the schema, I'd request either creating a new "network_user_attrs" data element which contains a list of user-defined attributes to be allowed within the struct, or use of a prefix such as "user_" for user-defined attributes.

thom311:
Yes, that would be possible to do... not sure though, if it's the right thing to do... :)

forsetti:
Agreed - unsure if it's the "right" thing.  Having a unified data model for
network info is a (very) nice-to-have; but if you see it interfering with
goals of this role, then I'd close this WONTFIX.

thom311:
I think the patch itself is trivial. Basically, it reserves the name "user_attrs" for external use. Only minor downside is the added code, and that this particular name is now taken.

I am just not sure anybody would use it, or that it looks nice for the user to make use of this...

Would this patch solve your use-case nicely?

forsetti:
Likely, yes - with your user_attrs approach, using my initial example, how
would the network_connections dict look to support the libbvirt_dev
attribute? And, for playing out this example, lets add another foo_bar
[...]

PR #16, opened 2017-09-25
library: support a network_connections[*].user_attrs attribute for ar…

thom311:
…bitrary data

The role/library performs a strict validation of the input data in
"network_connections". Support a new dictionary value "user_attrs"
that is ignored by the role. It allows the user to piggyback additional
settings in "network_connections" that will be ignored by the role.

https://github.com/linux-system-roles/network/issues/15

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -558,13 +558,14 @@ class ArgValidatorIPAddr(ArgValidatorStr):
             raise ValidationError(name, 'value "%s" is not a valid IP%s address with prefix length' % (value, Util.addr_family_to_v(self.family)))
 
 class ArgValidatorDict(ArgValidator):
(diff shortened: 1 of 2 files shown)


larskarlitski:
The new tests on this are failing because of the way the role is imported in one of the playbooks. I have a fix for that in #18.

thom311:
Ok, this might not be a good idea. Let's close this PR, at least for now...

thom311:
[0001-library-support-a-network_connections-.user_attrs-at.patch](https://github.com/linux-system-roles/network/files/1446562/0001-library-support-a-network_connections-.user_attrs-at.patch.txt)

Applies on top of commit 16ad34c20a54210a40a5a508b82218845ddd28da

PR #20, opened 2017-11-13
Fix exception when dns is present

roolebo:
ifcfg_create expects list of dns addresses to be non plain values, while
dns validator returns only a string.


thom311:
Merged as 9b8a46e3e30386591d57edce2bcc04a417d817ac
Thanks!!

PR #21, opened 2017-11-17
support for configuring static routes (th/static routes)

thom311:
Currently, only "dest/prefix,gateway,metric" are supported.

There is an option `route_append_only`, to only extend the pre-existing configuration by adding further routes. I think this setting is wrong in principle, because you should better define the entire configuration in the playbook, not just parts and not extend on the pre-existing configuration. It makes the outcome of the run dependent on pre-existing configuration. But it is here on request.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -517,46 +517,6 @@ class ArgValidatorBool(ArgValidator):
             pass
         raise ValidationError(name, 'must be an boolean but is "%s"' % (value))
 
-class ArgValidatorIP(ArgValidatorStr):
-    def __init__(self, name, family = None, required = False, default_value = None, plain_address = True):
-        ArgValidatorStr.__init__(self, name, required, default_value, None)
-        self.family = family
-        self.plain_address = plain_address
-    def _validate(self, value, name):
-        v = ArgValidatorStr._validate(self, value, name)
-        try:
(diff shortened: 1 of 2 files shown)


bengal:
Would it be possible to have a more generic way to keep some properties untouched? For instance, if I want to update the route metric without changing addresses and routes I could do something like:
```
 network_connections:
  - name: "Wired0"
    type: "ethernet"
    interface_name: "eth0"
    ip:
       route_metric4: 100
       address: keep
       routes: keep
```

I don't know ansible well and if there is a syntax that would allow that, though.

thom311:
```
       routes: keep
```

I like that, but it cannot express to amend the current routes, by adding the one from the list. I think that is kinda required. Do you have a suggestion for such a syntax?

bengal:
> routes: keep
> 
> I like that, but it cannot express to amend the current routes, by adding the one from the list. I think that is kinda required. Do you have a suggestion for such a syntax?

The only idea I have right now is that for every composite property X there should be also a X_append and X_remove property (handled generically) to add and remove elements from the list (similarly to the [+,-]ip4v.address= syntax in nmcli. I don't know if this is feasible and how well it integrates in the ansible model, as the playbook should only describe the final state.

thom311:
Merged. After discussion, the current form seems suitable.

Issue #9, opened 2017-08-04
Not all network_provider: initscripts options are supported, some crucial for enterprise.


hartsjc:
Many of the /etc/sysconfig/network-scripts/ifcfg-<interface-name> options documented in  /usr/share/doc/initscripts-*/sysconfig.txt  do not appear to be supported.  Some key/crucial ones that appear to be missing for enterprise servers are:
```
1. ETHTOOL_OPTS
2. LINKDELAY
3. IPV4_FAILURE_FATAL
```

Some others may include (have not tested to validate missing, just grep of code):
```
ARPCHECK
DHCLIENTARGS
DHCLIENT_IGNORE_GATEWAY
DHCP_FQDN
DHCP_HOSTNAME
DHCPRELEASE
HOTPLUG
MACADDR
NETMASK
NO_DHCP_HOSTNAME
NOZEROCONF
PEERDNS
PERSISTENT_DHCLIENT
SCOPE
SRCADDR
USERCTL
WINDOW
ZONE
```

thom311:
We probably won't ever support all options that either provider (initscripts or nm) supports.

For one, the provider (backends) may not support all options that another provider implement. It is possible, and sometimes reasonable to add a feature, even if it is not supported with every provider. However, unless there are good justifications to do it, we may not do that.

Also, some features may not be common enough to justify the work to add them.

I am closing this issue. We should open separate issues for each feature/option. That way, the issues themselves have a smaller and better defined scope, instead of being open-ended. 

Thank you.

PR #22, opened 2017-12-07
Add firewalld "zone" variable for connections

dsedivec:
Note that this is one of the variables requested by #9 (`ZONE`).

I ran into this because using both this network role in initscripts mode together with e.g.

```
- firewalld: zone=external interface=eth0 state=present permanent=yes
```

would cause Ansible to report changes on every run of my playbook.  By just having this role set the firewalld zone in the `ifcfg-eth0` file, I can skip using the firewalld module at all, and no more spurious changes as this role and the firewalld module fight over the `ifcfg-eth0` file.

I did not test this on a system whose interfaces are actually managed by NetworkManager, though I did run the unit tests on a system with the NM Python APIs installed, and all passed.  Though I'm not entirely sure that I have updated the unit tests to everyone's liking.


thom311:
Nice. Thanks!!

I merged it, with a follow-up fix.

I think the validation of the user configuration should be strict, hence, 'zone' should be rejected for slaves.

PR #23, opened 2018-01-19
th/ansible_independent

thom311:
Refactor the module, to make it mostly independent of the ansible runtime.

The module is just one large python script. Ansible will execute it and communicated with it via stdout/stdin.

However, most of the stuff that the script does, is independent of ansible or how it is called.

Refactor the code, so that the parts that require the ansible environment can be replaced. This way, one could import the code in another python application, or we could run it from the command line and feed our own input format.


bengal:
Looks good to me.

PR #26, opened 2018-01-24
th/nest_vlan_infiniband

thom311:
Deprecate the properties `vlan_id`, `infiniband_p_key` and `infiniband_transport_mode` and instead nest them under `vlan` / `infiniband` settings.

It follows what is done for `ip`, `ethernet` and `bond`. It also scales better as we are going to add more properties. It allows to do validation in the `ArgValidator_Dict*` instances, instead everything in `ArgValidator_DictConnection._validate_post()`.

The old values are still supported for backward compatibility.


bengal:
LGTM

PR #36, opened 2018-03-09
Improved documentation, especially of the new options added in 0.4, and DNS.

pcahyna:
Please review and add anything if needed, e.g. it might be good to document to what NetworkManager/initscripts settings are the newly documented options translated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ variable.
 Valid values for `type` are:
 
   - `ethernet`
+  - `infiniband`
   - `bridge`
   - `bond`
   - `team`
--- a/README.md
+++ b/README.md
@@ -317,6 +317,25 @@ be deleted by the role.
 
 Slaves to bridge/bond/team devices cannot specify `ip` settings.
 
+### `type: ethernet`
+
+Ethernet-specific options can be set using the connection profile variable `ethernet`. This
+variable should be specified as a dictionary with the following items (options): `autoneg`, `speed` and `duplex`,
+which correspond to the settings of the `ethtool` utility with the same name. `speed` is an
+integer giving the speed in Mb/s, the valid values of `duplex` are `half` and `full` and
+`autoneg` accepts a boolean value.
+
+```yaml
+network_connections:
+  - name: "eth0"
+    type: "ethernet"
+
+    ethernet:
+      autoneg: false
+      speed: 1000
+      duplex: full
+```
+
 ### Virtual types and Slaves
 
 Device types like `bridge`, `bond`, `team` work similar:
--- a/README.md
+++ b/README.md
@@ -253,6 +253,13 @@ network_connections:
       #dhcp4_send_hostname: no
       gateway4: 192.0.2.1
 
+      dns:
+        - 192.0.2.2
+        - 198.51.100.5
+      dns_search:
+        - example.com
+        - subdomain.example.com
+
       route_metric6: -1
       auto6: no
       gateway6: 2001:db8::1
@@ -273,6 +280,9 @@ the DHCPv4 request includes the hostname via `dhcp4_send_hostname`.
 Note that `dhcp4_send_hostname` is only supported by the `nm` provider and translates
 to [`ipv4.dhcp-send-hostname`](https://developer.gnome.org/NetworkManager/stable/nm-settings.html#nm-settings.property.ipv4.dhcp-send-hostname)
 property.
+
+Manual DNS configuration can be specified via a list of addresses
+`dns` and a list of domains to search `dns_search`.
 
 - For NetworkManager, `route_metric4` and `route_metric6` corresponds to the
 [`ipv4.route-metric`](https://developer.gnome.org/NetworkManager/stable/nm-settings.html#nm-settings.property.ipv4.route-metric) and
--- a/README.md
+++ b/README.md
@@ -295,7 +295,8 @@ to [`ipv4.dhcp-send-hostname`](https://developer.gnome.org/NetworkManager/stable
 property.
 
 Manual DNS configuration can be specified via a list of addresses
-`dns` and a list of domains to search `dns_search`.
+given in the `dns` option and a list of domains to search given in the
+`dns_search` option.
 
 - For NetworkManager, `route_metric4` and `route_metric6` corresponds to the
 [`ipv4.route-metric`](https://developer.gnome.org/NetworkManager/stable/nm-settings.html#nm-settings.property.ipv4.route-metric) and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -290,6 +302,19 @@ Manual DNS configuration can be specified via a list of addresses
  properties, respectively. If specified, it determines the route metric
 for DHCP assigned routes and the default route, and thus the priority for multiple interfaces.
 
+Static route configuration can be specified via a list of routes given in the `route`
+option. The default value is an empty list. Each route is a dictionary with the following
+entries: `network`, `prefix`, `gateway` and `metric`. `network` and `prefix` together specify
+the destination network. CIDR notation or network mask notation are not supported yet. If the
+boolean option `route_append_only` is `True`, the specified routes are appended to the
+existing routes, if it is `False` (default), the current routes are replaced. Setting this
+option without setting `route` has the effect of preserving the current static routes. The
```
thom311: "Setting this option without...." this sentence is not clear to me. In fact, if you do not specify "route" nor "route_append_only", the default is
```
    - route
    - route_append_only: false
```
in this case, it will configure no routes (or remove all existing routes on update).

Maybe it should say: "Setting this option to `true` without setting `route` has the effect..."

README.md
```diff
@@ -290,6 +302,19 @@ Manual DNS configuration can be specified via a list of addresses
...
+boolean option `rule_append_only` works in a similar way for routing rules. Note that there is
+no further support for routing rules at the moment, so this option serves merely the purpose
+of preserving the current routing rules.  Note also that when
+`route_append_only`/`rule_append_only` is not specified, the current routes/routing rules will
+be deleted by the role.
```
thom311: Yes, the later line clarifies on that... but it's a bit duplicated, isn't it?

README.md
```diff
@@ -114,6 +114,7 @@ variable.
 Valid values for `type` are:
 
   - `ethernet`
+  - `infiniband`
```
thom311: The subject of the commit is wrong: "infiniband is a valid connection <<TYPE>> as well".

`git commit --amend ...`

Maybe add the tag "doc: " as a prefix for each subject line of all these commits?

README.md
pcahyna: That was the intent, by "setting the option" I meant "to true", setting it to False would be actually unsetting...

README.md
pcahyna: It is, but I find the current behavior (remove all existing routes) dangerous, so a bit of redundancy does not harm.

README.md
pcahyna: you mean "infiniband is a valid connection type as well"?

thom311: Exactly: `connection "type"`. I used \< which caused a loss in markdown.

README.md
thom311: "setting the option" means to me just specifying it (instead of omitting it). Can we say "setting the option to `true`"? That's all I really had to say, in summary.

README.md
pcahyna: I rebased the branch with adding the tag as you suggest.

README.md
pcahyna: I made the change that you suggest for clarity.

thom311:
lgtm

README.md
```diff
@@ -274,6 +281,9 @@ Note that `dhcp4_send_hostname` is only supported by the `nm` provider and trans
 to [`ipv4.dhcp-send-hostname`](https://developer.gnome.org/NetworkManager/stable/nm-settings.html#nm-settings.property.ipv4.dhcp-send-hostname)
 property.
 
+Manual DNS configuration can be specified via a list of addresses
```
tyll: How about:

    The `dns` keyword accepts a list of IP addresses of DNS servers for manual configuration with the list of domains to search after the `dns_search` keyword.

Another possibility would be to add parenthesis to the key words, otherwise it appears to me that the sentence is not correct grammar-wise:

    Manual DNS configuration can be specified via a list of addresses (`dns`) and a list of domains to search (`dns_search`).

README.md
```diff
@@ -317,6 +317,25 @@ be deleted by the role.
 
 Slaves to bridge/bond/team devices cannot specify `ip` settings.
 
+### `type: ethernet`
+
+Ethernet-specific options can be set using the connection profile variable `ethernet`. This
+variable should be specified as a dictionary with the following items (options): `autoneg`, `speed` and `duplex`,
+which correspond to the settings of the `ethtool` utility with the same name. `speed` is an
+integer giving the speed in Mb/s, the valid values of `duplex` are `half` and `full` and
```
tyll: As far as I understand, there needs to be a comma before the `and`:
https://english.stackexchange.com/questions/30516/should-i-use-a-comma-before-and-or-or

README.md
```diff
@@ -317,6 +317,25 @@ be deleted by the role.
...
+`autoneg` accepts a boolean value.
```
tyll: maybe finish the sentence with

    (default is `yes`) to configure autonegotiation.

By adding the full name it is easier to find the option when searching for it. Also please add:

    The `speed` and `duplex` settings are required when autonegotiation is disabled.

README.md
```diff
@@ -317,6 +317,25 @@ be deleted by the role.
...
+
+```yaml
+network_connections:
+  - name: "eth0"
+    type: "ethernet"
+
+    ethernet:
+      autoneg: false
```
tyll: Most examples in the README use `yes` or `no`, therefore I would prefer `no` here, too.

README.md
```diff
@@ -290,6 +302,19 @@ Manual DNS configuration can be specified via a list of addresses
  properties, respectively. If specified, it determines the route metric
 for DHCP assigned routes and the default route, and thus the priority for multiple interfaces.
 
+Static route configuration can be specified via a list of routes given in the `route`
+option. The default value is an empty list. Each route is a dictionary with the following
+entries: `network`, `prefix`, `gateway` and `metric`. `network` and `prefix` together specify
+the destination network. CIDR notation or network mask notation are not supported yet. If the
+boolean option `route_append_only` is `True`, the specified routes are appended to the
```
tyll: The example uses `yes` instead of `True`, so maybe use

    boolean option `route_append_only` is `yes`

README.md
```diff
@@ -290,6 +302,19 @@ Manual DNS configuration can be specified via a list of addresses
...
+existing routes, if it is `False` (default), the current routes are replaced. Setting this
```
tyll: and
    
    existing routes, if it is `no` (default), the current routes are replaced. Setting this

README.md
```diff
@@ -290,6 +302,19 @@ Manual DNS configuration can be specified via a list of addresses
...
+option to `True` without setting `route` has the effect of preserving the current static routes. The
```
tyll: and
    option to `no` without setting `route` has the effect of preserving the current static routes. The

README.md
pcahyna: I would not call it a `keyword`, this would be the first time this term occurs in this document, and would therefore be inconsistent. I think the keys at this level are called `options` or just `keys`.

tyll: Yes, good catch.

pcahyna: I went for

Manual DNS configuration can be specified via a list of addresses
given in the `dns` option and a list of domains to search given in the
`dns_search` option.

But the original version was copied directly from the preexisting text about `address`. So that one is a bit ungrammatical as well.

README.md
pcahyna: I believe you meant
option to `yes` without setting route has the effect of preserving the current static routes. 
`no` makes no sense here IMHO.

tyll: yes, good catch!

PR #47, opened 2018-05-14
DO NOT MERGE - TEST ONLY remove an apparently superfluous python path setting.

pcahyna:


tyll:
Hi, FYI: #46 moves the file in the ```/test``` directory and changes this line. What do you want to test here? You can probably do this also manually.

pcahyna:
I want to check my understanding that this line really was superfluous, as I mentioned in a comment to #46

tyll:
A similar line will be needed again when the file moved to ```/test```. Did you maybe mean a comment in #45? I do not see a comment from you in #46.

pcahyna:
See https://github.com/linux-system-roles/network/pull/46/files/4c01128e73bac316f336d18585e3aa6f0ee95dba#diff-ba89226e1b53a3f2c52658ca722fc692
Yes, a similar line will be needed, but this one was not needed, IMO.

PR #50, opened 2018-05-18
More integration tests

tyll:
Add more integration tests and scripts to gather coverage data from integration tests

### DIFF
--- a/tests/get-coverage.yml
+++ b/tests/get-coverage.yml
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+# This expects the variable test_playbook to be set from the outside
+- name: Prepare for coverage extraction
+  hosts: all
+  tasks:
+    # Use set_fact to set variables to make them available in all plays
+    # 'vars:' Would only set variables for the current play
+    - name: set facts
+      set_fact:
+        coverage_module: network_connections
+        coverage: /root/.local/bin/coverage
+        destdir: "remote_coverage/{{ test_playbook }}"
+
+    # This uses variables from the other set_fact task, therefore it needs to
+    # be its own task
+    - name: set more facts
+      set_fact:
+        coverage_file: ansible-coverage-{{ coverage_module }}-{{ test_playbook|replace('.yml', '') }}
+
+    - name: debug info
+      debug:
+        msg: Getting coverage for '{{ coverage_module }}' with '{{ test_playbook }}'
+
+    # combine data in case old data is left there
+    - command: "{{ coverage }} combine"
+      environment:
+        COVERAGE_FILE: "{{ coverage_file }}"
+      ignore_errors: yes
+
+    - name: remove old data
+      file:
+        state: absent
+        path: "{{ coverage_file }}"
+
+    - name: remove old data
+      shell: rm -f .coverage.*
+
+    - name: copy coveragerc
+      copy:
+        content: "[run]\ndisable_warnings = no-data-collected\n"
+        dest: .coveragerc
+
+    - name: install latest pip
+      pip:
+        name: coverage
+        extra_args: --user --upgrade
+
+- import_playbook: "{{ test_playbook }}"
+  vars:
+    ansible_python_interpreter: "{{ coverage }} run -p --include *ansible_module_{{ coverage_module }}.py"
+
+- name: Gather coverage data
+  hosts: all
+  tasks:
+    - shell: "{{ coverage }} combine .coverage.*"
+      environment:
+        COVERAGE_FILE: "{{ coverage_file }}"
+
+- name: Get coverage data
+  hosts: all
+  tasks:
+    - fetch:
+        src: "{{ coverage_file }}"
+        dest: "{{ destdir }}"
+        flat: no
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.image-cache/
 *.pyc
 /.pytest_cache
 /tests/.coverage
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,10 @@
-*.pyc
 /.image-cache/
+*.pyc
 /.pytest_cache
 /tests/.coverage
-/tests/htmlcov-py*/
+/tests/htmlcov*
 /tests/__pycache__/
+/tests/remote-coveragedata-*
+/tests/tmp_merge_coveragerc
+/tests/total-*coveragedata
 /.tox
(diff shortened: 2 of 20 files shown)


tyll:
The Fedora 26 CI check failure is unexpected. The log at

https://larsu.fedorapeople.org/logs/linux-system-roles-network-pull-50-3af11c8-fedora-26-20180518-213512/test.log

ends with:

`TASK [linux-system-roles.network : Configure networking connection profiles] ***
task path: /tmp/tmp35497jy9/tests/roles/linux-system-roles.network/tasks/main.yml:21
fatal: [/cache/Fedora-Cloud-Base-26-1.5.x86_64.qcow2]: FAILED! => {"changed": false, "failed": true, "msg": "fatal error: failure loading libnm library: No module named gi"}
	to retry, use: --limit @/tmp/tmp35497jy9/tests/tests_ethernet.retry`

Not sure what is wrong here, any hints @thom311 ?

tyll:
This includes now commits from PR #54 to fix CI results, I will rebase once #54 is merged.

thom311:
the role uses Python + libnm + pygobject. Hence it needs some python+gobject-integration library...

Not sure which. It always confuses me. I think the package name also depends on the distibution.

How about `python3-gobject-base` package?

tyll:
@thom311 do you know how it works on RHEL7/CentOS7, IIRC there is no python-gobject-base installed by default. Should ```python3-gobject-base``` be added to the packages in the tasks file for newer systems?

thom311:
First of all, the pygobject dependency is only necessary if you use the nm provider. initscripts don't require it (e.g. RHEL6).

I guess yes, the right (which??) package should be included to the required packages list, with nm-provider.

tyll:
It seems that the required package is `python{,3}-gobject-base`.

tests/get-coverage.yml
```diff
@@ -0,0 +1,73 @@
...
+        coverage: /root/.local/bin/coverage
+    - name: set more facts
+      set_fact:
+        coverage_file: ansible-coverage-{{ coverage_module }}-{{ test_playbook|replace('.yml', '') }}
+
+    - name: set more facts
+      set_fact:
+        destdir: "remote_coverage/{{ test_playbook }}"
+      when: destdir is undefined
+
```
pcahyna: ditto. A variable set in `vars:` will be overriden by `-e `, so `when:` should not be needed.
http://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

tests/get-coverage.yml
```diff
@@ -0,0 +1,73 @@
...
+  hosts: all
+  tasks:
+    - name: set facts
+      set_fact:
+        coverage_module: network_connections
+        coverage: /root/.local/bin/coverage
+    - name: set more facts
+      set_fact:
+        coverage_file: ansible-coverage-{{ coverage_module }}-{{ test_playbook|replace('.yml', '') }}
+
```
pcahyna: Can't you just use `vars:` at the top of the play?

tests/get-coverage.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: remove old data
+      file:
+        state: absent
+        path: "{{ item }}"
+      loop:
+        - .coverage
+
+    - name: remove test data
+      shell: rm -rf /root/.ansible/ /root/.coverage.* /root/htmlcov /tmp/ansible_*
+
```
pcahyna: can this be merged with `file` above?

tests/get-coverage.yml
```diff
@@ -0,0 +1,73 @@
...
+      pip:
+        name: coverage
+        extra_args: --user --upgrade
+
+    - name: remove old data
+      file:
+        state: absent
+        path: "{{ item }}"
+      loop:
+        - .coverage
```
pcahyna: again? why?

tests/get-coverage.yml
tyll: The problem is the wildcards, AFAICS `with_fileglob` only works locally. I guess we only need `.coverage.*` anymore, the other paths are leftovers from my testing.

tests/get-coverage.yml
tyll: The problem with using `vars:` is that the variables are only set for the current playbook but I need it in several playbooks:

```
---
- name: set var
  hosts: all
  vars:
    - example_var: foo

- name: Show the var
  hosts: all
  tasks:
    - debug:
        var: example_var
```

```
ansible-playbook -i rhel7, test_things.yml

PLAY [set var] ****************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
ok: [rhel7]

PLAY [Show the var] ***********************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
ok: [rhel7]

TASK [debug] ******************************************************************************************************************************************************************************************************
ok: [rhel7] => {
    "example_var": "VARIABLE IS NOT DEFINED!"
}

PLAY RECAP ********************************************************************************************************************************************************************************************************
rhel7                      : ok=3    changed=0    unreachable=0    failed=0
```

tests/get-coverage.yml
tyll: I removed it.

tests/get-coverage.yml
tyll: The file is removed to ensure that we get fresh coverage data.

tests/get-coverage.yml
pcahyna: can you please add a comment explaining whi it needs to be this way? Also, why have two `set_facts:` ?

tyll: Done

tests/get-coverage.yml
pcahyna: yes, but why twice? (see line 31). Does any operation in-between recreate it?

tests/get-coverage.yml
```diff
@@ -1,19 +1,22 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
+# This expects the variable test_playbook to be set from the outside
 - name: Prepare for coverage extraction
   hosts: all
   tasks:
+    # Use set_fact to set variables to make them available in all playbooks
+    # vars: Would only set variables for the current playbook
```
pcahyna: "all plays" and "current play" is probably more accurate, as your example shows.

tyll: fixed

tests/get-coverage.yml
tyll: Good catch, it was wrong. Also it should have used the variable. It is all fixed and squashed now.

tyll:
Thank you for the review.

PR #51, opened 2018-05-18
Detect and fix keyerror

### LINKED ISSUE #40: KeyError "type" in connection
When a connection is to be removed ```connection['type']``` causes a KeyError when using the initscripts backend and no type is defined. This should fix it:

```
diff --git a/library/network_connections.py b/library/network_connections.py
index 271ab40..4922b13 100755
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -2650,7 +2650,7 @@ class Cmd_initscripts(Cmd):
     def run_prepare(self):
         Cmd.run_prepare(self)
         for idx, connection in enumerate(self.connections):
-            if connection['type'] in [ 'macvlan' ]:
+            if connection.get('type') in [ 'macvlan' ]:
                 self.log_fatal(idx, 'unsupported type %s for initscripts provider' % (connection['type']))
 
     def check_name(self, idx, name = None):
```

The exception was: ```File "/tmp/ansible_jD4yqb/ansible_module_network_connections.py", line 2813, in <module>     cmd.run()   File
"/tmp/ansible_jD4yqb/ansible_module_network_connections.py", line 2398, in run     self.run_prepare()   File "/tmp/ansible_jD4yqb/ansible_module_network_connections.py", line 2654, in run_prepare     if
connection['type'] in [ 'macvlan' ]: KeyError: 'type'```

tyll:
This demonstrates and fixes #40 

It also contains commits from #50 that should be merged there but are required to create the test case here

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -2673,7 +2673,7 @@ class Cmd_initscripts(Cmd):
     def run_prepare(self):
         Cmd.run_prepare(self)
         for idx, connection in enumerate(self.connections):
-            if connection['type'] in [ 'macvlan' ]:
+            if connection.get('type') in [ 'macvlan' ]:
                 self.log_fatal(idx, 'unsupported type %s for initscripts provider' % (connection['type']))
 
     def check_name(self, idx, name = None):
--- a/tests/remove-profile.yml
+++ b/tests/remove-profile.yml
@@ -1,0 +1,1 @@
+../examples/remove-profile.yml
--- a/tests/tests_ethernet.yml
+++ b/tests/tests_ethernet.yml
@@ -35,9 +35,9 @@
 #  vars:
 #    profile: "{{ interface }}"
 ## FIXME: assert profile/device down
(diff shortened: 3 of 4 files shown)


library/network_connections.py
```diff
@@ -2650,7 +2650,7 @@ def __init__(self, **kwargs):
     def run_prepare(self):
         Cmd.run_prepare(self)
         for idx, connection in enumerate(self.connections):
-            if connection['type'] in [ 'macvlan' ]:
+            if connection.get('type') in [ 'macvlan' ]:
```
thom311: shouldn't it be `connection.get('type', None)` ?

tyll: The default value is `None`.

thom311: A, I didn't realize. I thought by default it throws an exception, that's why I often explicitly set `None` as default value.

tyll: The exception happens when one uses `[]` instead of `.get()`.

Issue #35, opened 2018-03-07
[RFE] don't do any changes to network configuration for state present/absent


thom311:
The states `present` and `absent` are supposed to only deply/remove the persistant networking profiles on the host. They are not supposed to actually change the networking in any way. That is what the states `up` and `down` are for.

For initscripts, this is rather trivially the case, because the `present`/`absent` states just ensure that the ifcfg files are there or absent.

For NetworkManager, it's more complicated.

- deleting a profile that is currently active causes the device to go down. That is, because in NetworkManager an active device must always have a profile. The solution in 1.10 is [`Update2()`](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Settings.Connection.html#gdbus-method-org-freedesktop-NetworkManager-Settings-Connection.Update2) with the `NM_SETTINGS_UPDATE2_FLAG_VOLATILE` flag. Volatile means, that the connection is in-memory only, and will be automatically deleted once it goes down. If you make a connection volatile that currently is not active, it will be deleted right away.

- adding or modifying a profile might always make it a candidate for autoconnecting. For state present, we would like to add/modify the profile without any changes. For that, 1.10 supports `Update2()` flag `NM_SETTINGS_UPDATE2_FLAG_BLOCK_AUTOCONNECT`. That works for the update case, but doesn't work for the add case. To add a connection that has `connection.autoconnect=yes` without activating it right away, we would either need new NetworkManager D-Bus API to block autoconnect form the start, or we first add the connection with `connection.autoconnect=no`, and then follow up with an `Update2` call that sets `connection.autoconnect=yes` and uses `NM_SETTINGS_UPDATE2_FLAG_BLOCK_AUTOCONNECT`.

- modifying the properties `connection.metered` or `connection.zone` takes effect immediately. There is currently no way to prevent that. NetworkManager should get a new flag for `Update2` that allows to modify the profile and prevent these two properties from updating right away. Then, the role should use this new API.

tyll:
For removing a profile, there is no #95

tyll:
NetworkManager bugs for additional features:

- https://bugzilla.redhat.com/show_bug.cgi?id=1677068 - to allow blocking changes when creating a new profile/connection
- https://bugzilla.redhat.com/show_bug.cgi?id=1677070 - to support blocking `connection.metered`, `connection.zone` and ideally everything else when updating a profile

PR #53, opened 2018-05-18
Inconsistent behavior for bridges on RHEL6 and RHEL7

tyll:
This fails on RHEL6 and is related to #35 iirc.

### DIFF
--- a/tests/tests_bridge.yml
+++ b/tests/tests_bridge.yml
@@ -49,7 +49,6 @@
     profile: "{{ interface }}"
     task: tasks/assert-profile_absent.yml
 
-# FIXME: Devices might still be left when profile is absent
-#- import_playbook: run-tasks.yml
-#  vars:
-#    task: tasks/assert-device_absent.yml
+- import_playbook: run-tasks.yml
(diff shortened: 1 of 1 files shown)


tyll:
[citest]

tyll:
The first two commits are parte of #72

pcahyna:
Should this PR be merged?

tyll:
No, this was only to have a test case to be able to easily reproduce the inconsistent behavior. When the states are re-defined, then it can be adjusted and merged I guess.

pcahyna:
>  When the states are re-defined, then it can be adjusted and merged I guess.

States were re-defined in #77. Can the test be adjusted now, please?

tyll:
The new state definitions do not define the behavior for virtual interfaces. The test here checked whether the bridge device was removed after the profile was removed. This is now the `persitence_state` and it only defines whether the profile is on-disk but not whether the device exists. One could argue that `state:down` should remove the device but I am more convinced that it is not important to have a defined interface state. Therefore I close this PR now.

PR #54, opened 2018-05-22
Fix exception on Python3 in _link_read_permaddress() and for sorting network connections

tyll:
Fix python3 support.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -16,10 +16,11 @@ options: Documentation needs to be written. Note that the network_connections mo
   this module outside the role. Thus, consult README.md for examples for the role.
 '''
 
+import functools
+import os
 import socket
 import sys
 import traceback
-import os
 
 ###############################################################################
 
@@ -52,6 +53,21 @@ class ValidationError(MyError):
     @staticmethod
     def from_connection(idx, message):
         return ValidationError('connection[' + str(idx) + ']', message)
+
+
+# cmp() is not available in python 3 anymore
+if "cmp" not in dir(__builtins__):
+    def cmp(x, y):
+        """
+        Replacement for built-in function cmp that was removed in Python 3
+
+        Compare the two objects x and y and return an integer according to
+        the outcome. The return value is negative if x < y, zero if x == y
+        and strictly positive if x > y.
+        """
+
+        return (x > y) - (x < y)
+
 
 class Util:
 
@@ -1630,7 +1646,12 @@ class NMUtil:
             if t_b <= 0:
                 return -1
             return cmp(t_a, t_b)
-        cons.sort(cmp = _cmp)
+
+        if Util.PY3:
+            # functools.cmp_to_key does not exist in Python 2.6
+            cons.sort(key=functools.cmp_to_key(_cmp))
+        else:
+            cons.sort(cmp=_cmp)
         return cons
 
     def connection_compare(self, con_a, con_b, normalize_a = False, normalize_b = False, compare_flags = None):
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -83,13 +83,14 @@ class Util:
         return default
 
     @staticmethod
-    def check_output(argv, lang = None):
+    def check_output(argv):
         # subprocess.check_output is python 2.7.
         with open('/dev/null', 'wb') as DEVNULL:
             import subprocess
-            ev = os.environ.copy()
-            ev['LANG'] = lang if lang is not None else 'C'
-            p = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=DEVNULL, env=ev)
+            env = os.environ.copy()
+            env['LANG'] = 'C'
+            p = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=DEVNULL,
+                                 env=env)
             # FIXME: Can we assume this to always be UTF-8?
             out = p.communicate()[0].decode("UTF-8")
             if p.returncode != 0:
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -74,7 +74,8 @@ class Util:
             ev = os.environ.copy()
             ev['LANG'] = lang if lang is not None else 'C'
             p = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=DEVNULL, env=ev)
-            out = p.communicate()[0]
+            # FIXME: Can we assume this to always be UTF-8?
+            out = p.communicate()[0].decode("UTF-8")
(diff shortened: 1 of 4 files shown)


tyll:
It just uses ethtool to read the hardware address of the lo interface that should always be available. IMHO it is a unit test, since it just tests single units. Calling tools and working on the results is exactly what is broken with Python 3 since the string type changed. I could maybe add a demo executable that just works like ethtool and call that instead if you prefer. A better approach would be to use ioctl directly but I would prefer to get this working somehow before making it perfect.

thom311:
Does calling ethtool actually need root permissions? I think not.
Maybe you got a failure before, because ethtool is in /usr/sbin, which wasn't in $PATH. But you fixed that already, so do you still need `become`?

library/network_connections.py
```diff
@@ -1630,7 +1646,12 @@ def _cmp(a, b):
             if t_b <= 0:
                 return -1
             return cmp(t_a, t_b)
-        cons.sort(cmp = _cmp)
+
+        if sys.version_info[0] == 3:
```
thom311: There is also `Util.PY3` above. Either we make use of this `PY3` helper, or we drop it and use `sys.version_info[0]`.

tests/tests_unit.yml
```diff
@@ -29,18 +30,31 @@
       ignore_errors: true
       register: python2_available
 
-    - command: python2 /tmp/test-unit-1/test_network_connections.py
+    - command: python2 /tmp/test-unit-1/test_network_connections.py --verbose
       when: python2_available | succeeded
-
+      register: python2_result
```
thom311: if this command fails, I presume it ignores the error and proceeds with the other tests, right? That makes sense.

but at the end of the play, after all the debug messages, don't you lack a final command to fail the entire play/test, if any of the `python*_results` is not success?

tyll: It will run when python2 is available and fail when the tests fail. Then the ```python_results``` are not used. I added them to be able to see if/which tests are skipped.

tests/tests_unit.yml
```diff
@@ -1,5 +1,17 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
+- hosts: all
+  name: Setup for test running
+  tasks:
+    # FIXME: change with_items to loop when test-harness is updated
```
pcahyna: do you have a specific updatre to test-harness in mind?

tyll: It needs to be deployed with a newer Ansible version, e.g. using Fedora 28.

tests/tests_unit.yml
```diff
@@ -29,18 +43,31 @@
...
       register: python3_available
 
-    - command: python3 /tmp/test-unit-1/test_network_connections.py
+    - command: python3 /tmp/test-unit-1/test_network_connections.py --verbose
       when: python3_available | succeeded
+      register: python3_result
+
+    - command: python /tmp/test-unit-1/test_network_connections.py --verbose
+      register: python_result
+
```
pcahyna: Is there a need to invoke bare `python` when there are `python2` and `python3` invocations?

library/network_connections.py
```diff
@@ -74,7 +74,8 @@ def check_output(argv, lang = None):
             ev = os.environ.copy()
             ev['LANG'] = lang if lang is not None else 'C'
             p = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=DEVNULL, env=ev)
-            out = p.communicate()[0]
+            # FIXME: Can we assume this to always be UTF-8?
+            out = p.communicate()[0].decode("UTF-8")
             if p.returncode != 0:
```
pcahyna: I suppose we should assume it is the current locale, if it is an output of a command? That's the point of a locale after all.

pcahyna: That said in another locale we will not be able to parse the output. Could we omit `lang` and set `LC_ALL` to `C` unconditionally and then parse the output as pure ASCII?

tests/tests_unit.yml
```diff
@@ -11,6 +11,17 @@
...
+        state: directory
+        dest: /tmp/test-unit-1/helpers
+
+    - copy:
+        src: "{{ item }}"
+        dest: /tmp/test-unit-1/helpers
+      with_fileglob:
+        - helpers/*
+
+    - shell: chmod -R a+x /tmp/test-unit-1/helpers
```
pcahyna: Is that really necessary? The file is executable (should be, I suppose, if you want to run the test locally) and I believe `copy` preserves the permissions (if not, use the `mode` parameter).

library/network_connections.py
thom311: Maybe `Util.check_output()` should just return whatever python gives us. The caller must know how to interpret this and `decode()` it accordingly.

tests/tests_unit.yml
tyll: It is the only command that is guaranteed to be called, if there is no python2 or python3 (e.g. no symlink for these), then no tests would be called.

pcahyna: Can it really happen (on systems that we support) that `python` exists and neither `python2` nor `python3` do?
By the way, I would not call `python` "guaranteed" in the future, given https://fedoraproject.org/wiki/Changes/Move_usr_bin_python_into_separate_package

tests/tests_unit.yml
tyll: Good catch, I will fix it.

library/network_connections.py
pcahyna: > I doubt that many people use it to set some other encoding than UTF-8

I think we should not rely on such assumptions.

> The locale is usually about translation

That was my point. If one translates `ethtool` output, it will not be possible to parse it. Given that,
> I plan to drop support for the lang parameter, too, because we will not want to use some other locale here in the module.

makes perfect sense.

> I guess hardcoding to C and decoding as ascii should work unless we really approach something that is not ascii

One should hardcode C to ensure no translation is used if one relies on that.

> unless we really approach something that is not ascii

We could revisit it when such a use case appears. By the way,

> `cat /path/to/utf-8-file.txt` will not necessarily be correct ascii.

`cat /path/to/binary-file.txt`will not be correct UTF-8 either... I think the correct answer is "don't do that"...

tests/tests_unit.yml
tyll: Hm, I guess I will address this eventually, we can probably add something that fails explicitly if not at least one binary was called. I would like to address this in the future, maybe I will change the way the unit tests are called anyhow.

PR #58, opened 2018-05-30
Fix handling of autoconnect parameter

smasc:
This fixes a bug in the `network_connections` module when using the `initscripts` provider and encountering a connection with `autoconnect `set to false. It fails to write any `ONBOOT=` line to the interface's `ifcfg-*` file, which RHEL treats as equivalent to `ONBOOT=yes`.

This explicitly sets `ONBOOT` for all interfaces defined, instead of relying on what RHEL considers its default value.


tyll:
Thank you very much!

PR #60, opened 2018-06-11
th/main-function

thom311:
based on PR #59.

Untested, but by looking at the code, it should be fine (and nobody using these global variables)

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -2819,7 +2819,7 @@ class Cmd_initscripts(Cmd):
 
 ###############################################################################
 
-if __name__ == '__main__':
+def main():
     connections = None
     cmd = None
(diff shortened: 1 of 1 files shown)


pcahyna:
Something is wrong with the CI, even for the tests that succeeded I can't access their logs (404 error). @tyll , I think you touched the test harness recently, any idea?

tyll:
@pcahyna I fixed it here: https://github.com/linux-system-roles/test-harness/commit/437eb8a0bca1d8a46f91825b26e504d4d125c321 and updated the image, not sure why it still creates wrong URLs

tyll:
I guess I redeployed it too fast and it did not have the changes from my commit but only changes from an earlier PR, therefore it appeared to be updated but was not. Should be fixed in the next deployment.

PR #64, opened 2018-06-29
moved default-nm.yml to default.yml

xrowdominik:
You are not able to include a variable file in another variable file. Ansible will exit with an error. Also there seems to be no need to add two different default files if one default file is including the other one.

The ansible error: "message": "/roles/network/vars/default.yml must be stored as a dictionary/hash"

### DIFF
--- a/vars/default-nm.yml
+++ b/vars/default-nm.yml
@@ -1,2 +1,0 @@
-# SPDX-License-Identifier: BSD-3-Clause
-network_provider_default: nm
--- a/vars/default.yml
+++ b/vars/default.yml
@@ -1,1 +1,2 @@
-default-nm.yml
+# SPDX-License-Identifier: BSD-3-Clause
+network_provider_default: nm


tyll:
@xrowdominik Thank you for your report! What did you do to see the error that you show? Can you please wirte down the steps that you performed? The files do not include each other but there are symlinks, also this changes breaks the role:
```
TASK [linux-system-roles.network : Set version specific variables] ********************************************
task path: /tmp/tmpf7o2c61k/tests/roles/linux-system-roles.network/tasks/main.yml:2
fatal: [/home/till/lsr-test-harness-files/cache/rhel-guest-image-7.5-146.x86_64.qcow2]: FAILED! => {"msg": "No file was found when using with_first_found. Use the 'skip: true' option to allow this task to be skipped if no files are found"}
```

xrowdominik:
I simply executed your role and got this error. It is the first task of your role so nothing crazy happened :-). Could you explain how to use these symlinks in ansible? Do i need to configure something for that? It seems that my system won´t allow me to use them. I´m running on a Centos 7 maschine by the way, so none of the defaults would match. Also my ansible version is 2.5.2.

> TASK [network : Set version specific variables] *******************************************************************************************************************************
failed: [openshift.xrow.lan] (item=/scm/xrow.infrastructure/roles/network/vars/default.yml) => {"ansible_facts": {}, "ansible_included_var_files": [], "changed": false, "item": "/scm/xrow.infrastructure/roles/network/vars/default.yml", "message": "/scm/xrow.infrastructure/roles/network/vars/default.yml must be stored as a dictionary/hash"}
        to retry, use: --limit @/scm/xrow.infrastructure/openshift.xrow.lan.retry

pcahyna:
@xrowdominik I think you should show us your entire playbook which  triggers the error.

tyll:
Three questions:
- How did you install the role?
- What does `ls -la defaults/` return?
- Does it work for you with these changes?

 You should not need to do anything special, https://github.com/linux-system-roles/network/blob/master/tasks/main.yml#L4 will select the default.yml file which is a symlinkt to default-nm.yml.

xrow:
Hi,

if default.yml is a symlink to default-nm.yml than that is the actual problem. We windows desktop users that use ansible can`t user the symlinks. Resolving the symlink to a text file is therefore correct.

xrow:
They the way... the symlink is converted to a file with the contents:

`default-nm.yml`

xrow:
How do you want to see this resolved? We can prepare this for you.

tyll:
@xrow thank you for the offer and the details, please see #65 as a possible fix. Let's discuss it there.

pcahyna:
Hello @xrow , it is likely that we will return to using symlinks for deduplicating code, because we believe that avoiding them is not worth the loss of simplicity - other solutions are much less elegant. I.e. we may consider reverting the change or at least introducing symlinks in new code. Could you please explain more about your problem? Are symlinks not supported even under WSL? By the way, I see here that "Windows is not supported for the control node." ( https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#control-node-requirements ), so I don't understand where does the need of supporting Windows control hosts in linux-system-roles come from.

PR #65, opened 2018-07-02
Use set_facts to set default provider

tyll:
This avoids symlinks that do not work on Windows:
https://github.com/linux-system-roles/network/pull/64#issuecomment-401740870


xrow:
Looks good to me. That shall fix our issue.

PR #74, opened 2018-07-21
Use network manager only when it is running

### LINKED ISSUE #71: [RFE] default to currently used network provider
Instead of defaulting to a network provider depending on the platform check if the network service or NetworkManager are already enabled and use the respective service.

tyll:
- Fixes #71 

I guess some documentation update and tests are still missing. I am not sure how to use the standard test interface to test whether a role failed when NetworkManager is no running. Ideally we should be able to mock the Ansible service fact so we do not actually need to stop NetworkManager since it will most likely break access to the test VM.


tyll:
This will probably also need some doc changes when we decided on the final algorithm.

pcahyna:
I am a bit lost now, what is the algorithm currently implemented in the PR?

Btw: tests are failing.

pcahyna:
Have you pushed it?

tyll:
Yes, there is an error with using the variable `network_provider_current` before it is defined since it will only be defined when the role ran. I have now a fix locally but it seems there is a problem with overriding the network_provider (it does not seem to happen in the tests).

tyll:
I added another commit (not meant to be merged) so you can see the current code. For some reason `network_provider` is not properly overridden when importing the `tests_ethernet.yml` playbook, therefore the test runs twice against nm on RHEL7 AFAICS. I fear that the extra magic in the tasks file for the default variable somehow broke the precedence expectations.

pcahyna:
Would it help to set the variable in vars instead of tasks? See https://github.com/linux-system-roles/timesync/blob/master/vars/main.yml

tyll:
The problem is not the default value, since it is ignored anyhow, but the value set when doing import_playbook does not seem to be used. It might also just be related to setting variables when importing a playbook since it seems to work when doing a set_fact task in the same playbook. Or maybe it is related to first running the role to get the value of `network_provider_current`.

pcahyna:
>  Or maybe it is related to first running the role to get the value of network_provider_current

You should not need that, variables are evaluated lazily. https://github.com/ansible/ansible/issues/10374
Running a role twice can also require special magic, https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-duplication-and-execution

tyll:
> You should not need that, variables are evaluated lazily. ansible/ansible#10374

I need to know which network provider the role chooses by default for an image to explicitly run the role with the actual configuration for the other provider to ensure that both providers are tested for all images (except RHEL6/CentOS6). Previously this just meant to also use the initscripts provider for of these images but now it requires nm for the CentOS images for example.

The role already runs multiple times since it gets different variables.

Btw. when not running the role with default values before setting network_provider it seems to pick up the setting.

pcahyna:
>   I need to know which network provider the role chooses by default for an image to explicitly run the role with the actual configuration for the other provider

I still don't see why you need two role executions for that. Can you paste a code sample showing this (how the playbook runs the role for the other provider)?

tests/ensure_non_running_provider.py
```diff
@@ -19,19 +19,20 @@
...
+
+- hosts: all
+  tasks:
+    - debug:
+        var: network_provider_current
 
 - import_playbook: "{tests_playbook}"
   vars:
-    network_provider: nm
+    network_provider: "{{{{ 'initscripts' if network_provider_current == 'nm' else 'nm' }}}}"
```
pcahyna: Why so many `{`s?

pcahyna:
Do you have to split the tests into multiple plays using `import_playbook`? Can't you just include lists of tasks using `include_tasks` instead of whole playbooks?

tests/ensure_non_running_provider.py
tyll: Ansible requires two  curly braces but they are also used for str.format(), therefore they need to be escaped in Python, which requires to double them.

tyll:
> Do you have to split the tests into multiple plays using import_playbook? Can't you just include lists of tasks using include_tasks instead of whole playbooks?

The basic test structure ist
1. setup the test inferfaces
2. run the role with some config
3. verify the results
4. repeat 2./3.
5. cleanup

To run the role multiple times and verify the results I use multiple plays. See for example:
https://github.com/linux-system-roles/network/blob/master/tests/tests_ethernet.yml

Now I also need to run this playbook again (from a file with a different name to make sure that the test harness will use a new VM for it) but with the network_provider set to the value that was not detected.

I just learned about import_role/include_role, so maybe this could be used instead to be able to create a task list, however, then we would need at least three files per test

- list of tasks for the tests
- playbook to run the tasks for the current provider
- playbook to run the tasks for the other provider

maybe it could be reduced to just two files again, since it is also possible to import or include tasks from a playbook but I am not completely sure about this. But I am not sure if this will still allow to express everything the same way. Currently it is also possible to have generic playbooks to down or remove a profile and use them in the tests.

> I still don't see why you need two role executions for that. Can you paste a code sample showing this (how the playbook runs the role for the other provider)?

Here is the code to run the `tests_bridge.yml` playbook with the other network provider, and to know which one would be the other one, I need to run the role to get access to `network_provider_current`:

```
---
# empty play to gather facts
- hosts: all

- hosts: all
  roles:
    - linux-system-roles.network

- import_playbook: "tests_bridge.yml"
  vars:
    network_provider: "{{ 'initscripts' if network_provider_current == 'nm' else 'nm' }}"
  when:
```

I have another idea, maybe the code to determine `network_provider_current` could be separated to be able to access it without running the role completely.

tyll:
I cleaned it up, I guess that setting `network_provider` via `set_facts` caused the problem with overriding the variable. I created a minimal example so I might follow-up on this with the Ansible folks. Inspired by the advanced jinja-templating from the timesync role as used in https://github.com/linux-system-roles/timesync/blob/master/vars/main.yml I reworked the provider detection to be in `defaults/main.yml` now. Let's see what the CI thinks about this approach.

tyll:
[citest bad]

tyll:
The current approach is flawed, too, since network_provider_current will be evaluated each time the role is called and therefore once NetworkManager was installed once, it will be considered to be the current provider for later runs. So here is another attempt to just use the yaml module to parse the definition for network_provider_current from the defaults and put it into the other playbook. It is becoming more and more fragile :-/ I guess ideally we should have an inventory for each image that also contains the expected current provider, so that we can

1) test that the current provider detection works
2) use this value to define the other provider and run it with less magic. But this would also require changes to the test harness

pcahyna:
> The current approach is flawed, too, since network_provider_current will be evaluated each time the role is called and therefore once NetworkManager was installed once, it will be considered to be the current provider for later runs.

That's the intent, right? It is the expected behaviour that once a provider has been selected, the role preserves it.
To be sure, are you saying that the current approach to provider selection is flawed, or that the current approach to testing it is flawed?

tyll:
> To be sure, are you saying that the current approach to provider selection is flawed, or that the current approach to testing it is flawed?

I meant the approach to testing. It should be "solved" in the fixup commit as described in the comment by parsing the yaml code manually as part of generating the test case files for the "other" provider. It is not nice, also having generated files being commited to the repo. But there also seem to be limitations in using Ansible/the standard CI interface for testing.

tyll:
@pcahyna Do you have an idea for a better approach? IMHO having the info in the inventory would make most sense, since we could then also verify that the auto detection finds the correct value but it would also require changes to the test harness and possibly the inventory script from standard-test-roles. If it is ok, I will squash the fixup commit. Only need to figure out why yaml is not found in travis even though the logs show that it was installed...

pcahyna:
> I cleaned it up, I guess that setting network_provider via set_facts caused the problem with overriding the variable. Inspired by the advanced jinja-templating from the timesync role as used in https://github.com/linux-system-roles/timesync/blob/master/vars/main.yml I reworked the provider detection to be in defaults/main.yml now.

Did that and commit 96157ca8d958ee9e4dba368e9b0934224c7eae7c help in any way? This approach seems good to me.

pcahyna:
> IMHO having the info in the inventory would make most sense, since we could then also verify that the auto detection finds the correct value but it would also require changes to the test harness and possibly the inventory script from standard-test-roles.

The recommended practice is to define inventory variables using host_vars and group_vars directories, and not to define them directly in the inventory file or files. See https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#splitting-out-host-and-group-specific-data.

I suppose that creating those directories (`host_vars` actually since we don't seem to have any use for `group_vars`) should avoid changes to the inventory script and test harness, but maybe it is not as simple as I think.

Generally, defining the expected default provider for each image in a variable seems like a good approach to me. Normally, that should be the `network_provider_os_default` variable, right? And you need another variable just because the CentOS 7 image does not conform to the OS default? Or is there another reason for introducing it?

tyll:
> Did that and commit 96157ca help in any way? This approach seems good to me. 
Yes, it should do the right thing.

> Is every role run supposed to take place on a new VM?

No, there need to be multiple role runs in a VM to be able to test changing or removing profiles. I.e. create a profile, check that it is there, modify it, check that it was modified, removing it, ...

To test that the provider auto-detection works properly, we would need a list for all CI images that states which provider should be used, so this can be verified and the decision on which provider would be not the default one could also be based on this static decision.

pcahyna:
> No, there need to be multiple role runs in a VM to be able to test changing or removing profiles. I.e. create a profile, check that it is there, modify it, check that it was modified, removing it, ...

How does that work with multiple plays? Each play is supposed to start its own VM.

tyll:
> How does that work with multiple plays? Each play is supposed to start its own VM.

It is one VM per playbook (i.e. for each file that matches `tests_*.yml`) not per play. And the playbook contains multiple plays that will run on the same VM.

pcahyna:
> It is one VM per playbook (i.e. for each file that matches tests_*.yml) not per play. And the playbook contains multiple plays that will run on the same VM.

Interesting. I am afraid there is no guarantee that this will continue to work.

tyll:
> Interesting. I am afraid there is no guarantee that this will continue to work.

Why? The way it is implemented (using the inventory script to create a new VM) would not allow to parse playbooks and run new VMs for each play.

pcahyna:
because it is undefined when Ansible re-queries the dynamic inventory.

tyll:
[citest bad]

tyll:
[citest bad]

PR #75, opened 2018-07-25
Use profile name for interface name by default

### LINKED ISSUE #41: set interface_name to name by default unless mac is specified
```name``` used to specify only the profile name and ```interface_name``` or ```mac``` could be specified to define the interface that the profile is restricted, too. This seems to be unexpected, since a profile called ```eth0``` might configure ```eth2``` if no ```interface_name``` or ```mac``` is specified. Therefore change it to

* by default ```name``` is used as ```interface_name```
* if ```interface_name``` is explicitly set empty, it will not be set
* if ```mac``` is specified, ```interface_name``` defaults to not being set

### PR DESCRIPTION
**Pull request by tyll**
* by default ```name``` is used as ```interface_name```
* if ```interface_name``` is set to the empty string, it will not be set
* if ```mac``` is specified, ```interface_name``` defaults to not being set

This fixes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,31 +12,35 @@ The role can be used to configure:
 - Bonded interfaces
 - VLAN  interfaces
 - MacVLAN interfaces
+- Infiniband interfaces
 - IP configuration
 
 General
 -------
 
-The role supports two providers: `nm` and `initscripts`. The provider can be configured per host
-via the [`provider`](#provider) variable. In absence of explicit configuration, it is autodetected based on
-the distribution. The `nm` provider is used by default on RHEL7 and
-`initscripts` on RHEL6. However, note that the provider is not tied to a certain distribution,
-given that the required API is available. For `nm` this means that at least version 1.2 of NetworkManager's
-API is available. For `initscripts`, it requires the legacy network service as commonly available
-on Fedora/RHEL.
+The role supports two providers: `nm` and `initscripts`. The provider can be
+configured per host via the [`network_provider`](#provider) variable. In
+absence of explicit configuration, it is autodetected based on the
+distribution. The `nm` provider is used by default on RHEL7 and `initscripts`
+on RHEL6. However, note that the provider is not tied to a certain
+distribution, given that the required API is available. For `nm` this means
+that at least version 1.2 of NetworkManager's API is available. For
+`initscripts`, it requires the legacy network service as commonly available on
+Fedora/RHEL.
 
-For each host a list of networking profiles can be configure via the `network` variable.
+For each host a list of networking profiles can be configure via the
+`network_connections` variable.
 
 - For initscripts, profiles correspond to ifcfg files in `/etc/sysconfig/network-scripts/ifcfg-*`.
 
 - For NetworkManager, profiles correspond to connection profiles as handled by NetworkManager.  Fedora and RHEL use the `rh-plugin` for NetworkManager which also writes configuration files to `/etc/sysconfig/network-scripts/ifcfg-*` for compatibility.
 
 Note that the role primarily operates on networking profiles (connections) and
-not on devices. For example, in the role you would not configure the current IP address
-of an interface. Instead, you create a profile with a certain IP configuration and
-optionally activate the profile on a device. Which means, to apply the configuration
-to the actual networking interface.
-
+not on devices but it defaults to use the profile name as the interface name.
+But it is also possible to create generic profiles, by creating for example a
+profile with a certain IP configuration without activating the profile. To
+apply the configuration to the actual networking interface, a command like
+`nmcli` needs to be used on the target system.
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tyll** [comment]
[citest bad]

**Comment by tyll** [comment]
@pcahyna did you already take a look here?

File: README.md [hunk 1]
```diff
@@ -151,18 +151,22 @@ to match a non-virtual device with the profile.
...
-a profile is not restricted to any interface by name.
-Note that with [persistent interface naming](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/ch-Consistent_Network_Device_Naming.html),
+For type `ethernet`, this option restricts the profile to the given interface
+by name. This argument is optional and by default the profile name is used
+unless a mac address is specified using the `mac` key. Specifying an empty
+string (`""`) allows to specify that the profile is not restricted to a network
+interface.
+
+
+!!! note With [persistent interface naming](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/ch-Consistent_Network_Device_Naming.html),
```
pcahyna: What do the three exclamation marks do?

pcahyna:
The `!!! note` in documentation should be removed, otherwise it looks good (as far as I can say).

README.md
tyll: It was meant to create an admonition according to the python markdown docs but it does not seem to work in GitHub. I will change it.

README.md
```diff
@@ -151,18 +154,22 @@ to match a non-virtual device with the profile.
 
 ### `interface_name`
 
-For type `ethernet`, this option restricts the profile to the
-given interface by name. This argument is optional and by default
-a profile is not restricted to any interface by name.
-Note that with [persistent interface naming](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/ch-Consistent_Network_Device_Naming.html),
+For type `ethernet`, this option restricts the profile to the given interface
```
pcahyna: I guess it is not only for `ethernet` but for any physical interface type (e.g. `infiniband`)?

README.md
```diff
@@ -17,26 +17,29 @@ The role can be used to configure:
...
-not on devices. For example, in the role you would not configure the current IP address
-of an interface. Instead, you create a profile with a certain IP configuration and
-optionally activate the profile on a device. Which means, to apply the configuration
-to the actual networking interface.
-
+not on devices but it defaults to use the profile name as the interface name.
+But it is also possible to create generic profiles, by creating for example a
+profile with a certain IP configuration and optionally activate the profile on
+a device. Which means, to apply the configuration to the actual networking
+interface.
```
pcahyna: How does one do that?

README.md
```diff
@@ -17,26 +17,29 @@ The role can be used to configure:
...
-given that the required API is available. For `nm` this means that at least version 1.2 of NetworkManager's
-API is available. For `initscripts`, it requires the legacy network service as commonly available
-on Fedora/RHEL.
-
-For each host a list of networking profiles can be configure via the `network` variable.
+The role supports two providers: `nm` and `initscripts`. The provider can be
+configured per host via the [`network_provider`](#provider) variable. In
+absence of explicit configuration, it is autodetected based on the
+distribution. The `nm` provider is used by default on RHEL7 and `initscripts`
+on RHEL6. However, note that the provider is not tied to a certain
```
pcahyna: I think this belongs to #74 (and does not reflect the current state of that PR).

tyll: This describes still the old behavior. The change here is about the variable name, it just says `provider` but it is actually `network_provider`, the same is for `network` vs `network_connections`. I noticed this while fixing the other issues in the README.

README.md
tyll: This needs to be done separately, e.g. by using `nmcli`.

README.md
tyll: Good catch, infiniband is the other network device type that supports this. AFAICS, there is no other physical interface type supported in the code.

README.md
pcahyna: There is an unrelated change to line folding though.

README.md
pcahyna: Thank you for clarifying this, because from the previous version of the description it was not obvious at all that this would need to be done outside of the role.

README.md
pcahyna: What about
> For physical interface types (`ethernet` and `infiniband`)

to make clear what is the logic of supporting it for some types and not for others.

tyll: The `ethernet` also works for virtual interfaces that created outside the role, e.g. veth interfaces and probably also bridges or maybe dummy interfaces. Then the device will not be created by the role but only be configured like a regular ethernet interfaces as far as I can tell.

PR #83, opened 2018-08-24
Print network provider early

tyll:
The list of packages to install depends on the network provider, so
print it out early.


tyll:
Yes, it is for debugging when it fails to see if the correct provider was selected.

Issue #70, opened 2018-07-12
[RFE] add profile setting ignore-error-if-absent to ignore errors when downing a profile that does not exist


tyll:
Currently the role fails when ask to set a profile to `down` when the profile is not completely defined in the profile and does not exist on the target system. One could argue that a profile is `down` when  it does not exist on the system. However, one could also argue that the role should report this as an error in case there is a typo. Introduce a setting called `ignore-error-if-absent` that when it is true ignores the error when a profile is undefined and should be set to `down`.

pcahyna:
what does it mean "when the profile is not completely defined in the profile"? Can you provide an example of such a profile, as a snippet of the network_connections variable?
> when the profile is not completely defined in the profile and does not exist on the target system.

What will happen when the profile is not completely defined, but it DOES exist on the target system?

tyll:
> what does it mean "when the profile is not completely defined in the profile"? Can you provide an example of such a profile, as a snippet of the network_connections variable?

It is basically this, just a name and the state:
https://github.com/linux-system-roles/network/blob/master/tests/down-profile.yml

> What will happen when the profile is not completely defined, but it DOES exist on the target system?

Then it will now down the profile using `ifdown $profile` or via Network Manager because all the required information is available.

When `interface_name` is defined, we can also call ` ip link set dev  $interface state down` when the profile is not found but the interface exists to allow to down interfaces for which the profile was already set to `absent`.

pcahyna:
> Currently the role fails when ask to set a profile to `down` when the profile is not completely defined in the profile and does not exist on the target system. One could argue that a profile is `down` when it does not exist on the system. However, one could also argue that the role should report this as an error in case there is a typo.

What about reporting an error only if the profile does not exist AND the corresponding interface does not exist either? (By "corresponding interface" I mean the interface determined in the standard way - i.e. match on `interface_name` (with the default being `name`) or `mac`.)
If there is a corresponding interface, then presumably it was not a typo.

pcahyna:
Thinking more about it, I don't think we should worry about typos. Not being able to detect typos in some cases is the price to pay for idempotency, which is one of the key concepts in Ansible. The file module does not report an error with state=absent either, even if the file is not there and the name given could be a typo. This is by design; contrast with the rm command, which can detect typos in this way, but is not idempotent.

tyll:
[...]

PR #86, opened 2018-09-27
Update and cleanup README

yontalcar:


README.md
```diff
@@ -15,251 +18,331 @@ The role can be used to configure:
...
+  - `bridge`
+  - `bond`
   - `team`
   - `vlan`
+  - `macvlan`
+  - `infiniband`
+
+#### `type: ethernet`
+
+The `type:ethernet` should be specified as a dictionary with the following
```
tyll: If the type is ethernet, then there can be an extra `ethernet` dictionary with the following settings.

README.md
```diff
@@ -455,12 +469,7 @@ network_connections:
       auto6: no
 ```
 
-Like for `master`, the `parent` references the connection profile in the ansible
-role.
-
-### `type: macvlan`
-
-MACVLANs also work:
+Configuring macvlan:
```
tyll: Since VLAN is all upper-case, maybe it should be MACVLAN as well or at least MacVLAN.

README.md
```diff
@@ -483,102 +492,128 @@ network_connections:
...
+components that rely on the ifcfg files and react on changes.
+
+The `initscripts` provider requires the different profiles to be in the right
+order when they depend on each other. For example the bonding master device
+needs to be specified before the slave devices.
+
+When removing a profile for NetworkManager it also takes the connection
+down and possibly removes virtual interfaces. With the `initscripts` provider
+removing a profile does not change its current runtime state (this is a future
+deployment for NetworkManager as well).
```
tyll: maybe use "future feature" instead of "future deployment". Deployments sounds wrong in this context.

tyll:
@thom311 would you please also take a look?

README.md
```diff
@@ -15,251 +18,331 @@ The role can be used to configure:
...
 
-Valid values for `type` are:
 
-  - `bond`
-  - `bridge`
+### `persistent_state`
+
+The `persistent_state` option identifies if a connection profile is in a persistent state. The `persistent_state` option can be set to the following values:
+
+*   `present` (default)
```
tyll: The other options use a formatting like ```#### `persistent_state: present` ```, IMHO it makes sense here, too

README.md
```diff
@@ -15,251 +18,331 @@ The role can be used to configure:
...
-network_connections:
-  - name: eth0
-    #persistent_state: present  # default
-    type: ethernet
-    autoconnect: yes
-    mac: 00:00:5e:00:53:5d
-    ip:
-      dhcp4: yes
-```
+You can deactivate a connection profile, even if is currently not active. As a consequence, `state: down` always changes the system.
```
thom311: @tyll for 'nm' it seems the role should just silently accept if the profile is currently not active (or even not exists), no? Seems in code, that is currently not happening -- as described in the README here.

yontalcar:
Ok, I have applied the changes. If you are OK with the current state, I should probably squash these two patches.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
-incomplete (lacks the `type` setting) and `persistent_state` is `present`,
-the behavior is undefined. The value `absent` makes the role ensure
-that the profile is not present on the target host.
+- For `initscripts`, this corresponds to `ifup {{name}}`.
 
+When the `state` option is set to `up`, you can also specify the `wait` option (optional):
 
-#### Example
+* `wait: 0` - initiates only the activation, but not wait until the device is fully connected.
+The connection is completed, for example after a DHCP lease received.
```
pcahyna: This is not intelligible. I prever the original version:
> Connection will complete in the background, for example after a DHCP lease was received.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
 
-The value `present` for the `persistent_state` setting does not directly
-result in a change in the network configuration. That is, without `state`
-set to `up`, the profile is only created or modified, not activated.
+The `bridge`, `bond`, `team` device types work similar. Note that `team` is not supported in RHEL6 kernels.
+
+For slaves, the `slave_type` and `master` properties must be set. Note that slaves should not have `ip` settings.
+
+The `master` refers to the `name` of a profile in the ansible
+Playbook. It is neither an interface-name nor a connection-id of
```
pcahyna: Ansible playbook

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
-  `autoconnect` turned on by default. Thus, NetworkManager may very well decide
-  right away to activate the new profile on a currently disconnected device.
-  ([rh#1401515](https://bugzilla.redhat.com/show_bug.cgi?id=1401515)).
 
 ### `autoconnect`
 
+For NetworkManager, the new connection profile is created when the `autoconnect`
+option is enabled by default. Therefore, NetworkManager can activate the new
+profile on a currently disconnected device. ([rh#1401515](https://bugzilla.redhat.com/show_bug.cgi?id=1401515)).
+
```
pcahyna: This duplicates the information below and should have rather been keept together with the discussion of creating new profiles.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+- For `initscripts`, this corresponds to `ifup {{name}}`.
 
+When the `state` option is set to `up`, you can also specify the `wait` option (optional):
 
-#### Example
+* `wait: 0` - initiates only the activation, but not wait until the device is fully connected.
+The connection is completed, for example after a DHCP lease received.
+* `wait: <seconds>` is a timeout that enables you to decide how long you give the device to
+activate. The default is using a suitable timeout. Note that the `wait` option is
+only supported by NetworkManager.
```
pcahyna: The original version had
> **TODO** `wait` different from zero is not yet implemented.

@tyll @thom311 Is it intentional that it has disappeared?

thom311: Yes: https://github.com/linux-system-roles/network/pull/86#discussion_r220863356

README.md
```diff
@@ -273,96 +354,44 @@ network_connections:
...
+    mac: 00:00:5e:00:53:5d
+    ip:
+      dhcp4: yes
+```
 
-      address:
-        - 192.0.2.3/24
-        - 198.51.100.3/26
-        - 2001:db8::80/7
+Setting an absent connection profile:
```
pcahyna: It should read: "Deleting a connection profile", no? "Setting an absent connection profile" makes no sense IMO.

README.md
```diff
@@ -483,102 +490,128 @@ network_connections:
...
-as well, but again it is not guaranteed that plain initscripts can handle these ifcfg files
+as well, but it is not guaranteed that plain initscripts can handle these ifcfg files
 after disabling the NetworkManager service.
+
+Limitations
+-----------
+
+Ansible usually works via the network, for example via SSH. There are some limitations to be considered:
+
+The `network` role does not  support how to bootstrap networking configuration. One
```
pcahyna: This is less intelligible than the original version.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
 
-The role updates or creates all connection profiles on the target system as
-specified in the `network_connections` variable. Therefore, the role will
-remove settings from the specified profiles if the settings are only present on
-the system but not in the `network_connections` variable. The following
-exceptions apply:
+Variables
+---------
+The `network` role is configured via variables starting  with  `network_` as the name prefix.
+List of required variables:
```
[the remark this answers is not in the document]
yontalcar: Neither of them is required, right? So I'll just delete "required".

README.md
yontalcar: What about this:

> As Ansible usually works via the network, for example via SSH, there are some limitations to be considered:
>
> The `network` role does not support bootstraping networking configuration. One…

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
 
-* For profiles that only contain a `state` setting, the role will only activate
-  or deactivate the connection without changing its configuration.
+* `network_provider` - The `network_provider` variable allows to set a specific
+  provider (`nm` or `initscripts`) . Setting it to `network_provider_os_default`,
+  the provider is set depending on the operating system. This is usually `nm`
```
pcahyna: This is confusing. It does not say that `network_provider_os_default` is a fact set by the role. But this problem existed in the original version as well

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+The `bridge`, `bond`, `team` device types work similar. Note that `team` is not supported in RHEL6 kernels.
+
+For slaves, the `slave_type` and `master` properties must be set. Note that slaves should not have `ip` settings.
+
+The `master` refers to the `name` of a profile in the ansible
+Playbook. It is neither an interface-name nor a connection-id of
+NetworkManager.
+
+- For `NetworkManager`, `master` corresponds to the `connection.uuid`
+  of the corresponding profile.
```
pcahyna: This is confusing. The original version

> - For NetworkManager, `master` will be converted to the `connection.uuid` of the corresponding profile.

was better.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+
+- For `initscripts`, `master` determines the `DEVICE` from the corresponding
+  ifcfg file.
```
pcahyna: The previous version
> For initscripts, the master is looked up as the `DEVICE` from the corresponding ifcfg file.

was probably better.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+network_provider: nm
+network_connections:
+  - name: eth0
+    #...
+```
 
-### `name`
+Options
+-------
+The `network_connections` variable is a list of dictionaries that includes the following options.
```
pcahyna: "that include" - "dictionaries" is the subject, not "list".

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+List of required options:
```
pcahyna: Only `name` seems to be required, others should be optional.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+Note that if `persistent_state` is `present` and the connection profile contains
+the `type` option, the profile will be created or updated. If the connection profile is
+incomplete (no `type` option), the behavior is undefined. Also, the `present` value
+does not directly result in a change in the network configuration. If the `state` option
+is not set to `up`, the profile is only created or modified, not activated.
+
+#### `persistent_state: absent`
+
+The `absent` value ensures that the profile is not present on the target host. For
+example, if a profile with `name` `eth0` exists, it will be deleted. In this case:
```
pcahyna: The matching example configuration was moved to a different place, so this part become unintelligible (it is not clear why one mentions `eth0` here).

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
-that at least version 1.2 of NetworkManager's API is available. For
-`initscripts`, it requires the legacy network service as commonly available on
-Fedora/RHEL.
-
-For each host a list of networking profiles can be configured via the
-`network_connections` variable.
+distribution. However, note that either `nm` or `initscripts` is not tied to a certain
+distribution. The `network` role works everywhere the required API is available.
+This means that `nm` requires at least NetworkManager's API version 1.2 available.
+For `initscripts`, the legacy network service is required as used in Fedora or RHEL.
```
pcahyna: This does not make sense in the new version, because we start talking about profiles below without introducing them (after the paragraph above was deleted or moved to a different place).

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
-But it is also possible to create generic profiles, by creating for example a
+Note that the `network` role primarily operates on networking profiles (connections) and
+not on devices, but it uses the profile name by default as the interface name.
+It is also possible to create generic profiles, by creating for example a
 profile with a certain IP configuration without activating the profile. To
-apply the configuration to the actual networking interface, a command like
-`nmcli` needs to be used on the target system.
+apply the configuration to the actual networking interface, use the `nmcli`
+commands on the target system. For each host, a list of networking profiles can
+be configured via the `network_connections` variable.
```
pcahyna: I suggest to move the last phrase back.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+The `absent` value ensures that the profile is not present on the target host. For
+example, if a profile with `name` `eth0` exists, it will be deleted. In this case:
+
+- `NetworkManager` deletes all connection profiles with the corresponding `connection.id`.
+    Deleting a profile usually does not change the current networking configuration, unless
+    the profile was currently activated on a device. Deleting the currently
+    active connection profile disconnects the device. That makes the device eligible
+    to autoconnect another connection (for more details, see [rh#1401515](https://bugzilla.redhat.com/show_bug.cgi?id=1401515)).
+
+- `initscripts` deletes the ifcfg file in most cases with no impact on the system unless a component relies on the sysconfig directory.
```
pcahyna: The original version had 
> For `initscripts`, it has no side-effect, unless some component is watching the sysconfig directory.

What was meant is that the runtime configuration is not changed, unless there is some service watching (not relying on) the sysconfig directory. The meaning is now garbled.

README.md
```diff
@@ -15,251 +18,329 @@ The role can be used to configure:
...
+role.
 
-- For NetworkManager, note the new connection profile is created with
-  `autoconnect` turned on by default. Thus, NetworkManager may very well decide
-  right away to activate the new profile on a currently disconnected device.
-  ([rh#1401515](https://bugzilla.redhat.com/show_bug.cgi?id=1401515)).
 
 ### `autoconnect`
 
+For NetworkManager, the new connection profile is created when the `autoconnect`
```
pcahyna: This is wrong. The new profile is not created when the `autoconnect` option is enabled. It is created with the `autoconnect` property enabled.

README.md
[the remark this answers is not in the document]
pcahyna: I think `network_connections` is sort of required (the role should work without it, but does not do anything interesting).

tyll:
I just realized that the name option is not strictly required because there is one special case where it is not. This is when using `persistent_state: absent` to remove all other profiles except the ones that are specified.

README.md
tyll: It works for me to ignore state:down if the profile does not exist. The last discussion we had about this lead to make this a configuration setting: https://github.com/linux-system-roles/network/issues/70

thom311: Yes, this covers the case where the profile does not exist. There is also the case, where the profile is already down. For `nm` provider, calling `state: down` on such a profile leaves the system unchanged. That is partly also the case for `initscripts` provider, except that the role does only very badly understand whether the profile is already down or not (https://github.com/linux-system-roles/network/blob/master/library/network_connections.py#L581).

Anyway, the sentence "always changes the system" is not correct.

PR #88, opened 2018-10-15
Fix Fedora 29 support

tyll:


tyll:
[citest bad]

pcahyna:
84fe985 is because of Ansible 2.7, not because of Fedora 29, right?

tyll:
https://github.com/linux-system-roles/network/commit/84fe98508e82cbe9cfa83ffe2f0df9a2aa2602b1 is not required for Fedora 29, since I had to use `is version()` for the new commits, I adjusted it here, too.

PR #90, opened 2018-10-15
Fix Ansible 2.7 support

tyll:


pcahyna:
Fedora 29 tests are failing because of missing #88 ?

tyll:
Yes, the failure in CI is expected because of #88 - I will rebase and then it should be gone.

PR #91, opened 2018-11-16
Tests: Add names for other provider tasks/playbooks

tyll:
This makes it easier to figure out what is going on from the logs


tyll:
[citest bad]

PR #94, opened 2019-01-08
tasks: Do not log full service facts

tyll:
The service facts create too much noise in test logs, therefore hide
them.


tyll:
[citest bad]

PR #95, opened 2019-01-09
Make connections volatile instead of removing them

tyll:


module_utils/network_lsr/utils.py
```diff
@@ -126,6 +126,29 @@ def GMainLoop_iterate_all(cls):
...
+        def callback(source_object, res, user_data):
+            success = None
+            try:
+                success = getattr(source_object, finish_method)(res)
+            except Exception as e:
+                if Util.error_is_cancelled(e):
+                    return
+                user_data["error"] = str(e)
+            user_data["success"] = success
+            Util.GMainLoop().quit()
```
thom311: `create_callback()` is a class method. I think that is preferable over a static-method if the method references itself (its class). But here you still use `Util` directly. Either make the method `@staticmethod` or use `cls` instead of `Util`. I prefer the latter, because I find it a bit ugly that a static-method would access itself again via the `Util` name.

library/network_connections.py
```diff
@@ -1010,23 +1007,71 @@ def delete_cb(connection, result, cb_args):
...
+        c_uuid = connection.get_uuid()
+        gone = self.wait_till_connection_is_gone(c_uuid)
+        if not gone:
+            raise MyError(
+                "connection %s was supposedly deleted successfully, but it's still here"
+                % (c_uuid)
+            )
+
+    def volatilize_connection(self, connection, timeout=10):
+        update2_cb = Util.create_callback("update2_finish")
```
thom311: this variable is only used at one place. I think it's preferable to just create and use the callback at the place, instead of assigning it first to another variable.

library/network_connections.py
```diff
@@ -1010,23 +1007,71 @@ def delete_cb(connection, result, cb_args):
...
+            cancellable,
+            update2_cb,
+            cb_args,
+        )
+
+        if not Util.GMainLoop_run(timeout):
+            cancellable.cancel()
+            raise MyError("failure to volatilize connection: %s" % ("timeout"))
+
+        # Do not check of success if ther connection does not exist anymore
```
thom311: maybe here issue a `Util.GMainLoop_iterate_all()`? `GMainLoop_run()` will quit once the callback returns, but there might be still other D-Bus messages pending. I think we should process them all before checking `if connection not in self.connection_list()`.

library/network_connections.py
```diff
@@ -983,18 +984,14 @@ def update_cb(connection, result, cb_args):
...
-            except Exception as e:
-                if Util.error_is_cancelled(e):
-                    return
-                cb_args["error"] = str(e)
-            cb_args["success"] = success
-            Util.GMainLoop().quit()
+        if "update2" in dir(connection):
+            return self.volatilize_connection(connection, timeout)
+
+        delete_cb = Util.create_callback("delete_finish")
```
thom311: also here. Drop variable, like `update2_cb` below

module_utils/network_lsr/utils.py
tyll: good catch, will use `cls`.

library/network_connections.py
tyll: > ```
> +        gone = self.wait_till_connection_is_gone(c_uuid)
> +        if not gone:
> ```
> same here. Do you think such variables help readability? I would think the contrary, but that's of course a matter of opinion.

To me it looks easier to read because it is smaller pieces to parse at a time.

tyll:
Thank you for the review! I merged your commits and added a commit to rename the `no_timeout` variable.

thom311:
> Thank you for the review! I merged your commits and added a commit to rename the no_timeout variable.

Thanks. I am fine with renaming. Please rebase and squash the commits as you see fit.

module_utils/network_lsr/utils.py
```diff
@@ -126,6 +126,29 @@ def GMainLoop_iterate_all(cls):
...
+        quit the GMainLoop
+
+        :param finish_method str: Name of the finish method to call from the
+        source object in the callback
+        """
+
+        def callback(source_object, res, user_data):
+            success = None
+            try:
+                success = getattr(source_object, finish_method)(res)
```
thom311: Maybe resolve the closure first, and not during the callback? Then, on invalid function name, the exception happens synchronously during `create_callback()`. Also, I feels slightly nicer to resolve the function name at the moment when creating the callback, then later.

tyll: Yes, it would be better but I do not see that it is possible since `source_object` is not defined in `create_callback()`. Let me know if I am missing something.

thom311: you are right of course. And having it available would make the function more cumbersome to use. Good then!

tyll:
[citest bad]

tyll:
[citest bad]

pcahyna:
is there a test for this feature?

PR #100, opened 2019-01-24
Routes: add option to specify network interface

silanea:
Option was missing, so I added it. It works for me, and to the best of my understanding of both the Python script and the routing commands themselves my changes should not break anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ The IP configuration supports the following options:
 
     Static route configuration can be specified via a list of routes given in the `route`
     option. The default value is an empty list. Each route is a dictionary with the following
-    entries: `network`, `prefix`, `gateway` and `metric`. `network` and `prefix` specify
+    entries: `network`, `prefix`, `gateway`, `device` and `metric`. `network` and `prefix` specify
     the destination network.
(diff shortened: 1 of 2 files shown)


tyll:
Hi, thank you for your contribution. It seems that the indentation is not quit correct (see the travis status). Also the role is meant to support NetworkManager as well but the current change is only for the initscripts backend. Would it be possible for you to look at the NetworkManager support as well?

thom311:
In NetworkManager, static routes cannot specify a device. The device is always the one one which the route is activated. It would be an interesting feature, to support configuring routes on a profile, that are then added to another device than the one on which the profile activates. Anyway, not supported.

For initscripts, `/etc/sysconfig/network-scripts/ifup-routes` just calls `ip route add`, and hence of course the ifcfg requires the device to be specified. Initscripts don't automatically add that argument.

But I don't think that it makes sense to specify a device for each route. At least, it seems misuse of initscripts to configure routes in a profile, that then are added to another device. Instead, I think the role can only write routes to the ifcfg file, if the device name is also specified via `interface_name`. Then, there is no need to specify the device per-route.

PR #103, opened 2019-02-15
README: Mention mtu option

### LINKED ISSUE #98: document MTU
There seem to be support for MTU in the code (but see #19), but I can't find it in the README.

tyll:
This fixes #98.

---
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/103)

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,12 @@ profile.
 
 - For `initscripts`,  `mac` is the currently configured MAC address of the device (`HWADDR`).
 
+### `mtu`
+
+The `mtu` option denotes the maximum transmission unit for the profile's
+device. The maximum value depends on the device. For virtual devices, the
+maximum value of the `mtu` option depends on the underlying device.
+
 ### `interface_name`
 
 For the `ethernet` and `infiniband`  types, the `interface_name` option restricts the profile to


tyll:
[citest bad]

README.md
```diff
@@ -249,6 +249,12 @@ profile.
 
 - For `initscripts`,  `mac` is the currently configured MAC address of the device (`HWADDR`).
 
+### `mtu`
+
+The `mtu` denotes the maxium transmission unit for the profile's device. Please
```
pcahyna: maximum

tyll: thx, fixed

tyll:
[citest bad]

tyll:
@pcahyna I added `option` as well.

README.md
```diff
@@ -249,6 +249,12 @@ profile.
 
 - For `initscripts`,  `mac` is the currently configured MAC address of the device (`HWADDR`).
 
+### `mtu`
+
+The `mtu` option denotes the maximum transmission unit for the profile's
+device. Please note, if a profile depends on another profile, the value of the
+`mtu` option cannot be higher than the value for the parent's device `mtu`.
```
pcahyna: what about writing it a bit differently:
"Please note, for logical devices (for example those of type `bond` or `vlan`) the value of the `mtu` option cannot be higher than the value for the underlying physical device `mtu`."
The reason is that the relation of dependency is different for `bond` and for `vlan` - in `bond` the slaves (physical devices) refer to (depends on) their `master` (logical device) while for `vlan` the vlan (logical device) refers to (depends on) its physical device.

pcahyna: that said, maybe this note makes really sense only for VLANs - can a bond have a different MTU than its slaves?

tyll:
[citest bad]

tyll:
[citest bad]

tyll:
I reworded it, so it is not too specific and not to vague. IMHO the documentation should describe the role and not generic networking concepts or restrictions.

pcahyna:
:+1: to the change of wording which removes the notion of parent or dependency and talks about underlying devices. I think it is clear now. lgtm.

PR #104, opened 2019-04-15
Initscripts workarounds

tyll:
Add two fixes/workarounds in case network.service from initscripts has issues. See commit messages for details.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -38,6 +38,12 @@
   when:
     -  network_provider == "initscripts"
 
+- name: Ensure initscripts network file dependency is present
+  copy:
+    dest: /etc/sysconfig/network
+    content: "# Created by network system role"
+    force: false
+
 - name: Configure networking connection profiles
   network_connections:
       provider:            "{{ network_provider           | mandatory }}"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -23,11 +23,20 @@
   when:
     - not network_packages is subset(ansible_facts.packages.keys())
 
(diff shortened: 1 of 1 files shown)


tasks/main.yml
```diff
@@ -23,11 +23,26 @@
...
       enabled: yes
+  when:
+    - network_provider == "nm"
+
+- name: Enable network service
+  service:
+      name: "{{ network_service_name }}"
+      enabled: yes
+  when:
+    -  network_provider == "initscripts"
```
pcahyna: When will it be actually started?

pcahyna: I mean, don't you still need to start it to apply the new settings?

tyll: No, the new settings are applied by the module (it issues `ifup` directly). The service only ensures that the profiles on disk are applied when the system starts (or when the service is started).

tyll:
Thank you for the fast review!

pcahyna:
Will it (the network service) be really started after boot, if there are network profiles for unknown devices? (Aren't we just postponing the problem to a later moment?)

tyll:
> Will it (the network service) be really started after boot, if there are network profiles for unknown devices? (Aren't we just postponing the problem to a later moment?)

With the workaround users can specify `peristent_state: absent` as a profile to remove all unmentioned profiles to make it more likely that this will not be a problem.

Also even if the network service failed to start from systemd perspective, it will still try to bring up all the profiles. So the system will still have the network access that is specified in the profiles.

PR #105, opened 2019-04-16
More container improvements/cleanups

tyll:
With these commits, the tests now pass

```
for f in tests/tests_*.yml ; do TEST_DOCKER_EXTRA_ARGS="-it --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro" TEST_SUBJECTS=docker:milcom/centos7-systemd ansible-playbook -i $HOME/standard-test-roles/inventory/standard-inventory-docker  $f &> $f.log & :; done
```

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # get service facts, used in defaults/main.yml
+---
 - name: Check which services are running
   service_facts:
   no_log: true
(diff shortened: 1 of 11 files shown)


tyll:
[citest bad]

tyll:
> ITYM /etc/sysconfig/network-scripts (in the commit message).

Yes, unfortunately it is already merged...

> Shouldn't you then revert [5273f55](https://github.com/linux-system-roles/network/commit/5273f55166994f4baa39835977d02296a64ab96a) ?

I considered this as well but IMHO the new implementation is still better because it is faster (no need to gather all the facts that often anymore) and is simpler.

PR #107, opened 2019-04-17
Configure Molecule in Travis CI

vcrhonek:
Configures Molecule to do basic checks (lint/syntax, run, idempotence, Ansible check mode functionality) of the role with default parameters (tests/tests_default.yml) using Travis CI and centos6/7 docker containers.

Because Travis was already used by network role, please check mainly .travis.yml, whether it makes sense to incorporate molecule into it in this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,9 @@ matrix:
     - python: 2.6
       dist: trusty
     - python: 2.7
-    - python: 3.6
+    - python: 3.5
       env: aptpkgs=python3-selinux
+    - python: 3.6
     - python: 3.7
     - python: 3.7-dev
     - python: 3.8-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,17 @@ matrix:
       dist: trusty
     - python: 2.7
     - python: 3.6
+      env: aptpkgs=python3-selinux
     - python: 3.7
     - python: 3.7-dev
     - python: 3.8-dev
     # - python: nightly
+
+services:
+  - docker
+
+before_install:
+  - if [ -n "${aptpkgs}" ]; then sudo apt-get install -y python3-selinux; fi
 
 install:
   - pip install tox tox-travis
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -1,0 +1,38 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+lint:
+  name: yamllint
+  options:
+    config-file: molecule/default/yamllint.yml
+platforms:
+  - name: centos-6
+    image: linuxsystemroles/centos-6
+    privileged: true
+  - name: centos-7
+    image: linuxsystemroles/centos-7
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
+provisioner:
+  name: ansible
+  log: true
+  lint:
+    name: ansible-lint
+  playbooks:
+    converge: ../../tests/tests_default.yml
+scenario:
+  name: default
+  test_sequence:
+    - destroy
+    - create
+    - converge
+    - idempotence
+    - check
+    - destroy
+verifier:
+  name: testinfra
+  lint:
+    name: flake8
--- a/tox.ini
+++ b/tox.ini
@@ -169,6 +169,7 @@ max-line-length = 88
 python =
   2.6: py26
   2.7: py27,coveralls,flake8,pylint
-  3.6: py36,black,ensure_non_running_provider,molecule_lint,molecule_syntax,molecule_test
+  3.5: molecule_lint,molecule_syntax,molecule_test
+  3.6: py36,black,ensure_non_running_provider
   3.7: py37
   3.8: py38
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,11 @@ deps =
     molecule_{lint,syntax,test}: docker
     molecule_{lint,syntax,test}: jmespath
     molecule_{lint,syntax,test}: molecule
-    molecule_{lint,syntax,test}: selinux
+    # The selinux pypi shim does not work with Ubuntu (as used by Travis), yet.
+    # Therefore use a fork with Ubuntu support. This can be changed once the
+    # update is available on PyPi.
+    # molecule_{lint,syntax,test}: selinux
+    molecule_{lint,syntax,test}: git+https://github.com/tyll/selinux-pypi-shim@fulllocation
 
 [base]
 passenv = *
--- a/tox.ini
+++ b/tox.ini
@@ -4,12 +4,16 @@ skipsdist = true
 skip_missing_interpreters = True
 
 [testenv]
-basepython = python2.7
+basepython = python3
 deps =
     py{26,27,36,37,38}: pytest-cov
     py{27,36,37,38}: pytest>=3.5.1
     py{26,27}: mock
     py26: pytest
+    molecule_{lint,syntax,test}: docker
+    molecule_{lint,syntax,test}: jmespath
+    molecule_{lint,syntax,test}: molecule
+    molecule_{lint,syntax,test}: selinux
 
 [base]
 passenv = *
@@ -21,17 +25,16 @@ covtarget = {toxinidir}/library --cov {toxinidir}/module_utils
 pytesttarget = .
 
 [testenv:black]
-basepython = python3.6
 deps = black
 
 commands = black --check --diff --include "^[^.].*\.py$" .
 
 [testenv:py26]
+basepython = python2.6
 passenv = {[base]passenv}
 setenv =
     {[base]setenv}
 changedir = {[base]changedir}
-basepython = python2.6
 commands =
     pytest \
         --durations=5 \
@@ -41,11 +44,11 @@ commands =
         {[base]pytesttarget}
 
 [testenv:py27]
+basepython = python2.7
 passenv = {[base]passenv}
 setenv =
     {[base]setenv}
 changedir = {[base]changedir}
-basepython = python2.7
 commands =
     pytest \
         --durations=5 \
@@ -55,11 +58,11 @@ commands =
         {[base]pytesttarget}
 
 [testenv:py36]
+basepython = python3.6
 passenv = {[base]passenv}
 setenv =
     {[base]setenv}
 changedir = {[base]changedir}
-basepython = python3.6
 commands =
     pytest \
         --durations=5 \
@@ -69,11 +72,11 @@ commands =
         {[base]pytesttarget}
 
 [testenv:py37]
+basepython = python3.7
 passenv = {[base]passenv}
 setenv =
     {[base]setenv}
 changedir = {[base]changedir}
-basepython = python3.7
 commands =
     pytest \
         --durations=5 \
@@ -112,6 +115,7 @@ commands =
         tests/test_network_connections.py
 
 [testenv:flake8]
+basepython = python2.7
 deps =
     flake8>=3.5
 whitelist_externals = flake8
@@ -120,6 +124,7 @@ commands=
         .
 
 [testenv:coveralls]
+basepython = python2.7
 passenv = TRAVIS TRAVIS_*
 deps =
     coveralls
@@ -128,11 +133,22 @@ commands =
     coveralls
 
 [testenv:ensure_non_running_provider]
-basepython = python3.6
 deps =
     PyYAML
 changedir = {toxinidir}/tests
 commands = {toxinidir}/tests/ensure_non_running_provider.py
+
+[testenv:molecule_lint]
+commands_pre =
+    molecule --version
+    ansible --version
+commands = molecule {posargs} lint
+
+[testenv:molecule_syntax]
+commands = molecule {posargs} syntax
+
+[testenv:molecule_test]
+commands = molecule {posargs} test
 
 [pytest]
 addopts = -rxs
@@ -153,6 +169,6 @@ max-line-length = 88
 python =
   2.6: py26
   2.7: py27,coveralls,flake8,pylint
-  3.6: py36,black,ensure_non_running_provider
+  3.6: py36,black,ensure_non_running_provider,molecule_lint,molecule_syntax,molecule_test
   3.7: py37
   3.8: py38
--- a/molecule/default/Dockerfile.j2
+++ b/molecule/default/Dockerfile.j2
@@ -1,0 +1,14 @@
+# Molecule managed
+
+{% if item.registry is defined %}
+FROM {{ item.registry.url }}/{{ item.image }}
+{% else %}
+FROM {{ item.image }}
+{% endif %}
+
+RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
+    elif [ $(command -v yum) ]; then yum makecache fast && yum install -y python sudo yum-plugin-ovl bash && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
+    elif [ $(command -v zypper) ]; then zypper refresh && zypper install -y python sudo bash python-xml && zypper clean -a; \
+    elif [ $(command -v apk) ]; then apk update && apk add --no-cache python sudo bash ca-certificates; \
+    elif [ $(command -v xbps-install) ]; then xbps-install -Syu && xbps-install -y python sudo bash ca-certificates && xbps-remove -O; fi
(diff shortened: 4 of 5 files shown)


pcahyna:
`Could not find a version that satisfies the requirement flake8<4,>=3.6.0 (from molecule) (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.3.1, 1.4, 1.5, 1.6, 1.6.1, 1.6.2, 1.7.0, 2.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.6.0, 2.6.1, 2.6.2, 3.0.0b1, 3.0.0b2, 3.0.0, 3.0.1, 3.0.2.dev0, 3.0.2.dev1, 3.0.2, 3.0.3, 3.0.4, 3.1.0, 3.1.1, 3.2.0, 3.2.1, 3.3.0.dev0, 3.3.0, 3.4.0, 3.4.1, 3.5.0)
No matching distribution found for flake8<4,>=3.6.0 (from molecule)
The command "pip install tox tox-travis molecule docker" failed and exited with 1 during .`

vcrhonek:
> Travis fails with: `Could not find a version that satisfies the requirement flake8<4,>=3.6.0 (from molecule) (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.3.1, 1.4, 1.5, 1.6, 1.6.1, 1.6.2, 1.7.0, 2.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.6.0, 2.6.1, 2.6.2, 3.0.0b1, 3.0.0b2, 3.0.0, 3.0.1, 3.0.2.dev0, 3.0.2.dev1, 3.0.2, 3.0.3, 3.0.4, 3.1.0, 3.1.1, 3.2.0, 3.2.1, 3.3.0.dev0, 3.3.0, 3.4.0, 3.4.1, 3.5.0) No matching distribution found for flake8<4,>=3.6.0 (from molecule) The command "pip install tox tox-travis molecule docker" failed and exited with 1 during .`

I see, that happens only with Python 2.6, other jobs are fine. But it causes skip of tox testing which is unfortunate.

tyll:
> > Travis fails with: `Could not find a version that satisfies the requirement flake8<4,>=3.6.0 (from molecule) (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.3.1, 1.4, 1.5, 1.6, 1.6.1, 1.6.2, 1.7.0, 2.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.6.0, 2.6.1, 2.6.2, 3.0.0b1, 3.0.0b2, 3.0.0, 3.0.1, 3.0.2.dev0, 3.0.2.dev1, 3.0.2, 3.0.3, 3.0.4, 3.1.0, 3.1.1, 3.2.0, 3.2.1, 3.3.0.dev0, 3.3.0, 3.4.0, 3.4.1, 3.5.0) No matching distribution found for flake8<4,>=3.6.0 (from molecule) The command "pip install tox tox-travis molecule docker" failed and exited with 1 during .`
> 
> I see, that happens only with Python 2.6, other jobs are fine. But it causes skip of tox testing which is unfortunate.

IMHO it would be nicer to add the molecule commands to the tox config and  install the dependencies there. Then we can run the tests only for certain python versions. I will take a look.

vcrhonek:
> > > Travis fails with: `Could not find a version that satisfies the requirement flake8<4,>=3.6.0 (from molecule) (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.3.1, 1.4, 1.5, 1.6, 1.6.1, 1.6.2, 1.7.0, 2.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.4.0, 2.4.1, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.6.0, 2.6.1, 2.6.2, 3.0.0b1, 3.0.0b2, 3.0.0, 3.0.1, 3.0.2.dev0, 3.0.2.dev1, 3.0.2, 3.0.3, 3.0.4, 3.1.0, 3.1.1, 3.2.0, 3.2.1, 3.3.0.dev0, 3.3.0, 3.4.0, 3.4.1, 3.5.0) No matching distribution found for flake8<4,>=3.6.0 (from molecule) The command "pip install tox tox-travis molecule docker" failed and exited with 1 during .`
> > 
> > 
> > I see, that happens only with Python 2.6, other jobs are fine. But it causes skip of tox testing which is unfortunate.
> 
> IMHO it would be nicer to add the molecule commands to the tox config and install the dependencies there. Then we can run the tests only for certain python versions. I will take a look.

Thanks, that would be great as I have no experience with tox yet.

tyll:
I tested this locally and noticed that `molecule test` fails with this error message:

```
    failed: [localhost] (item={'name': 'centos7', 'image': 'vcrhonek/centos7-systemd', 'volumes': ['/sys/fs/cgroup:/sys/fs/cgroup:ro'], 'privileged': True}) => {"changed": false, "checksum": "0a234f1e797e8263d3a567b965259ff9bc2bf42b", "item": {"image": "vcrhonek/centos7-systemd", "name": "centos7", "privileged": true, "volumes": ["/sys/fs/cgroup:/sys/fs/cgroup:ro"]}, "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}
```

tyll:
My change for tox support are here: https://github.com/tyll/network/tree/molecule

Also this change seems helpful to make the log output of molecule test useful:

```
diff --git a/molecule/default/molecule.yml b/molecule/default/molecule.yml
index 486238a..666d732 100644
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -18,6 +18,7 @@ platforms:
     privileged: true
 provisioner:
   name: ansible
+  log: true
   lint:
     name: ansible-lint
   playbooks:
```

tyll:
@vcrhonek how is the Dockerfile.j2 template used? When I change it to install libselinux-python, it still fails. When I remove it, it does not behave differently... Seems like it is not needed and only confusing.

vcrhonek:
> @vcrhonek how is the Dockerfile.j2 template used? When I change it to install libselinux-python, it still fails.
> Removing makes a difference, though...

It is used to build docker container used as a test environment for role. We can get rid of it and install everything right into the docker images from molecule.yml (vcrhonek/centos7-systemd for example). See:
https://www.jeffgeerling.com/blog/2018/testing-your-ansible-roles-molecule
(search for Use pre-built Docker images with Molecule)

I don't hit libselinux-python issue locally though (without tox).

tyll:
> I don't hit libselinux-python issue locally though (without tox).

After I found out about `molecule --debug test` I noticed that it fails because the selinux python module is not installed on the host instead of not being available in the container. I will give it another try and report back.

tyll:
I got it working now - the centos6 image is not supported as-is and exposes an interesting problem.
It does not contain the initscripts package which contains the `service` command and service files. Without it, the service facts from Ansible do not return anything. This breaks the detection of which provider is currently active. It could be worked around with a patch like

```
diff --git a/defaults/main.yml b/defaults/main.yml
index 1a11a96..a0d72a8 100644
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,7 +6,7 @@ network_connections: []
 network_provider_os_default: "{{ 'initscripts' if ansible_distribution in ['RedHat', 'CentOS'] and ansible_distribution_major_version is version('7', '<') else 'nm' }}"
 # If NetworkManager.service is running, assume that 'nm' is currently in-use,
 # otherwise initscripts
-network_provider_current: "{{ 'nm' if 'NetworkManager.service' in ansible_facts.services and ansible_facts.services['NetworkManager.service']['state'] == 'running' else 'initscripts' }}"
+network_provider_current: "{{ 'nm' if 'NetworkManager.service' in ansible_facts.services | default([]) and ansible_facts.services['NetworkManager.service']['state'] == 'running' else 'initscripts' }}"
 # Default to the auto-detected value
 network_provider: "{{ network_provider_current }}"
```

But this would also mask any other errors with service facts. To me this seems to be an artificial problem with this docker image that users would not hit in the wild. Therefore I think the best approach would be to use a docker image that contains initscripts already for CentOS 6.
It might even be a good idea to also have two CentOS 7 images, one with NetworkManager running and one without to cover both cases and have a test that checks the provider detection is working as intended. What are your opinions?

vcrhonek:
> But this would also mask any other errors with service facts. To me this seems to be an artificial problem with this docker image that users would not hit in the wild. Therefore I think the best approach would be to use a docker image that contains initscripts already for CentOS 6.

I agree with you.

> It might even be a good idea to also have two CentOS 7 images, one with NetworkManager running and one without to cover both cases and have a test that checks the provider detection is working as intended. What are your opinions?

In my opinion this makes very good sense for network role. But probably not for other linux-system-roles - if the idea is still to use the same docker images and molecule configuration for all of them.

tyll:
> > But this would also mask any other errors with service facts. To me this seems to be an artificial problem with this docker image that users would not hit in the wild. Therefore I think the best approach would be to use a docker image that contains initscripts already for CentOS 6.
> 
> I agree with you.

Will you create a custom container image for this or how do we proceed?

> > It might even be a good idea to also have two CentOS 7 images, one with NetworkManager running and one without to cover both cases and have a test that checks the provider detection is working as intended. What are your opinions?
> 
> In my opinion this makes very good sense for network role. But probably not for other linux-system-roles - if the idea is still to use the same docker images and molecule configuration for all of them.

For other roles with multiple providers it might also make sense to test with different initial providers, for example for timesync, I guess. So maybe we need some variations anyhow,

molecule/default/molecule.yml
```diff
@@ -0,0 +1,37 @@
...
+lint:
+  name: yamllint
+  options:
+    config-file: molecule/default/yamllint.yml
+platforms:
+  - name: centos6
+    image: centos:6
+    privileged: true
+  - name: centos7
+    image: vcrhonek/centos7-systemd
```
tyll: Please add the image the to linuxsystemroles docker organization and use it from there.

pcahyna:
> > > It might even be a good idea to also have two CentOS 7 images, one with NetworkManager running and one without to cover both cases and have a test that checks the provider detection is working as intended. What are your opinions?
> > 
> > 
> > In my opinion this makes very good sense for network role. But probably not for other linux-system-roles - if the idea is still to use the same docker images and molecule configuration for all of them.
> 
> For other roles with multiple providers it might also make sense to test with different initial providers, for example for timesync, I guess. So maybe we need some variations anyhow,

I think you are overthinking it. We found many limitations of Molecule testing, so we want to use it only when it is reasonably simple to do it to collect the low-hanging fruits like static linting and basic check mode/idempotency checks. Configuring more complex checks, such as images with other providers, is IMO above the "reasonably simple" threshold. Anyway, using different system versions (centos-6, centos-7, centos-8) will lead to testing of different providers without any additional effort, because different systems have different defaults, won't it?

vcrhonek:
> 
> Will you create a custom container image for this or how do we proceed?
> 

I've created  the image for CentOS 6 with initscripts and added it along with CentOS 7 image to  linuxsystemroles docker organization. Where do you think should Dockerfiles for them live?

I'm going to take a look at your proposed changes for tox support, thanks.

pcahyna:
This can't be merged because of a failure:
> You need to install \"jmespath\" prior to running json_query filter 
One needs to install jmespath on the control host: (https://github.com/tyll/network/blob/a1a9450b4b33cf6d54f4bb26b37176efb5d60e8c/tox.ini#L14 )

and python2.6 tests errored: https://github.com/linux-system-roles/network/pull/107#issuecomment-483993294

tyll:
I moved my cleanup commits to #108 and split them properly.

pcahyna:
one Travis CI test still fails:
ERROR! Unexpected Exception, this is probably a bug: Failed to detect selinux python bindings at /usr/lib64/python3.6/site-packages
This is the Python3.6 test, others are passing.

.travis.yml
```diff
@@ -1,3 +1,4 @@
+---
```
pcahyna: see https://github.com/linux-system-roles/network/pull/108/files#r278538735

.travis.yml
```diff
@@ -7,16 +7,15 @@ python:
...
-  - pip install tox tox-travis molecule docker
+  - pip install tox tox-travis
 
 script:
   - tox
-  - molecule --version
-  - ansible --version
-  - molecule lint
-  - molecule syntax
-  - molecule test
```
pcahyna: The two blocks above seem to revert parts of b83d15f09b91fe92cde761f91f154fc4f1b13eb3, can you squash those two commits to minimize the diffs and reduce the chance of conflicts, please?

.travis.yml
```diff
@@ -7,16 +7,15 @@ python:
   - 2.7
   - 3.6
   - 3.7-dev
+
 services:
   - docker
```
pcahyna: How does Travis start Docker, if Docker installation has been moved to Tox? (below)

pcahyna:
> > > But this would also mask any other errors with service facts. To me this seems to be an artificial problem with this docker image that users would not hit in the wild. Therefore I think the best approach would be to use a docker image that contains initscripts already for CentOS 6.
> > 
> > 
> > I agree with you.
> 
> Will you create a custom container image for this or how do we proceed?

An alternative to using our own image would be to build it in Travis from a vanilla image.

https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile

Would that work?

.travis.yml
tyll: tox installs only the docker python bindings, not the service itself.

.travis.yml
tyll: resolved this.

.travis.yml
tyll: I moved the changes to .travis.yml in one squashed commit.

tyll:
> An alternative to using our own image would be to build it in Travis from a vanilla image.
> 
> https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile
> 
> Would that work?

It will take a little bit more ressources because it requires to always rebuild the image. Also it might be a bit more complicated since we need to build the two images a little bit more different (one with initscripts, the other with systemd).

tyll:
> > Will you create a custom container image for this or how do we proceed?
> 
> I've created the image for CentOS 6 with initscripts and added it along with CentOS 7 image to linuxsystemroles docker organization. Where do you think should Dockerfiles for them live?

How about in a molecule repo that contains a sample/reference molecule configuration for all/new roles.

tyll:
Regarding the selinux failure: Ubuntu seems to be special here compared to Fedora/CentOS... I have some changes prepared but I need to verify/upstream them. Would be great to merge #108 already, so that those changes are already done. We can also skip the travis changes here for now because the other changes already work when running it locally on Fedora.

tyll:
It seems to be ready now to be merged. We need to update the tox config in case https://github.com/ssbarnea/selinux/pull/3 gets merged and published.

tyll:
Updated the tox config since upstream merged my PR.

tyll:
> Updated the tox config since upstream merged my PR.

Changed back to the custom selinux module since upstream needs to publish the change on PyPi before it can be used by tox.

tyll:
@pcahyna @vcrhonek from my POV this is ready now. @pcahyna can you please review my yaml changes here or in #108.

tyll:
[citest bad]

tox.ini
```diff
@@ -13,7 +13,11 @@ deps =
     molecule_{lint,syntax,test}: docker
     molecule_{lint,syntax,test}: jmespath
     molecule_{lint,syntax,test}: molecule
-    molecule_{lint,syntax,test}: selinux
+    # The selinux pypi shim does not work with Ubuntu (as used by Travis), yet.
+    # Therefore use a fork with Ubuntu support. This can be changed once the
+    # update is available on PyPi.
+    # molecule_{lint,syntax,test}: selinux
+    molecule_{lint,syntax,test}: git+https://github.com/tyll/selinux-pypi-shim@fulllocation
```
pcahyna: @i386x @vcrhonek I think we can revert this part given that https://github.com/pycontribs/selinux/pull/3 is merged and released.

PR #108, opened 2019-04-25
Cleanup yaml in preparation for molecule testing

tyll:
This is a preparation for #107

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
     - python: 3.7
     - python: 3.7-dev
     - python: 3.8-dev
-    - nightly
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,3 +1,4 @@
+---
```
pcahyna: Do you need this now? It apparently causes a conflict. I think you can add it later, given that #107 contains
```yaml
  document-start: disable
```

tyll: I will resolve the conflict. It was already part of #106

tyll:
[citest bad]

tyll:
[citest bad]

Issue #102, opened 2019-02-15
error setting mtu for vlan interface despite success (eventually)


tyll:
Playbook:
```
---
- hosts: all
  vars:
    network_connections:
      - name: eth1
        type: ethernet
        autoconnect: no
        state: up
        mtu: 1492
        ip:
          dhcp4: no
          auto6: no

      - name: eth1.90
        parent: eth1
        type: vlan
        vlan_id: 90
        mtu: 1280
        state: up
        ip:
          dhcp4: no
          auto6: no

  tasks:
    - name: Run network role
      import_role:
        name: linux-system-roles.network
```

ansible-playbook output:
```
ansible-playbook -i rhel76-cloud, vlan-mtu.yml 

PLAY [all] ***********************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************************************
ok: [rhel76-cloud]

TASK [linux-system-roles.network : Check which services are running] *************************************************************************************************************************************
ok: [rhel76-cloud]

TASK [linux-system-roles.network : Check which packages are installed] ***********************************************************************************************************************************
ok: [rhel76-cloud]

TASK [linux-system-roles.network : Print network provider] ***********************************************************************************************************************************************
ok: [rhel76-cloud] => {
    "msg": "Using network provider: nm"
}

TASK [linux-system-roles.network : Install packages] *****************************************************************************************************************************************************
skipping: [rhel76-cloud]

TASK [linux-system-roles.network : Enable network service] ***********************************************************************************************************************************************
ok: [rhel76-cloud]

TASK [linux-system-roles.network : Configure networking connection profiles] *****************************************************************************************************************************
 [WARNING]: [005] <info>  #0, state:up persistent_state:present, 'eth1': add connection eth1, dc5eb498-f5b1-4124-b393-a7cfaec4e51d

 [WARNING]: [006] <info>  #0, state:up persistent_state:present, 'eth1': up connection eth1, dc5eb498-f5b1-4124-b393-a7cfaec4e51d (not-active)

 [WARNING]: [007] <info>  #1, state:up persistent_state:present, 'eth1.90': add connection eth1.90, 226409d1-c1a5-456c-8f67-e4494cf540ac

 [WARNING]: [008] <info>  #1, state:up persistent_state:present, 'eth1.90': up connection eth1.90, 226409d1-c1a5-456c-8f67-e4494cf540ac (not-active)

[...]

PR #101, opened 2019-02-07
[th/nm connection-type] fix creating vlan and mac profiles for NetworkManager

thom311:
Fixes #19

PR #110, opened 2019-05-30
tests: Test setting the MTU of VLAN interfaces

tyll:
Test setting the MTU interfaces. This needs autoconnect:false for now,
possibly because of a bug in NetworkManager.

This tests that #101 works but because of #102 it cannot enable autoconnect (which is the default, therefore it is explicitly disabled here).

---
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/110)

### DIFF
--- a/tests/ensure_non_running_provider.py
+++ b/tests/ensure_non_running_provider.py
@@ -13,7 +13,13 @@ import yaml
 
 OTHER_PROVIDER_SUFFIX = "_other_provider.yml"
 
-IGNORE = ["tests_unit.yml", "tests_helpers-and-asserts.yml", "tests_states.yml"]
+IGNORE = [
+    "tests_helpers-and-asserts.yml",
+    "tests_states.yml",
+    "tests_unit.yml",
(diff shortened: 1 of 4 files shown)


pcahyna:
The consequence of `autoconnect: false` is that the interface will not be configured after a reboot, right? Do you need it both on the parent profile and the VLAN profile? I am using `autoconnect: false` only on the parent profile and it seems to DTRT.
(It is not so much a question about the test itself but about what to recommend as a workaround to the users, if #102 is not fixed quickly.)
Also, this #102 is really triggered by setting the MTU? VLAN without MTU works fine?

pcahyna:
`autoconnect: false` is needed on the VLAN interface. Probably not on the parent, but it corresponds to the example and we can revisit it when we clarify the reason: https://github.com/linux-system-roles/network/commit/05f49d792271e632744b38f60ce0439eb0a3375b#r33769912 .
Also to be revisited is the use of true/false: https://github.com/linux-system-roles/timesync/pull/35

PR #114, opened 2019-06-06
Cleanup/prepare unit tests for ethtool option support

tyll:
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/114)


tyll:
[citest bad]

PR #115, opened 2019-06-06
Support ethtool -K|--features|--offload

### LINKED ISSUE #17: Support for ETHTOOL_OPTS such as ethtool -K|--features|--offload and others
Currently there is no way via ansible to support device specific options set via ethtool command.  This allows for setting a variety of NIC parameters from speed, duplex, variety offloads, and even buffers.  modifying these values is critical to network performance in many enterprise environments, and thus would be good if linux-system-roles supported configuring them.

tyll:
Initial support for ethtool features

TODO:
- [x] Integration test
- [x] documentation for README
- [x] Agreement on schema

This fixes #17

Deferred:
- unit tests for ArgumentValidator
- tests for provider logic

---
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/115)


tyll:
[citest bad]

library/network_connections.py
```diff
@@ -837,6 +859,41 @@ def connection_create(self, connections, idx, connection_current=None):
                     NM.SETTING_WIRED_SPEED, connection["ethernet"]["speed"]
                 )
 
+            ethernet_features = connection["ethernet"]["features"]
+
+            # Change ethtools settings only if setting is already present or
+            # if any feature is not default
```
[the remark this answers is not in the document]
tyll: thank you! I fixed this. Please take another look.

library/network_connections.py
```diff
@@ -837,6 +859,40 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                    else:
+                        if not nm_feature:
+                            raise MyError(
+                                "ethtool feature %s not support by NM" % feature
+                            )
+                        if setting:
+                            s_ethtool.set_feature(nm_feature, NM.Ternary.TRUE)
+                        else:
+                            s_ethtool.set_feature(nm_feature, NM.Ternary.FALSE)
+            elif set(ethtool_features.values()) != set([None]):
```
thom311: is this pythonic? Seems hard to read. Why not
```
    if not all([f is None for f in ethtool_features.values()]):
```
(untested)

library/network_connections.py
```diff
@@ -837,6 +859,40 @@ def connection_create(self, connections, idx, connection_current=None):
...
+            # Set them accordingly
+            # If NM does not support ethtool settings:
+            # (a) if some are specified, fail
+            # (b) if none are specified, ignore them
+            ethtool_features = connection["ethtool"]["features"]
+            try:
+                s_ethtool = NM.SettingEthtool.new()
+            except AttributeError:
+                # NM is too old to support SettingEthtool
+                s_ethtool = False
```
thom311: Could you assign here `None`? `s_ethtool` should be a NMSetting instance and not a boolean. Note that like most `GObject` constructors (*), `NM.SettingEthtool.new()` never fails -- unless libnm is too old.

Depending on how you look at it it, `None` is what pygobject would make for a non-existing GObject. Hence, one could argue that `None` is still a sensible value for a `GObject`/`NMSetting`. But `False` is an entirely different type. It seems unnecessarily confusing to mix booleans and GObject value for the same variable.

Of course, there is no difference in practice.

---


(*) Only GObjects that implement `GAsyncInitable` or `GInitable` (which few do) can fail to initialize during `g_async_initable_init_async` or `g_initable_init()`, but even then it's always a step after `g_object_new()`. `g_object_new()` can never fail, what can fail is to initiate the object afterwards. `g_initiable_new()` is basically `g_object_new() + g_initable_init()` together, and that might fail again.

library/network_connections.py
```diff
@@ -359,16 +360,37 @@ def ifcfg_create(
...
+            if setting:
+                value = "on"
+            elif setting is not None:
+                value = "off"
+
+            if value:
+                configured_features.append("%s %s" % (feature, value))
+
+        if configured_features:
+            ethtool_options += "-K %s %s" % (
```
thom311: I think this misses a space when you combine it with duplex/speed options above.

thom311: will this ETHTOOL_OPTS line actually work when passing to ethtool?


See `/etc/sysconfig/network-scripts/network-functions`
```
    IFS=';';
    if [ -n "${ETHTOOL_DELAY}" ]; then
      # Convert microseconds to seconds:
      local ETHTOOL_DELAY_SEC=$(convert2sec ${ETHTOOL_DELAY} micro)
      sleep ${ETHTOOL_DELAY_SEC}
    fi
    for opts in $ETHTOOL_OPTS ; do
        IFS=$oldifs;
        if [[ "${opts}" =~ [[:space:]]*- ]]; then
            /sbin/ethtool $opts
        else
            /sbin/ethtool -s ${REALDEVICE} $opts
        fi
        IFS=';';
    done
```


If you combine dulex/speed options, you need a ';'

library/network_connections.py
tyll: awesome advice, I think I would like it even more as:

```python
if any([value is not None for value in ethtool_features.values()])
```

Will change it.

library/network_connections.py
tyll: ok, makes sense now.

library/network_connections.py
thom311: and "any()" is faster than "not all()", because it can stop iterating the list comprehension once the result is decided...

library/network_connections.py
tyll: yes, thank you, fixed it.

library/network_connections.py
```diff
@@ -359,16 +360,39 @@ def ifcfg_create(
...
+                value = "on"
+            elif setting is not None:
+                value = "off"
+
+            if value:
+                configured_features.append("%s %s" % (feature, value))
+
+        if configured_features:
+            if ethtool_options:
+                ethtool_options += " ;"
```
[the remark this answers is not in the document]
tyll: It seems to me that it does not even need the space, just the semicolon.

library/network_connections.py
```diff
@@ -837,6 +861,40 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                        if not nm_feature:
+                            raise MyError(
+                                "ethtool feature %s not support by NM" % feature
+                            )
+                        if setting:
+                            s_ethtool.set_feature(nm_feature, NM.Ternary.TRUE)
+                        else:
+                            s_ethtool.set_feature(nm_feature, NM.Ternary.FALSE)
+            elif any([value is not None for value in ethtool_features.values()]):
+                raise MyError("ethtool.features specified but not supported by NM")
```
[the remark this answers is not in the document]
tyll: move the check to `run_prepare()`

library/network_connections.py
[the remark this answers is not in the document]
thom311: OK. That's fine with me. For the record, I find the spaces around it nicer to look at, but doesn't matter.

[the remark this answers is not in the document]
tyll: ok, added them both. :-)

tyll:
[citest pending]

PR #116, opened 2019-06-07
Revise VLAN MTU test

tyll:
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/116)

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1349,7 +1349,7 @@ class RunEnvironment:
 class RunEnvironmentAnsible(RunEnvironment):
 
     ARGS = {
-        "ignore_errors": {"required": False, "default": False, "type": "str"},
+        "ignore_errors": {"required": False, "default": False, "type": "bool"},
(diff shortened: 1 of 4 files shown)


tyll:
> Was there a wrong behaviour (false positive/false negative) with the previous version? (before commit
> [ea56ea5](https://github.com/linux-system-roles/network/commit/ea56ea5a7f02c34dad516ad4a941b45650c68aef) specifically)

Before that commit it was only testing the detected provider for each image twice instead of initscripts and NM each once.

pcahyna:
LGTM. Concerning "Set the network_provider as fact", aren't there other tests suffering from this poroblem/needing the same fix?

tyll:
> LGTM. Concerning "Set the network_provider as fact", aren't there other tests suffering from this poroblem/needing the same fix?

The other tests are using a different logic to test the current and the other provider that uses `set_fact`, therefore they do not have this problem. I plan to switch them all to use explicit providers and have a dedicated test to test provider detection eventually.

PR #118, opened 2019-06-12
Add tags to tests

i386x:
By having tags, we can select (a subset of) tests or plays depending on the given situation (i.e. depending on CI needs).


tests/restore_state.yml
```diff
@@ -11,3 +11,18 @@
 
 - name: Restore services
   import_tasks: restore_services_state.yml
+
+- name: restart NetworkManager
+  tags:
+    - 'tests::cleanup'
+  service:
+    name: NetworkManager
```
thom311: Restarting NetworkManager is not the right thing to reload profiles from disk. In fact, the only reason to ever restart NetworkManager service is when upgrading the package (without doing a reboot, which is already dirty), or when there is a serious bug and NM is dead (in which case, it is merely a workaround but not the solution).

To reload profiles either
```
    busctl call org.freedesktop.NetworkManager \
            /org/freedesktop/NetworkManager/Settings \
            org.freedesktop.NetworkManager.Settings \
            ReloadConnections 
```
or
```
    nmcli connection reload
```

tyll: The restart is a workaround to make NetworkManager forget that devices might be unmanaged (when earlier initscripts profiles with NM_CONTROLLED=no were deployed). Should your commands be enough in this case?

pcahyna: Thanks for your comment. To give you a context, the problem is, we want to run the tests against localhost, in a CI setting where we want fast results and we can't spawn VMs. The problem is now interaction between the various tests. We implemented cleanups, but we found out that there is still a problem if a test playbook for initscripts runs, creates a connection (veth or bridge) and removes it. As NetworkManager is continuously running during that, it sees the connection and remembers that it should ignore it, and it persists in its state even after the connection is removed. This leads to subsequent tests for Networkmanager failing if they create a connection of the same name - NetworkManager refuses to create it. @tyll suggested restart as a solution to this (and it works). @tyll, will the approach proposed by @thom311 work as well?

tyll:
How about also adding a tests_tier1.yml playbook that imports all the tier1 test playbooks to test in the CI whether running all tests together actually works?

tests/restore_state.yml
thom311: Not sure I understand, but `nmcli connection reload` is supposed to correctly handle changes to `NM_CONTROLLED=no`.

If you write `NM_CONTROLLED=no` to an ifcfg file and issue `ifup eth0`, then this implicitly does a `nmcli connection load /etc/sysconfig/network-scripts/ifcfg-eth0`, which results in the device becoming unmanaged. If you remove the `NM_CONTROLLED=no` line again and issue `nmcli connection reload` (or load), then the reason why the device was unmanaged is gone and it should become managed again.


> This leads to subsequent tests for Networkmanager failing if they create a connection of the same name - NetworkManager refuses to create it.

I think, NM wouldn't refuse to create the profile, it would fail to activate a profile on the device (if that device is unmanaged due to `NM_CONTROLLED=no`). But yeah... the effect is the same. It would fail.

pcahyna: We don't remove the line, we remove the whole profile.

pcahyna: (the ifcfg script)

thom311: > We don't remove the line, we remove the whole profile.

Sure. `nmcli connection reload` is supposed to cover that as well.

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +51,28 @@
...
+        - 'tests::cleanup'
+        - 'tests::tier1::cleanup'
+        - 'tests::net::bridge::cleanup'
+      import_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: statebr
+            state: down
+            persistent_state: absent
```
pcahyna: @i386x This fails, as the CI shows. You can not set anything on a nonexistent connection. (This is arguably a bug, because it means that the role is not properly idempotent, but for now the test must take this into account.)

tests/tests_bridge.yml
```diff
@@ -27,24 +40,40 @@
...
   vars:
     profile: "{{ interface }}"
 # FIXME: assert profile/device down
 
 - import_playbook: remove-profile.yml
+  tags:
+    - 'tests::tier1'
+    - 'tests::cleanup'
+    - 'tests::tier1::cleanup'
+    - 'tests::net::bridge::cleanup'
```
pcahyna: Those two plays above should not be tagged as cleanups. They are integral part of the test, given that we assert the profile absence below. If we skip them, the test will fail.

tests/restore_state.yml
tyll: > Not sure I understand, but `nmcli connection reload` is supposed to correctly handle changes to `NM_CONTROLLED=no`.
> 
> If you write `NM_CONTROLLED=no` to an ifcfg file and issue `ifup eth0`, then this implicitly does a `nmcli connection load /etc/sysconfig/network-scripts/ifcfg-eth0`, which results in the device becoming unmanaged. If you remove the `NM_CONTROLLED=no` line again and issue `nmcli connection reload` (or load), then the reason why the device was unmanaged is gone and it should become managed again.

Is there a way to unload a profile when we remove the profile without doing a generic reload? Then we could do this for persistent_state:absent with initscripts to ensure the state is also wiped from NM. Does a load for the path trigger it when the file does not exist anymore?

tyll:
@xjezda00 @richm added you for awareness since this might be a task for you now.

richm:
Do we really need this PR?  Can we close it?

tyll:
> Do we really need this PR? Can we close it?

Since there does not seem to be someone who would like to move this forward, let's close it for now and re-open if someone wants to take over.

tyll:
@pcahyna since you re-openend this PR, please clarify your intentions/next steps for it. Thank you!

pcahyna:
Next steps:
- fix the problems I pointed out above (this will make the CI pass)
- simplify by getting rid of 'tests::tier1' etc. that refer to details of our testing infrastructure and are actually uneeded
- update and merge

tests/playbooks/tests_states.yml
tyll: This is now tracked in #188

tyll:
@pcahyna it has been another month without progress. Who is going to work on this? Can we re-open or file a new PR once there is someone who will be working on it?

i386x:
I must rereview all commits I made to this branch. @pcahyna what are allowed tags now?

tyll:
Please file this again against the main branch once it is updated.

pcahyna:
We will have to do this differently anyway, the method used in this PR is too invasive and tedious.

PR #119, opened 2019-06-12
Modules/NM: Wrap changes in checkpoint

### LINKED ISSUE #33: [RFE] use NetworkManager's Checkpoint feature to rollback on error
NetworkManager's [Checkpoint](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html#gdbus-method-org-freedesktop-NetworkManager.CheckpointCreate) feature allows to record the current configuration (snapshot), together with a timeout. Commonly, the user/script then would go ahead and change the networking configuration. If the timeout expires without the checkpoint being expired, the configuration of the snapshot is restored. The user can also manually issue a rollback before the timeout expires, or just cancel the timeout (in which case the snapshot is forgotten).

This can be used for two cases:

- if the module operates and notices an issue in the middle of the operation, it can manually issue a rollback to restore the previous (working) setup.
- if the module succeeds, but the user is unable to reach the host afterwards -- that is, the new configuration cut the host off the network -- then the role won't be able to cancel the checkpoint, and the timeout will hit -- to restore the previously working configuration.

tyll:
Create a checkpoint before changing NetworkManager profiles and rollback
on failures (destroy it on success).

This fixes #33 

Deferred integration tests, since the code is covered by the other tests. only the rollback is not tested. 

---
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/119)

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -39,6 +39,8 @@ options: Documentation needs to be written. Note that the network_connections
 
 ###############################################################################
 
+DEFAULT_ACTIVATION_TIMEOUT = 90
+
 
 class CheckMode:
     PREPARE = "prepare"
@@ -1061,6 +1063,36 @@ class NMUtil:
                 % (cb_args.get("error", "unknown error"), cb_args)
             )
 
+    def create_checkpoint(self, timeout):
+        """ Create a new checkpoint """
+        checkpoint = Util.call_async_method(
+            self.nmclient,
+            "checkpoint_create",
+            [
+                [],  # devices, empty list is all devices
+                timeout,
+                Util.NM().CheckpointCreateFlags.DELETE_NEW_CONNECTIONS
+                | Util.NM().CheckpointCreateFlags.DISCONNECT_NEW_DEVICES,
+            ],
+        )
+
+        if checkpoint:
+            return checkpoint.get_path()
+        return None
+
+    def destroy_checkpoint(self, path):
+        """ Destroy the specified checkpoint """
+        Util.call_async_method(self.nmclient, "checkpoint_destroy", [path])
+
+    def rollback_checkpoint(self, path):
+        """ Rollback the specified checkpoint """
+        Util.call_async_method(
+            self.nmclient,
+            "checkpoint_rollback",
+            [path],
+            mainloop_timeout=DEFAULT_ACTIVATION_TIMEOUT,
+        )
+
     def wait_till_connection_is_gone(self, uuid, timeout=10):
         """
         Wait until a connection is gone or until the timeout elapsed
@@ -1667,6 +1699,9 @@ class Cmd:
                 self.log_fatal(idx, str(e))
         self.run_prepare()
         while self.check_mode_next() != CheckMode.DONE:
+            if self.check_mode == CheckMode.REAL_RUN:
+                self.start_transaction()
+
             for idx, connection in enumerate(self.connections):
                 try:
                     for action in connection["actions"]:
@@ -1680,11 +1715,13 @@ class Cmd:
                             self.run_action_down(idx)
                         else:
                             assert False
-                except Exception as e:
-                    self.log_warn(
-                        idx, "failure: %s [[%s]]" % (e, traceback.format_exc())
-                    )
+                except Exception as error:
+                    if self.check_mode == CheckMode.REAL_RUN:
+                        self.rollback_transaction(idx, action, error)
                     raise
+
+            if self.check_mode == CheckMode.REAL_RUN:
+                self.finish_transaction()
 
     def run_prepare(self):
         for idx, connection in enumerate(self.connections):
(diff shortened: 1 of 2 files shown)


tyll:
[citest bad]

library/network_connections.py
```diff
@@ -1789,6 +1861,14 @@ def run_prepare(self):
             connection["nm.exists"] = exists
             connection["nm.uuid"] = uuid
 
+    def finish_run_error(self, idx, action, error):
+        Cmd.finish_run_error(self, idx, action, error)
+        self.nmutil.rollback_checkpoint(self._checkpoint)
+
+    def finish_run_success(self):
```
thom311: I would call these `transaction_commit()` and `transaction_rollback()`. Of course, it's not a real "transaction" as we know it from data bases (ACID), but I think the term is better. At least, we also need a "start()" method, and the name of the start method should be reminiscent of the accompanying finish-run methods.

library/network_connections.py
```diff
@@ -1061,6 +1061,56 @@ def volatilize_connection(self, connection, timeout=10):
...
+        mainloop_timeout = 10
+
+        action_method = getattr(self.nmclient, action_name)
+        fullargs = []
+        fullargs += args
+        fullargs += (cancellable, callback, user_data)
+
+        action_method(*fullargs)
+
+        if not Util.GMainLoop_run(mainloop_timeout):
```
thom311: is having these local variables really preferable? If you don't like `if not Util.GMainLoop_ru(10):`, then `if not Util.GMainLoop_run(timeout = 10):` seems still better.

The same could be said for other local variables. Too much verbosity does not make code more legible, on the contrary it just clutters.

library/network_connections.py
```diff
@@ -1061,6 +1061,56 @@ def volatilize_connection(self, connection, timeout=10):
...
+        self._call_checkpoint("rollback", path)
+
+    def _call_checkpoint(self, action, *args):
+        action_name = "checkpoint_" + action
+        finish = action_name + "_finish"
+        callback = Util.create_callback(finish)
+        cancellable = Util.create_cancellable()
+        user_data = {}
+
+        mainloop_timeout = 10
```
thom311: timeout of 10 might not be enough for rollback.

tyll: What is your recommendation for the timeout?

thom311: it probable should be the same as it takes to activate one profile... see
```
            wait_time = connection["wait"]
            if wait_time is None:
                wait_time = 90

            try:
                self.nmutil.connection_activate_wait(ac, wait_time)
```

library/network_connections.py
```diff
@@ -1768,6 +1837,9 @@ def nmutil(self):
 
     def run_prepare(self):
         Cmd.run_prepare(self)
+
+        self._checkpoint = self.nmutil.create_checkpoint()
```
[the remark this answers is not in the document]
tyll: Thank you, changed it!

library/network_connections.py
tyll: renamed/added start/rollback/finish methods for a transaction.

library/network_connections.py
tyll: changed it to 60 for the rolback.

thom311: Why do you like that better?

library/network_connections.py
tyll: Usually I like the variables more but in this case some were not needed, changed it.

library/network_connections.py
```diff
@@ -1680,12 +1712,12 @@ def run(self):
...
                             assert False
-                except Exception as e:
-                    self.log_warn(
-                        idx, "failure: %s [[%s]]" % (e, traceback.format_exc())
-                    )
+                except Exception as error:
+                    self.rollback_transaction(idx, action, error)
                     raise
 
+        self.finish_transaction()
```
thom311: I would move the finish call inside the loop and only call it for `CheckMode.REAL_RUN`, so that it mirrors the call to start.

It makes no difference in practice, because `start_transaction()` is called zero or one times, and calling `finish_transaction()` is correct in both cases. But still, I think from the code it should be obvious that every start is exactly followed by one rollback/finish.

library/network_connections.py
```diff
@@ -1789,6 +1845,18 @@ def run_prepare(self):
...
+        Cmd.start_transaction(self)
+        self._checkpoint = self.nmutil.create_checkpoint()
+
+    def rollback_transaction(self, idx, action, error):
+        Cmd.rollback_transaction(self, idx, action, error)
+        self.nmutil.rollback_checkpoint(self._checkpoint)
+
+    def finish_transaction(self):
+        Cmd.finish_transaction(self)
+        self.nmutil.destroy_checkpoint(self._checkpoint)
```
thom311: both `destroy_checkpoint()` and `rollback_checkpoint()` do nothing if `self._checkpoint` is `None` (good).

But both calls will invalidate the checkpoint path. It would be nice if the `self._checkpoint` attribute would be reset to None, and not leave this invalid state there. Since both functions can raise an exception, to get it really right, this should be
```python
     path = self._checkpoint
     self._checkpoint = None
     self.nmutil.destroy_checkpoint(path)
```
which is kinda cumbersome...

library/network_connections.py
tyll: changed this

library/network_connections.py
tyll: I am not sure that  discarding the checkpoint before actually destroying it is right. Does an exception always indicate that the checkpoint is invalid? I changed to to at least resetting the checkpoint to None after destroying it.

library/network_connections.py
tyll: I did not notice your comment when choosing 60, I will change it to 90.

library/network_connections.py
thom311: Arguably, it doesn't really matter because on every execption from this point the program will terminated (and fail). But there really is no point in calling `destroy_checkpoint()` or `rollback_checkpoint()` more than once. After calling it, the path is certainly no longer suitable to destroy/rollback again and clearing the field seems more correct to me.

tyll: I put resetting the checkpoint into a finally block, IMHO it makes it more clear why it is there.

thom311:
lgtm

tyll:
[citest pending]

tyll:
[citest pending]

tyll:
[citest bad]

PR #120, opened 2019-06-13
Fix travis coverage

tyll:
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/120)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -268,9 +268,9 @@ max-line-length = 88
 
 [travis]
 python =
-  2.6: py26,custom
-  2.7: py27,flake8,pylint,custom
-  3.5: coveralls,custom
-  3.6: py36,black,yamllint,custom
-  3.7: py37,custom
-  3.8: py38,custom
+  2.6: py26,coveralls,custom
+  2.7: py27,coveralls,flake8,pylint,custom
+  3.5: custom
+  3.6: py36,coveralls,black,yamllint,custom
+  3.7: py37,coveralls,custom
+  3.8: py38,coveralls,custom


travis-fix-coverage.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+cat > .coveragerc <<EOF
+[paths]
+source =
+    .
+    $PWD
+EOF
+mv .coverage .coverage.merge
```
i386x: Is `.coverage` the right name? `mv` complains that there is no `.coverage`: https://travis-ci.org/linux-system-roles/network/jobs/545422641#L238

travis-fix-coverage.sh
```diff
@@ -0,0 +1,10 @@
...
+coverage combine --append .
```
i386x: Is this related to the command one line above? Can this and previous line be written as
```bash
if [[ -f .coverage ]]; then
  mv .coverage .coverage.merge
  coverage combine --append .
fi
```
?

pcahyna:
@tyll there is some error in the Travis logs that does not result in a failure of the test, but still indicates there is something wrong: https://travis-ci.org/linux-system-roles/network/jobs/545422641#L238

travis-fix-coverage.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
```
i386x: @tyll, is this script necessary as it is not invoked within `tox.ini`?

tox.ini
```diff
@@ -139,15 +139,14 @@ commands =
     flake8 --statistics {posargs} .
 
 [testenv:coveralls]
+changedir = {toxinidir}
 envdir = {toxworkdir}/env-{env:TRAVIS_PYTHON_VERSION:coveralls}
 passenv = TRAVIS TRAVIS_*
-whitelist_externals = bash
```
i386x: `whitelist_externals` should be `mv` imho.

travis-fix-coverage.sh
tyll: Please ignore this PR for now, it is not ready it I am in the middle of investigating it. A new version of python-coverage was released, that make allow a different approach.

i386x: Ok, thank You. I walked through several commits and found that in the case of d945925 the results was reported successfully (https://coveralls.io/jobs/56225730).

i386x:
If I understand it well, this happens:
1. Travis spawns virtual machine for each job specified in `.travis.yml`
1. on every virtual machine, virtual env is activated for a given version of Python and within it is run `tox`
1. `tox` runs `pytest`, for `py27` and `py36` the coverage results are reported to [coveralls.io](https://coveralls.io/)

The problem here is that recently [coveralls.io](https://coveralls.io/) is not able to combine results from multiple Travis jobs correctly (please, correct me if I am wrong). Some possibilities how to solve this:
- [parallel builds webhook](https://docs.coveralls.io/parallel-build-webhook)
- using [Travis build stages together with a file sharing service](https://docs.travis-ci.com/user/build-stages/#sharing-files-between-jobs-via-s3)
- using [codecov.io](https://codecov.io/)

What's your tips?

Some references:
1. https://github.com/mitodl/mitx-grading-library/pull/226
1. https://github.com/travis-ci/beta-features/issues/42
1. https://github.com/lemurheavy/coveralls-public/issues/1224
1. https://travis-ci.community/t/how-can-i-share-coverage-information-between-build-stages-in-prs-from-forks/859

richm:
@tyll if you still need this PR, go ahead and fix it and merge it into network, and I'll extract the common parts for template

tyll:
So, one problem that I caused IIRC was that I requested coveralls to be included only for 3.5. This was wrong since for the roles it needs to be included for each unit test. I was confusing the setup with nmstate, where coveralls runs in [`after_success`](https://github.com/nmstate/nmstate/blob/master/.travis.yml#L61). Maybe this would also be a better approach here. After fixing this, coverage is reported again but it uses the wrong paths - the coverage is prefixed with `/home/travis/build/...` which makes it annoying to browse through it. I guess this is what I intentionally wanted to fix in this PR.

tyll:
I will merge this to restore basic coverage reporting and then possibly follow-up regarding the path.

PR #121, opened 2019-06-13
Try to reapply changes

### LINKED ISSUE #38: [RFE] with NetworkManager try to use Reapply as a graceful way to change configuration
Currently, the role does the corresponding of `nmcli con up` when it notices changes to the profile. If the profile already is active, this caues NM to first tear the device down, before reactivating it. This is more disruptive than desired. The [`Reapply`](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Device.html#gdbus-method-org-freedesktop-NetworkManager-Device.Reapply) D-Bus command should be used where possible.

Note that with `Reapply`, NetworkManager first checks whether it can perform the desired actions, and abort with failure if it cannot (without touching the system). Hence, a suitable way to do this, is first try to reapply, and if that fails, fall back to full activation.

Note also, how reapply changes the ["applied-connection"](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Device.html#gdbus-method-org-freedesktop-NetworkManager-Device.GetAppliedConnection), and there is also a version-id, that allows to perform the reapply only, if the state is as expected. 

Maybe older NM versions don't yet sufficiently support Reapply. It needs to be seen whether the role needs some special handling.

tyll:
This includes one shared commit with #119 

Fixes #38

---
This change is [Reviewable](https://reviewable.io/reviews/linux-system-roles/network/121)

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1414,6 +1414,13 @@ class NMUtil:
             finally:
                 ac.handler_disconnect(ac_id)
 
+    def reapply(self, device, connection=None):
+        version_id = 0
+        flags = 0
+        return Util.call_async_method(
+            device, "reapply", [connection, version_id, flags]
+        )
+
 
 ###############################################################################
 
@@ -2144,6 +2151,9 @@ class Cmd_nm(Cmd):
         )
         self.connections_data_set_changed(idx)
         if self.check_mode == CheckMode.REAL_RUN:
+            if self._try_reapply(idx, con):
+                return
+
             try:
                 ac = self.nmutil.connection_activate(con)
             except MyError as e:
@@ -2157,6 +2167,33 @@ class Cmd_nm(Cmd):
                 self.nmutil.connection_activate_wait(ac, wait_time)
             except MyError as e:
                 self.log_error(idx, "up connection failed while waiting: %s" % (e))
+
+    def _try_reapply(self, idx, con):
+        """ Try to reapply a connection
+
+        If there is exactly one active connection with the same UUID activated
+        on exactly one device, ask the device to reapply the connection.
+
+        :returns: `True`, when the connection was reapplied, `False` otherwise
+        :rtype: bool
+        """
+        NM = Util.NM()
+
+        acons = list(self.nmutil.active_connection_list(connections=[con]))
+        if len(acons) != 1:
+            return False
+
+        active_connection = acons[0]
+        if active_connection.get_state() == NM.ActiveConnectionState.ACTIVATED:
+            devices = active_connection.get_devices()
+            if len(devices) == 1:
+                try:
+                    self.nmutil.reapply(devices[0])
+                    self.log_info(idx, "connection reapplied")
+                    return True
+                except MyError as error:
+                    self.log_info(idx, "connection reapply failed: %s" % (error))
+        return False
 
     def run_action_down(self, idx):
         connection = self.connections[idx]
--- a/tests/ensure_non_running_provider.py
+++ b/tests/ensure_non_running_provider.py
@@ -21,6 +21,7 @@ IGNORE = [
     "tests_ethtool_features_initscripts.yml",
     "tests_ethtool_features_nm.yml",
     "tests_helpers-and-asserts.yml",
+    "tests_reapply_nm.yml",
     "tests_regression_nm.yml",
     "tests_states.yml",
     "tests_unit.yml",
--- a/tests/playbooks/tests_reapply.yml
+++ b/tests/playbooks/tests_reapply.yml
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# This test should check whether the NMDevice.reapply method is called by the
+# role. This is probably a good candidate to test with pytest directly instead
+# of via Ansible. Until there is better test support for this, just check the
+# log output for the respective log message.
+---
+- hosts: all
+  vars:
+    interface: rpltstbr
+    profile: "{{ interface }}"
+    network_provider: nm
+  pre_tasks:
+    - debug:
+        msg: Inside states tests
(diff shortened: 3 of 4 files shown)


library/network_connections.py
```diff
@@ -1299,7 +1299,14 @@ def check_deactivated_cb():
...
                 ac.handler_disconnect(ac_id)
 
+    def reapply(self, interface, connection):
+        device = self.nmclient.get_device_by_iface(interface)
+        if not device:
+            return False
 
+        version_id = 0
+        flags = 0
+        return Util.call_async(device, "reapply", [connection, version_id, flags])
```
thom311: I don't think this is right.

For one, the role should not pass a `connection`. Instead, it should first make sure that the connection profile is up to date, and then pass no connection.

Also, the selection of the device doesn't seem right. The interface name does not matter because you can have profile that are not tied to a device by interface-name. What matters is whether the profile you want to activate is already active (on the right device). So, you need to find the profile, then the device(s) on which it's active, and if that the right one, then call reapply.

thom311: `Util.call_async()` is a bit of a misnomer, because the makes an async API call and waits for it to complete. It is syncronously, but the `_async` prefix sounds like it would be async. But I don't have a better suggestions, so ok.

tyll: Interesting, I used the `async` suffix because I want to call something that is async... Maybe call it `call_async_method()` instead?

pcahyna: `call_async_synchronously()` ?

thom311: I am fine with all suggestions. Just wanted to point it out. Please choose what you find best.

The "_async()" and "_sync()" suffixes you often find in glib and libnm, to indicate the nature of the function.

tyll: Changed it to `call_async_method()`.

tyll:
[citest bad]

tyll:
[citest bad]

tyll:
[citest bad]

pcahyna:
@thom311 I guess this is ready to be merged?
Also, does it require some new NM features, or will it work for all versions in RHEL 7 and RHEL 8?

thom311:
> Also, does it require some new NM features, or will it work for all versions in RHEL 7 and RHEL 8?

well,optimally there are tests that confirm (somewhat) that it works well. In practice, I think it should work just fine on rhel-7 (and rhel-8 of course too).


> I guess this is ready to be merged?

Branch looks good to me, but I didn't test it.

pcahyna:
@tyll can this be merged?

tyll:
> @tyll can this be merged?

It is missing a specific test to verify that the implemented feature is really triggered which is a challenge with the current integration test framework. From the discussion I recall the idea was to call the python code without ansible on the target system to be able to monitor the link change at the same time on the tested host by using a function like https://github.com/nmstate/nmstate/blob/ca3450dc93a4dc3af9764c832214c2c250a92535/tests/integration/testlib/iproutelib.py#L37

So far I only tested this manually with  a second shell. Maybe this could also be simulated with a async job.

pcahyna:
@tyll ah, I haven't realized the test is missing, although from looking at the changes it should have been obvious. Sorry.

You say "call the python code without ansible on the target system to be able to monitor the link change at the same time on the tested host". Do you distinguish between "target system" and "tested host"? I.e. would it be a multihost test?

pcahyna:
Also, isn't this intended to avoid disrupting services? Can't one just run a dummy service and check whether it has not been disrupted?
(I don't know right now in which way exactly does the device teardown and reactivation disrupt the service, this needs to be specified in order to detect it, as it is not described in detail in #38, just that it "is more disruptive than desired").

tyll:
> @tyll ah, I haven't realized the test is missing, although from looking at the changes it should have been obvious. Sorry.
> 
> You say "call the python code without ansible on the target system to be able to monitor the link change at the same time on the tested host". Do you distinguish between "target system" and "tested host"? I.e. would it be a multihost test?

I meant the target system with tested host. Basically use standard test roles in the original way to execute test code on the provisioned host instead of having the test completely defined in the playbook.

> Also, isn't this intended to avoid disrupting services? Can't one just run a dummy service and check whether it has not been disrupted?
> (I don't know right now in which way exactly does the device teardown and reactivation disrupt the service, this needs to be specified in order to detect it, as it is not described in detail in #38, just that it "is more disruptive than desired").

The disruption is that the interface  link goes down for a short time. I guess the disruption is mainly visible in real-time data streaming services. When I tried to disrupt a simple service implemented by socat I did not see a problem with the original code. It might also be that this has a greater effect when switches are involved and a link state change might delay connectivity because of the Spanning Tree Protocol. In Nmstate we did not find a real-world test case for the CI.

pcahyna:
> The disruption is that the interface link goes down for a short time. I guess the disruption is mainly visible in real-time data streaming services. When I tried to disrupt a simple service implemented by socat I did not see a problem with the original code. It might also be that this has a greater effect when switches are involved and a link state change might delay connectivity because of the Spanning Tree Protocol. In Nmstate we did not find a real-world test case for the CI.

What if a socket is bound to the IP address that goes away, this has no effect on the socket?
If not, isn't this event broadcast on the routing socket?

pcahyna:
if there is a link state change, wouldn't it be reported at the other end of a veth connection? Also, isn't such event reported to userspace on the routing socket (netlink)?

tyll:
To be able to move forward with the PR, I decided to add a check that will just look into the log output to see whether reapply was attempted. The exact effects of reapply can be tested as part of NetworkManager IMHO. If we have a request for a specific scenario for the role, we can still add a more specific test. Also when we have support for integration tests with pytest, we can improve/migrate this test.

PR #122, opened 2019-06-14
Tagging tests rebased

i386x:
Rebased https://github.com/linux-system-roles/network/pull/118.


tyll:
Thank you! Please push the rebased commits into #118, because the discussion is happening. I will close this PR because it is taking CI ressources away from the other PRs.

PR #127, opened 2019-07-08
Added OracleLinux as a RHEL clone for package detection purposes.

i-hill:
Oracle Linux is a RHEL clone and therefore needs to be included alongside RedHat and CentOS when deciding which packages need to be installed. OracleLinux versioning follows the same scheme as RedHat/CentOS.


i386x:
[citest commit:a022dea5829f95633d741c8e2eca6576e3eac19f]

tyll:
Thank you, LGTM!

PR #128, opened 2019-07-08
Allow more BONDING_OPTS

i-hill:
At the moment the role supports configuring the BONDING_OPTS parameters `mode` and `miimon`, however a great variety of parameters might also be supported.

This PR adds support for `use_carrier`, `updelay`, `downdelay`, `primary_reselect` and `primary`. Appreciate that's still only a small subset so I understand if you'd rather reject and wait for someone to do a fuller job, but these are the ones I needed for my project so thought I'd share.

Probably the biggest gap is that `primary` refers to the name of another network_connection. I've left that direct (i.e., you specify the `interface_name`) but I guess you might want to match the behaviour you have for `master` which takes the Ansible profile name and converts.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -51,6 +51,16 @@ ABSENT_STATE = "absent"
 
 DEFAULT_ACTIVATION_TIMEOUT = 90
 
+# Valid keys for the BONDING_OPTS parameter on a Bond interface
+VALID_BONDING_OPTS = [
+    "mode",
+    "miimon",
+    "use_carrier",
+    "updelay",
+    "downdelay",
+    "primary_reselect",
+]
+
 
 class CheckMode:
     PREPARE = "prepare"
@@ -339,9 +349,14 @@ class IfcfgUtil:
         elif connection["type"] == "bond":
             ifcfg["TYPE"] = "Bond"
             ifcfg["BONDING_MASTER"] = "yes"
-            opts = ["mode=%s" % (connection["bond"]["mode"])]
-            if connection["bond"]["miimon"] is not None:
-                opts.append(" miimon=%s" % (connection["bond"]["miimon"]))
+            opts = []
+            for opt_key in VALID_BONDING_OPTS:
+                if connection["bond"][opt_key] is not None:
+                    opts.append("%s=%s" % (opt_key, connection["bond"][opt_key]))
+            if connection["bond"]["primary"] is not None:
+                opts.append("primary=%s" % (ArgUtil.connection_find_master(
+                    connection["bond"]["primary"], connections
+                )))
             ifcfg["BONDING_OPTS"] = " ".join(opts)
         elif connection["type"] == "team":
             ifcfg["DEVICETYPE"] = "Team"
@@ -804,9 +819,9 @@ class NMUtil:
         elif connection["type"] == "bond":
             s_con.set_property(NM.SETTING_CONNECTION_TYPE, NM.SETTING_BOND_SETTING_NAME)
             s_bond = self.connection_ensure_setting(con, NM.SettingBond)
-            s_bond.add_option("mode", connection["bond"]["mode"])
-            if connection["bond"]["miimon"] is not None:
-                s_bond.add_option("miimon", str(connection["bond"]["miimon"]))
+            for opt_key in VALID_BONDING_OPTS:
+                if connection["bond"][opt_key] is not None:
+                    s_bond.add_option(opt_key, str(connection["bond"][opt_key]))
         elif connection["type"] == "team":
             s_con.set_property(NM.SETTING_CONNECTION_TYPE, NM.SETTING_TEAM_SETTING_NAME)
         elif connection["type"] == "vlan":
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -789,12 +789,29 @@ class ArgValidator_DictBond(ArgValidatorDict):
         "balance-alb",
     ]
 
+    VALID_RESELECT = ["always", "better", "failure"]
+
     def __init__(self):
         ArgValidatorDict.__init__(
             self,
             name="bond",
             nested=[
                 ArgValidatorStr("mode", enum_values=ArgValidator_DictBond.VALID_MODES),
+                ArgValidatorNum(
+                    "use_carrier", val_min=0, val_max=1, default_value=None
+                ),
+                ArgValidatorNum(
+                    "updelay", val_min=0, numeric_type=int, default_value=None
+                ),
+                ArgValidatorNum(
+                    "downdelay", val_min=0, numeric_type=int, default_value=None
+                ),
+                ArgValidatorStr("primary", default_value=None),
+                ArgValidatorStr(
+                    "primary_reselect",
+                    enum_values=ArgValidator_DictBond.VALID_RESELECT,
+                    default_value=None,
+                ),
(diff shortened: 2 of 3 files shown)


thom311:
hey. Could you squash the commit together? I think it's better to force-push the branch.

Also, could you implement this also for the NetworkManager provider? These particular options are all supported by NetworkManager alike, so it's trivial to add. See https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/libnm-core/nm-setting-bond.h?id=3d35ea645e297d80968770b7663ce0ee1607f78a#n42 for supported options (those keys are literally the same as the sysctl value that initscripts use).

i-hill:
Done!

pcahyna:
@thom311 do you think it is good to merge, or should we rather wait for @tyll ?

thom311:
The concern about how to reference to the primary interface is a good one. I didn't think of that. Yes, probably we should refer to the interface in terms of the playbook (e.g. reference another profile, and from there derive the interface-name to use).

for that reason, I would wait for @tyll's opinion.

i386x:
[citest pending]

i386x:
[citest]

yontalcar:
[citest c12f70ce2d41fd03f12d794b71c627b3f4a8ab63]

i386x:
[citest commit:c12f70ce2d41fd03f12d794b71c627b3f4a8ab63]

i386x:
[citest bad]

i386x:
[citest bad]

tyll:
> The concern about how to reference to the primary interface is a good one. I didn't think of that. Yes, probably we should refer to the interface in terms of the playbook (e.g. reference another profile, and from there derive the interface-name to use).
> 
> for that reason, I would wait for @tyll's opinion.

I agree that using a profile name instead of a device name as the primary interface makes the role interface more uniform. @i-hill Could you please change this in the PR?

tyll:
@i-hill thank you for your patience. Are you still interested in finishing this PR?
Currently, it seems that the following is missing:
-[ ] Documentation for the new settings in the README
-[ ] An example playbook with the options
-[ ] An integration test
-[ ] Supporting the profile name for primary

Please tell us your intentions. If you do not have the time to finish this, we will try to find another volunteer.

library/network_connections.py
```diff
@@ -42,6 +42,17 @@
 
 DEFAULT_ACTIVATION_TIMEOUT = 90
 
+# Valid keys for the BONDING_OPTS parameter on a Bond interface
+VALID_BONDING_OPTS = [
```
tyll, line 46: This is probably not necessary, since the ArgumentValidator takes care of ensuring that only valid options are provided - as far as I remember. Even if not, it contains the list of valid options already, so that could be re-used. What do you think?

i-hill: Indeed - this isn't about validation at this stage, it's just a list that I can iterate over in line 827-829 (of my branch) rather than building each option one by one as the existing code did.

I'll look into pulling this from ArgumentValidator.

i-hill:
> Please tell us your intentions. If you do not have the time to finish this, we will try to find another volunteer.

Makes sense. I'll certainly see what I can do.

tyll:
> > Please tell us your intentions. If you do not have the time to finish this, we will try to find another volunteer.
> 
> Makes sense. I'll certainly see what I can do.

Great. Please also take a look at #226 which adds information about the tests and other development steps.

i-hill:
> I agree that using a profile name instead of a device name as the primary interface makes the role interface more uniform. @i-hill Could you please change this in the PR?

I've been looking into this. Am I mistaken or does the circular reference here make this impossible?

I have to define the bond before the physical interface (so that the phy can resolve the name of the bond), but If I'm understanding correctly the bond therefore won't be able to resolve the phy which comes after it. 

```
network_connections:
  - name: bond_interface
    type: bond
    interface_name: bond0
    bonding:
      primary: eth_interface     <----- I don't think this can resolve

  - name: eth_interface
    type: ethernet
    interface_name: eth0
    master: bond_interface  <---- This resolves OK
    slave_type: bond
```

Do I have that wrong?

tyll:
> > I agree that using a profile name instead of a device name as the primary interface makes the role interface more uniform. @i-hill Could you please change this in the PR?
> 
> I've been looking into this. Am I mistaken or does the circular reference here make this impossible?
> 
> I have to define the bond before the physical interface (so that the phy can resolve the name of the bond), but If I'm understanding correctly the bond therefore won't be able to resolve the phy which comes after it.
> 
> ```
> network_connections:
>   - name: bond_interface
>     type: bond
>     interface_name: bond0
>     bonding:
>       primary: eth_interface     <----- I don't think this can resolve
> 
>   - name: eth_interface
>     type: ethernet
>     interface_name: eth0
>     master: bond_interface  <---- This resolves OK
>     slave_type: bond
> ```
> 
> Do I have that wrong?

Thank you for the research. Yes, this seems to be problematic. Should we maybe rename `primary` to `primary_interface` to show that it is an interface and not a profile? @thom311 what do you think?

@richm @nhosoi what is your opinion?

thom311:
for NetworkManager, the bond's `primary` option is really just the interface name, as it will be set in sysfs.

for the ansible role, I think there should be more logic. Such names should be validated, like `master` properties (that refer to other devices/profiles) are validated.

I don't think there is a technical problem that the validation considers the interface name regardless of the order. I would even think that such things are already done otherwise, are they not?

tyll:
> for NetworkManager, the bond's `primary` option is really just the interface name, as it will be set in sysfs.
> 
> for the ansible role, I think there should be more logic. Such names should be validated, like `master` properties (that refer to other devices/profiles) are validated.
> 
> I don't think there is a technical problem that the validation considers the interface name regardless of the order. I would even think that such things are already done otherwise, are they not?

Using the interface name would be less of a problem except that it might also be ok to add a profile for an interface that is not yet present, for example if it is going to be renamed after a reboot or if it is an interface in a docking station (might be useful when using bonding between wifi and docking station for a more stable VPN).

But the original idea was to use the profile instead of the interface name, which will lead to the circular dependency.

i-hill:
I've been looking further into this. The reason why `master` references have to come after the interface they refer to is that `ArgValidator_ListConnections._validate_post` calls `ArgUtil.connection_find_by_name` in a way that only considers interfaces before the one currently being validated (the third argument `idx`/`n_connections`):

```
1450: c = ArgUtil.connection_find_by_name(
1451:         connection["master"], result, idx
1452: )
```
The thing is, I'm struggling to work out exactly why. 

On the face of it, I think if I resolve `primary` without specifying `n_connections` then it _should_ succeed, but I'm keen to understand why `master` doesn't work this way in case the reason has some bearing on what I'm doing here.

I've done a scratch branch removing this limitation and added a few more unit tests which all _seem_ to pass (https://github.com/i-hill/network/commit/23547b8df7a392805112649f298a1cacbc78b400) but I think there must be an edge case I'm missing.

Do you happen to know?

i-hill:
In the meantime I've updated this PR to try this technique for resolving `primary` (init scripts only, for now) and I've added a unit test, just to get a flavour for things.

Like I say though, I'd feel more comfortable with this approach if I knew why the same wasn't done for `master`.

thom311:
because the validation of the profile is strict and forces the user to define master profiles first. There seems little reason to allow it in any order, especially because the other order seems odd.

The "circular depdendency" is no problem. Of course, you can validate the presence of the referenced profile/device in any order.

PR #129, opened 2019-07-17
Refactor Travis CI setup

i386x:
This commit introduces the refactored Travis CI setup and `tox.ini` to fullfil the need to have common CI setup across all roles.

The current proposal and discussion can be found at https://github.com/linux-system-roles/template/issues/4.

For now, this PR is in the experimental phase. After we catch all the flaws, it can be merged.


pcahyna:
Hi, does it take into account the comments in https://github.com/linux-system-roles/template/issues/4#issuecomment-512167587 ?

tox.ini
```diff
@@ -84,27 +78,40 @@ setenv =
...
+           --cov-report=term \
+           {posargs} \
+           {[base]pytesttarget}
 
+[testenv:black]
+basepython = python3.6
+deps =
+    black
+commands =
+    black --check --diff --include "^[^.].*\.py$" --exclude "/(\.[^.].*|tests/roles)/" .
```
pcahyna: can't we recurse in to subdirectories? this will avoid the nned of listing the subdirectories explicitly in `LSR_PYLINT_DIRS`

i386x: In fact, we can write a custom `init-hook` in `pylintrc` which can do all the necessary work behind directory probing for us (including handling the special cases) and then abandon `LSR_PYLINT_DIRS` completely.

tox.ini
```diff
@@ -173,9 +232,9 @@ max-line-length = 88
```
i386x: @tyll, is `[pylint]` section really necessary as we are using `pylintrc`? And what about `[pycodestyle]`?

tox.ini
pcahyna: sounds good

pcahyna:
could we use tox to test (via Molecule) with multiple versions of Ansible?

greg-hellings:
> could we use tox to test (via Molecule) with multiple versions of Ansible? Cc: @robled @greg-hellings @seandst .

I don't see why not. There's a lot of things you can do with newer versions of tox to simplify configuration (e.g. py36 will automatically set basepython to python3.6, etc) and generate a matrix of dependencies and Python versions with tox. Setting the command to include molecule directly in the commands is completely a reasonable thing to do.

seandst:
Yes, tox can be used to test with multiple versions of Ansible. We haven't yet implement this in any OASIS roles, but IIRC it tended to "just work" with deps like `molecule 'ansible>=2.7,<2.8'`, for non-latest ansible. In some cases a pip install constraint file was needed to limit the ansible version, but I didn't run into that just now in double-checking what I remember about this.

This is something I plan to revisit soon, so I might have a more in-depth answer in the coming weeks if nobody beats me to it.

tox.ini
i386x: I made a wrapper around `pylint`. Details in commit f21d01a.

tox.ini
i386x: I merge the content of `[pylint]` section with `pylintrc` and remove `[pylint]` and `[pycodestyle]` from `tox.ini` (I suppose `[pycodestyle]` is redundant as `pycodestyle` linter is not used).

molecule_requirements.txt
```diff
@@ -0,0 +1,2 @@
+jmespath
+git+https://github.com/tyll/selinux-pypi-shim@fulllocation
```
pcahyna: do we still need this, when https://github.com/pycontribs/selinux/pull/3 is already merged and released?

pcahyna: Original commit: https://github.com/linux-system-roles/network/pull/107/commits/f9fff1776d64b37432215eee8cc30a587df808ce#diff-b91f3d5bd63fcd17221b267e851608e8

i386x: I suppose no. Just `selinux` should work.

tox.ini
```diff
@@ -138,44 +143,93 @@ changedir = {[base]changedir}
...
+commands =
+    molecule {posargs} syntax
 
 [testenv:molecule_test]
-commands = molecule {posargs} test
+envdir = {toxworkdir}/env-{env:TRAVIS_PYTHON_VERSION:molecule}
+deps =
+    {[molecule]deps}
+commands =
+    molecule {posargs} test
```
seandst: This and other molecule commands only allow for testing the default scenario by default. You can manually specify other scenarios by lightly abusing posargs (`tox -e molecule_test -- -s scenarioname`), but that won't help much in CI. 

molecule testing is heavily scoped in to the scenario, and having multiple scenarios per role is common. I recommend handling support for multiple scenarios now, before potentially using this is a template for other roles tested with molecule scenarios.

pcahyna: @seandst any idea how to do that?

tox.ini
```diff
@@ -1,23 +1,20 @@
 [tox]
-envlist = black, flake8, pylint, py{26,27,36,37}, ensure_non_running_provider
+envlist =
```
tyll: AFAIU this changes the default execution behavior. I would prefer to do the (fast) static analysis before running the possibly slower unit tests.

tox.ini
```diff
@@ -138,44 +143,93 @@ changedir = {[base]changedir}
...
 [pytest]
 addopts = -rxs
 
 [flake8]
-show_source = True
+show_source = true
 max-line-length = 88
 ignore = E402,W503
 
-[pylint]
```
tyll: After removing these options, are they still picked up somehow by IDEs? AFAIK tox.ini is a standard file that is used by many tools to allow to configure the code checkers used with vim-python for example.

tox.ini
i386x: Ok. So the preferred order should be
```ini
envlist = black, flake8, pylint, py{26,27,36,37,38}, molecule_{lint,syntax,test}, extra
```
?

tox.ini
i386x: It depends on IDE imho. I don't know much about vim-python, but if it prefers `tox.ini` over another config files (like `pylintrc`), I will return these options back.

tox.ini
i386x: Sadly, `tox` is not so expressive when one needs to implement something a bit more complex than substitution in it. My idea how to handle the support for multiple scenarios is make a python wrapper around Molecule.

greg-hellings: Why can't you just set the scenario name in environment variables straight from .travis.yml?

i386x: Having something like `molecule -s {env:LSR_MSCENARIO:default} {posargs} test` in `tox` is of course possible, but if we want to run a test matrix where for every scenario from the role specific scenarios run a set of molecule commands, I am afraid that environment variables will be not enough (but maybe there exists a workaround in Molecule that I am not yet familiar with).

pcahyna:
Hi @t-woerner , IIRC you had some problems with pylint when you have a module that uses two API versions (RHEL 7 and RHEL 8), can you please point to an example? @i386x is there an easy way to disable pylint if needed?

pcahyna:
> Yes, tox can be used to test with multiple versions of Ansible. We haven't yet implement this in any OASIS roles, but IIRC it tended to "just work" with deps like `molecule 'ansible>=2.7,<2.8'`, for non-latest ansible. In some cases a pip install constraint file was needed to limit the ansible version, but I didn't run into that just now in double-checking what I remember about this.
> 
> This is something I plan to revisit soon, so I might have a more in-depth answer in the coming weeks if nobody beats me to it.

Hi @seandst , do you have any updates on this topic, please? (If not, I think we will merge this PR now and do the testing for multiple Ansible versions separately.)

pcahyna:
[citest bad]

pcahyna:
@i386x can it be merged with #120? (Another PR that touches the Travis setup.)

pcahyna:
Hi @seandst we would like to merge this PR soon, have you please looked at the way multiple Ansible versions could be supported? If not, this will continue in a separate PR.

pcahyna:
@sradco @nkinder you may be interested in this, as this is a basis for common CI setup, to be generic enough to be used eventually for all the roles. If you see here something that is contrary to the way you would do it, or something you would need is not supported, please let us know.

i386x:
[citest bad]

seandst:
> Hi @seandst we would like to merge this PR soon, have you please looked at the way multiple Ansible versions could be supported? If not, this will continue in a separate PR.

My priorities at the moment are such that I can't really spend a lot of time on this, so unfortunately all I can really do is say that testing multiple versions of Ansible is a good idea, don't have specific to provide on exactly how to do that. The same is true of being able to easily test multiple scenarios.

pcahyna:
> > Hi @seandst we would like to merge this PR soon, have you please looked at the way multiple Ansible versions could be supported? If not, this will continue in a separate PR.
> 
> My priorities at the moment are such that I can't really spend a lot of time on this, so unfortunately all I can really do is say that testing multiple versions of Ansible is a good idea, don't have specific to provide on exactly how to do that. The same is true of being able to easily test multiple scenarios.

I think the support for multiple scenarios is already there.

i386x:
Yes, support for multiple scenarios was introduced by b9b3561.

.travis.yml
```diff
@@ -1,27 +1,34 @@
...
 language: python
 matrix:
   include:
     - python: 2.6
       dist: trusty
+      # If the role contains python code that should be linted and tested, say
+      # this by setting LSR_USE_PYTHON environment variable to the value that
+      # evaluates as true in the role's repository settings page at
+      # https://travis-ci.org/
+      if: env(LSR_USE_PYTHON) IS NOT blank
```
tyll: Is it possible to configure `LSR_USE_PYTHON` in the `.travis/config.sh` file? I would prefer to have as much configuration as possible in one place. Splitting the config between the repo and the web interface makes it harder to check/understand.

.travis/config.sh
```diff
@@ -0,0 +1,13 @@
+export LSR_MOLECULE_DEPS='-rmolecule_requirements.txt'
+
+case "x${TRAVIS_PYTHON_VERSION}" in
```
tyll: The `x` prefix is not needed afaik.

.travis/config.sh
```diff
@@ -0,0 +1,13 @@
...
+  x3.5)
```
tyll: Since this code seems to run in the CI environment, can we just check here that `$TRAVIS_PYTHON_VERSION` is the system python version? Then 3.5 does not need to be hardcoded here.

.travis/config.sh
```diff
@@ -0,0 +1,13 @@
+export LSR_MOLECULE_DEPS='-rmolecule_requirements.txt'
```
tyll: It would be nice to add some copyright information to all code files to simplify license audits, for example a `# SPDX-License-Identifier: ...` header with the chosen license.

.travis/preinstall
```diff
@@ -0,0 +1,15 @@
...
+set -ex
+
+SCRIPTDIR=$(dirname $0)
+CONFIG=${SCRIPTDIR}/config.sh
+
+if [[ -f ${CONFIG} ]]; then
+  . ${CONFIG}
+fi
+
+if [[ "x${LSR_EXTRA_PACKAGES}" != "x" ]]; then
```
tyll: The `x` is not needed.

.travis/runtox
```diff
@@ -0,0 +1,19 @@
...
+
+SCRIPTDIR=$(dirname $0)
+CONFIG=${SCRIPTDIR}/config.sh
+
+if [[ -f ${CONFIG} ]]; then
+  . ${CONFIG}
+fi
+
+tox "$@"
+if [[ "x${LSR_MSCENARIOS}" != "x" ]]; then
```
tyll: AFAICS, the if condition is not needed since the for loop will not do anything if `$LSR_MSCENARIOS `is not defined.

run_pylint.py
```diff
@@ -0,0 +1,162 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
```
tyll: Please add an SPDX license identifier, this one seems to be correct here:
```# SPDX-License-Identifier: GPL-3.0-or-later```

File: tox.ini (hunk 5 above)
Comment (tyll) [reply tox.ini]: Yes, the order looks ok. I am a little bit hesitant of adding molecule to the default envlist since it requires docker/root privileges to run them IIRC. Therefore it makes more sense to run it explicitly.

File: .travis.yml (hunk 7 above)
Comment (i386x) [reply .travis.yml]: That's a good idea. It makes a sense as `./travis/config.sh` contains project specific CI settings.

File: run_pylint.py (hunk 13 above)
Comment (pcahyna) [reply run_pylint.py]: Unless it is based on another GPL-licensed source, I would prefer to use the MIT license (sorry for not having spotted it earlier)

File: .travis.yml (hunk 7 above)
Comment (i386x) [reply .travis.yml]: Oh, I forgot what was the main reason behind having `LSR_USE_PYTHON` set at the Travis side: saving the Travis resources. I am not sure if when `LSR_USE_PYTHON` will be set in `.travis/config.sh` then it will become accessible via `env()` from `.travis.yml`. Maybe there is a trick how to include additional configuration files into `.travis.yml` but I didn't find it yet.

File: .travis/config.sh (hunk 9 above)
Comment (i386x) [reply .travis/config.sh]: Sorry, I was wrong previously and misunderstood the previous comment. `TRAVIS_PYTHON_VERSION` is generally different from system python version. Instead of hardcoded 3.5 it will be better to use something like:
```bash
if [[ "${TRAVIS_PYTHON_VERSION}" == $(/usr/bin/python3 -c 'import sys; print("{}.{}".format(sys.version_info.major, sys.version_info.minor))') ]]; then
  LSR_EXTRA_PACKAGES='python3-selinux'
fi
```

tox.ini
i386x: `[pycodestyle]` and `[pylint]` sections were returned back to `tox.ini` as some IDEs may rely on them.

tox.ini
i386x: Good point, molecule is now outside of default `envlist` and can be run explicitly by `tox -e molecule`.

.travis.yml
i386x: After some considerations, having a configuration in a one place took priority over saving Travis resources when role does not contain parts written in Python[*], so `LSR_USE_PYTHON` was omitted completely.

[*] Sadly, it seems that there is no other way how to save Travis resources other than the trick with `if: env(VAR) IS NOT blank`. Note that other alternatives of `if: ...` can result in changing the common CI configuration frequently, like in case if we use repo name tests hardcoded in `.travis.yml`, which is not the path we want to follow.

run_pylint.py
i386x: I prefer MIT as well, GPL 3.0 was used only because it is the same license that is  used in [linux-system-roles/template](https://github.com/linux-system-roles/template). The license was changed to MIT now.

.travis/config.sh
pcahyna: I suggest to make the `/usr/bin/python3 -c 'import sys; print("{}.{}".format(sys.version_info.major, sys.version_info.minor))'` call a shell function to avoid such a long one-liner in the if condition.

i386x:
[citest bad]

pcahyna:
> [citest bad]

Travis will not obey that.

.travis/config.sh
i386x: After some consideration it was decided to install `python3-selinux` on all distros and silently ignore "package missing" errors. This handles the situation when system python bumps to newer version, i.e. 3.6, which cause that `python3-selinux` will not be installed in a 3.5 job, which is a job in which we run molecule.

But there is an another problem: in case of system python bump, even if `python3-selinux` will be installed in 3.5 job, selinux C binding will not be found, which results in infamous `Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!` error message.

The (imho) cleanest solution to this problem is [Allow to filter tox environment for system python](https://github.com/tox-dev/tox-travis/issues/128) proposed by @tyll.

pcahyna: I believe that we will see a system python bump only when Travis switches to a new major release of Ubuntu, which will not happen that often.

PR #132, opened 2019-11-04
Policy routing

michalmedvecky:
I needed to manage policy routing rules, so I did something that fits my needs. 

Only IP4 rules are implemented, as I don't have playground for ipv6.

Can you have a look if this fits module maintaner's expectations?


module_utils/network_lsr/argument_validator.py
```diff
@@ -373,6 +374,16 @@ def __init__(self, name, family=None, required=False, default_value=None):
                 ArgValidatorNum(
                     "metric", default_value=-1, val_min=-1, val_max=0xFFFFFFFF
                 ),
+                ArgValidatorNum(
+                    "table", default_value=None, val_min=1, val_max=253
+                ),
+                ArgValidatorStr(
+                    "dev", default_value=None, enum_values=listdir('/sys/class/net/')
```
tyll: Thank you very much for your PR. This is just some initial feedback that I saw from glancing over it. It seems to me that these `enum_values` are not correct. They would not allow adding rules for a virtual interface that is created in the same ansible run. This applies also to the other instances.

michalmedvecky: Do you think I should remove this or try fetching values of those possible new interfaces and pass them to `enum_values` as well?

michalmedvecky: @tyll can you please be more specific? I am really willing to finish this code, I already also patched ArgValidator to support number ranges for newly added ip rule parameters like sport, dport and uidrange (which are not supported on rhel/centos 7, by the way).

I am also wondering, why some tests below are pending?

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+    def __init__(self, name, family=None, required=False, default_value=None):
+        ArgValidatorDict.__init__(
+            self,
+            name,
+            required,
+            nested=[
+                ArgValidatorIP(
+                    "from", family=family, required=False, plain_address=False
+                ),
+                ArgValidatorNum("from_prefix", default_value=32, val_min=0, val_max=32),
```
thom311: I think `from_prefix` and `from` should be merged as `ArgValidatorIPAddr` types. Likewise for `to`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+                    "iif", default_value=None, enum_values=listdir('/sys/class/net/')
+                ),
+                ArgValidatorStr(
+                    "oif", default_value=None, enum_values=listdir('/sys/class/net/')
+                ),
+                ArgValidatorNum(
+                    "tos", default_value=None, val_min=0, val_max=255
+                ),
+                ArgValidatorNum(
+                    "dsfield", default_value=None, val_min=0, val_max=255
```
thom311: in iproute2, `dsfield` is an alias for `tos`. It has no meaning beyond that.

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+            required,
+            nested=[
+                ArgValidatorIP(
+                    "from", family=family, required=False, plain_address=False
+                ),
+                ArgValidatorNum("from_prefix", default_value=32, val_min=0, val_max=32),
+                ArgValidatorIP(
+                    "to", family=family, required=False, plain_address=False
+                ),
+                ArgValidatorNum("to_prefix", default_value=32, val_min=0, val_max=32),
```
thom311: `val_max` is wrong for IPv6 rules.

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+                    "ipproto", default_value=None, val_min=0, val_max=255
+                ),
+                # ArgValidatorRange(
+                #     "sport", default_value=None, val_min=0, val_max=65535
+                # ),
+                # ArgValidatorDport(
+                #     "dport", default_value=None, val_min=0, val_max=65535
+                # ),
+                ArgValidatorNum(
+                    "priority", default_value=None, val_min=0, val_max=32767
```
thom311: priority ranges up to `2^32-1`.

Also, a priority of zero means that kernel will automatically assign a priority. I think that the rule should not allow for that, because the priority is one of the most important parameters and specifying multiple rules with the same priority seems error prone. I think the user should specify this parameter. Or otherwise, the role should always generate it (e.g. by taking the index of the rule in the playbook).

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+                ),
+                ArgValidatorStr(
+                    "table", default_value="main"
+                ),
+                ArgValidatorNum(
+                    "protocol", default_value=None, val_min=1, val_max=255
+                ),
+                ArgValidatorNum(
+                    "suppress_prefixlength", default_value=None, val_min=0, val_max=32
```
thom311: the maximum depends on the address family. This has to be checked in `_validate_post()`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+                #     "sport", default_value=None, val_min=0, val_max=65535
+                # ),
+                # ArgValidatorDport(
+                #     "dport", default_value=None, val_min=0, val_max=65535
+                # ),
+                ArgValidatorNum(
+                    "priority", default_value=None, val_min=0, val_max=32767
+                ),
+                ArgValidatorStr(
+                    "table", default_value="main"
```
thom311: I think the table should first an foremost be a uint32. That is certainly what `ArgValidatorIPRule` should expose after parsing. Possibly the property can accept aliases like "main" or "local", but then this should be validated and parsed to an integer by `ArgValidatorIPRule`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,6 +414,73 @@ def _validate_post(self, value, name, result):
...
+                ),
+                ArgValidatorNum(
+                    "protocol", default_value=None, val_min=1, val_max=255
```
thom311: NetworkManager supports 20 rule attributes, "protocol" is not one of them.

That makes me wonder whether this is actually useful. Do you have an actual use-case for this? I think we should add features carefully as they are used. In particular, because the patch does not add other attributes which I think might be more useful. Let's start small, and add features as they are used.

library/network_connections.py
```diff
@@ -465,7 +465,12 @@ def ifcfg_create(
                     line += " via " + r["gateway"]
                 if r["metric"] != -1:
                     line += " metric " + str(r["metric"])
-
+                if r["dev"]:
+                    line += " dev " + str(r["dev"])
```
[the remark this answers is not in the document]
michalmedvecky: Oh no, I did not know that. My tests prove that having ".... dev eth1" in rule-eth1 still does not break anything, but I agree it makes no sense to use it in this role. I will remove it.

module_utils/network_lsr/argument_validator.py
```diff
@@ -373,6 +373,16 @@ def __init__(self, name, family=None, required=False, default_value=None):
...
                     "metric", default_value=-1, val_min=-1, val_max=0xFFFFFFFF
                 ),
+                ArgValidatorNum(
+                    "table", default_value=None, val_min=1, val_max=253
+                ),
+                ArgValidatorStr(
+                    "dev", default_value=None
+                ),
+                ArgValidatorStr(
+                    "scope", enum_values=["host", "link", "global"], default_value=None
```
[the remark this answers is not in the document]
michalmedvecky: In routes, I just implemented what I needed. I can have a look at man ip-route and implement other values as well.

The problem I see here is that test cases should be implemented for newly added features/attributes and I am not sure the company which pays me is willing to spend money on that; but I can slowly do so in my spare time.

module_utils/network_lsr/argument_validator.py
michalmedvecky: I don't disagree; but in routes, network and prefix are split as well, so I did not want to introduce any confusions.

module_utils/network_lsr/argument_validator.py
michalmedvecky: rules6 are not implemented, but it probably should be :/

library/network_connections.py
```diff
@@ -428,6 +428,11 @@ def ifcfg_create(
                 ifcfg["BOOTPROTO"] = "static"
             else:
                 ifcfg["BOOTPROTO"] = "none"
+
+            if ip["zeroconf_routes"] == False:
+                ifcfg["NOZEROCONF"] = "yes"
+            elif ip["zeroconf_routes"] == True:
+                ifcfg["NOZEROCONF"] = ""
```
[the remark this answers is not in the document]
michalmedvecky: As `NOZEROCONF` was not originally supported, I did not want to affect the existing users of the role by adding this parameter if not specified. Making `zeroconf_routes` `true` by default would mean, that every user of this role gets `NOZEROCONF=""` in his `ifcfg-<iface>` and cause changes, what is in my opinion undesired.

module_utils/network_lsr/argument_validator.py
michalmedvecky: No, I don't.

In ip rules, I implemented everything what was in man ip-rule. The other approach I could have had is to implement only attributes I need, which would not be totally fair to possible other users of the role.

But I agree with removing not really used attributes and add just some basic.

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
thom311: Actually, I was not asking to implement more features. Instead, only implement features as they are actually used.

I am surprised that "scope" is one of the attributes you care about, and I would like to know what your exact use case for it is.

library/network_connections.py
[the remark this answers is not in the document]
thom311: I meant that `zeroconf_routes` set to `true` means to not set the variable. Hence, there wouldn't be a change in behavior (with `true` being the default). Basically, just drop the second part with
```
            elif ip["zeroconf_routes"] == True:
                ifcfg["NOZEROCONF"] = ""
```

module_utils/network_lsr/argument_validator.py
michalmedvecky: Btw, in our use case we need to use `uidrange` and `sport` parameters, which are not supported below iproute2 4.15 and I don't exactly now which kernel version, but definitely not in 7.x's 3.10, and present in 8.x's 4.18. How should I treat this?

thom311: They can still be added.

Possibly `run_prepare()` could detect such missing features and either fail (fatal) or warn about it. In practice it's however not easy to understand whether the feature is available or not...

library/network_connections.py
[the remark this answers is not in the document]
michalmedvecky: ok, I understand and agree

library/network_connections.py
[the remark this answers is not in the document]
michalmedvecky: @thom311 you are not completely right. In `ifcfg-routes` of my Centos 7.6 installation there are two functions: `handle_file()`, which contains those lines you mention, and `handle_ip_file` which does NOT add `dev` as in `handle_file()`. So removing this parameter from my routing configuration effectively disabled the config, because the second line from the `ifcfg-eth0` looks like this:

```
# cat route-eth0
0.0.0.0/0 via 11.224.79.49 table 1
11.224.79.48/28 scope link table 1
```

and executing 

```
# ip ro add 11.224.79.48/28 scope link table 1
```

results in:
```
RTNETLINK answers: No such device
```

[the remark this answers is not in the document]
thom311: you are right. Still "dev" should not be something that the user configures in the playbook, but the role should figure it out itself. Like it does for 
```
            ethtool_options += "-K %s %s" % (
                connection["interface_name"],
                " ".join(configured_features),
            )
```

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
michalmedvecky: I am adding zeroconf routes in AWS, which need to have `scope link`.

module_utils/network_lsr/argument_validator.py
michalmedvecky: There is a bug in kernel (and all it's patches, 3.10-xxx) in RH/Centos 7.x, which allows adding multiple rules with the same priority. That has very bad impact on this role :/ They promised the fix should make it into 7.8 next year.

So talking about "importance" here ... I will fix the range.

module_utils/network_lsr/argument_validator.py
michalmedvecky: The problem here is, that you can specify `table` with a name. I did not yet implement that, because I had no idea whether the module or some task in the role should take care of this. Good place to answer :)

thom311: iproute2 allows you to specify the table by name. That will look into `/etc/iproute2/rt_tables`. I think that should explicitly *not* be supported. I would only support specifying the table by number.

If you care, we can allow certain aliases (`main` ==  254, `local` == 255, `default` == 253), but that should *not* look at `/etc/iproute2/rt_tables`, because then the behaviour of the role depends on some files on that target system.

module_utils/network_lsr/argument_validator.py
thom311: Right (https://bugzilla.redhat.com/show_bug.cgi?id=1700691).

But omitting the priority doesn't solve that, does it?

If you issue `ip rule delete ...` without specifying a priority, then kernel will delete the first rule that matches the remaining parameters. I don't think that is desirable and omitting the priority only makes it worse.

michalmedvecky: No, omitting priority is not an option. In such case, the role should have a deterministic system of numbering rules for the user, which would be probably hard to manage.

So we should probably just update README warning users about this potential issue.

module_utils/network_lsr/argument_validator.py
michalmedvecky: TBH I don't care in my boring use case of having 3 different route tables, but why do you think the role should not do it? We can add a task before the module is called and ensure all named tables exist in `/etc/iproute2/rt_tables`. Why do you think the module can't rely on a prerequisite, that is satisfied by the role which calls the module? I would understand such approach if that module would have ambitions to become part of Ansible core.

thom311: * it's extra work
* because then the behaviour of the playbook depends on files in /etc. I think that is generally undesirable.

michalmedvecky: I don't think we are on the same wave, but let's keep it numbered for now and possibly add names later.

library/network_connections.py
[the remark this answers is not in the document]
michalmedvecky: Hm, that's a hard job, because we need to decide which routing rule requires `dev` and which not. For example, dev is invalid for blackhole routes:

```
# ip ro add blackhole 0.0.0.0/0 table 3 dev eth0
RTNETLINK answers: Invalid argument
```

I am not sure we should implement the logic of attribute combo validation into the role :/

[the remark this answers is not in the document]
thom311: blackhole routes are currently not supported by the role.
And if support is added, then it's simple for the role to know whether to add `dev` or not.

The entire point is that the role fully understands what it is configuring there. It's not merely passing some command line parameters through to ip-route.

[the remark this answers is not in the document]
michalmedvecky: It is implemented, on my laptop :)

The point is, I know it's not needed by blackhole roles but I have no god damn idea which routes expect dev and which not. But okay, let me dive into it...

module_utils/network_lsr/argument_validator.py
michalmedvecky: @tyll @thom311 any suggestion how to cleanly pass the list or set of valid interfaces to `enum_values` in `ArgValidatorIPRoute`? I'm not that good in Python to figure out a clean solution (which would be also accepted by maintainers...)

library/network_connections.py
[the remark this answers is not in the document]
thom311: > which routes expect dev and which not.

You support router of a certain type. For these types you know it. The ones where you don't know it don't matter, because they are not supported.

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
pcahyna: @thom311 should there be a RFE for NM then about adding `scope`? Otherwise the role will miss this feature with NM (the preferred provider).

pcahyna:
> The patch only adds support of these feature for initscripts provider. That is fine in principle. The only problem is that way we don't think about whether that feature makes sense also for NetworkManager. That is relevant only insofar as to check how NetworkManager calls/uses the same thing, and whether the chosen naming and API is the most suitable one.
> 
> For example, NetworkManager calls rules `ipv4.routing-rules`. That doesn't mean "routing-rules" is a better name than "rule", but at least it raises the question. (networkd calls them `RoutingPolicyRule`, `man systemd.network`)

@thom311 I would say the name was already introduced with `rule_append_only` (if there is `route_append_only` with `route`, there should be also `rule` for `rule_append_only` for symmetry).

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
thom311: maybe. We should add features when somebody asks them with a real use-case. If anybody needs them, please open an RFE with details.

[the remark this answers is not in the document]
thom311: Adding support for "scope" for IPv4 routes at https://gitlab.freedesktop.org/NetworkManager/NetworkManager/merge_requests/349

PR #141, opened 2020-01-08
Add support for xmit_hash_policy for bond interface

unrulyv:
See: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-using_channel_bonding

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -346,6 +346,10 @@ class IfcfgUtil:
             opts = ["mode=%s" % (connection["bond"]["mode"])]
             if connection["bond"]["miimon"] is not None:
                 opts.append(" miimon=%s" % (connection["bond"]["miimon"]))
+            if connection["bond"]["xmit_hash_policy"] is not None:
+                opts.append(
+                    " xmit_hash_policy=%s" % (connection["bond"]["xmit_hash_policy"])
+                )
             ifcfg["BONDING_OPTS"] = " ".join(opts)
         elif connection["type"] == "team":
             ifcfg["DEVICETYPE"] = "Team"
@@ -815,6 +819,10 @@ class NMUtil:
             s_bond.add_option("mode", connection["bond"]["mode"])
             if connection["bond"]["miimon"] is not None:
                 s_bond.add_option("miimon", str(connection["bond"]["miimon"]))
+            if connection["bond"]["xmit_hash_policy"] is not None:
+                s_bond.add_option(
+                    "xmit_hash_policy", str(connection["bond"]["xmit_hash_policy"])
+                )
         elif connection["type"] == "team":
             s_con.set_property(NM.SETTING_CONNECTION_TYPE, NM.SETTING_TEAM_SETTING_NAME)
         elif connection["type"] == "vlan":
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -619,6 +619,15 @@ class ArgValidator_DictBond(ArgValidatorDict):
         "balance-alb",
     ]
 
+    VALID_XMIT_HASH_POLICIES = [
+        "0",
+        "layer2",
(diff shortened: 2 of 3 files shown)


thom311:
current kernel supports 5 modes:

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/if_bonding.h?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n92
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/net/bonding/bond_options.c?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n98

Also, the patch only exposes the values as numbers, while kernel accepts both numbers and names. Compare to
```
                 ArgValidatorStr("mode", enum_values=ArgValidator_DictBond.VALID_MODES),
```
Here the role opts to only expose the names only. I think the role should be consistent here. And if it opts to only expose either the names or the numbers, the names seem more important. Also, because the numbers aren't actually documented in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/bonding.txt?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n843

If anybody cares about the numbers, a new ArgValidator type could be added to handle numbers and names both.

.travis.yml
```diff
@@ -4,8 +4,6 @@ dist: xenial
 language: python
 matrix:
   include:
-    - python: 2.6
```
tyll: Why did you remove this? It is an unrelated change for the new bond policy.

unrulyv: Hi @tyll , sorry that was me just mucking around as the py2.6 test was failing with PyYAML and deprecated version. I've fixed this with my latest change to this pull request.

unrulyv:
> current kernel supports 5 modes:
> 
> [1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/if_bonding.h?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n92
> [2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/net/bonding/bond_options.c?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n98
> 
> Also, the patch only exposes the values as numbers, while kernel accepts both numbers and names. Compare to
> 
> ```
>                  ArgValidatorStr("mode", enum_values=ArgValidator_DictBond.VALID_MODES),
> ```
> 
> Here the role opts to only expose the names only. I think the role should be consistent here. And if it opts to only expose either the names or the numbers, the names seem more important. Also, because the numbers aren't actually documented in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/bonding.txt?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n843
> 
> If anybody cares about the numbers, a new ArgValidator type could be added to handle numbers and names both.

Great suggestion, I've amended my change to use just a big list of values including the names, as suggested by the RHEL doco.

unrulyv:
@thom311 what do you think about the last update I made to this commit? Any suggestion for further changes?

pcahyna:
@unrulyv thanks for the update! At first glance it seems to me that docs (README) needs updating, as noted by @tyll .

thom311:
> @thom311 what do you think about the last update I made to this commit? Any suggestion for further changes?

Hi. It looks pretty good. Thank you.

Only, I would also add the two additional modes that (recent) kernel supports. See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/net/bonding/bond_options.c?id=ae6088216ce4b99b3a4aaaccd2eb2dd40d473d42#n98

PR #133, opened 2019-11-20
Synchronize files from linux-system-roles/template

systemroller:
This PR propagates files from [linux-system-roles/template](https://github.com/linux-system-roles/template) which should be in sync across [linux-system-roles](https://github.com/linux-system-roles) repos. In case of changing affected files via pushing to this PR, please do not forget also to push the changes to [linux-system-roles/template](https://github.com/linux-system-roles/template) repo.

CC: @i386x, @pcahyna.

PR #142, opened 2020-01-09
Onboard refactored tox & Travis CI setup and configuration vol. 1

i386x:
This is a first batch of changes introduced by PR #133 (changes are introduced in batches to make reviews easier). List of changes introduced in this PR:
- document used/introduced environment variables in comments
- use absolute pathes to `config.sh` and `utils.sh`
- include `config.sh` unconditionally (`config.sh` is now a part of template, so it is created by autosynchronization script if it does not exist)
- if `python` is a system python, install `python3-selinux` (`python3-selinux` is installed by default; in case there is a reason to not install `python3-selinux`, it can be removed from `LSR_EXTRA_PACKAGES` by `config.sh`)
- install `LSR_EXTRA_PACKAGES` in one command (increase performance)

### DIFF
--- a/.travis/preinstall
+++ b/.travis/preinstall
@@ -1,18 +1,30 @@
 #!/bin/bash
 # SPDX-License-Identifier: MIT
 
-set -ex
+# Install package specified by user in LSR_EXTRA_PACKAGES. Executed by Travis
+# during before_install phase.
+#
+# LSR_EXTRA_PACKAGES, set by user in .travis/config.sh, is a space separated
+# list of packages to be installed on the Travis build environment (Ubuntu).
 
-SCRIPTDIR=$(dirname $0)
-CONFIG=${SCRIPTDIR}/config.sh
+set -e
 
-if [[ -f ${CONFIG} ]]; then
-  . ${CONFIG}
+SCRIPTDIR=$(readlink -f $(dirname $0))
+
+. ${SCRIPTDIR}/utils.sh
+
+# Add python3-selinux package (needed by Molecule on selinux enabled systems,
+# because Molecule is using `copy` and `file` Ansible modules to setup the
+# container).
+if lsr_venv_python_matches_system_python; then
+  LSR_EXTRA_PACKAGES='python3-selinux'
 fi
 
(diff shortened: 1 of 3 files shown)


.travis/preinstall
```diff
@@ -1,18 +1,30 @@
...
-if [[ -f ${CONFIG} ]]; then
-  . ${CONFIG}
+SCRIPTDIR=$(readlink -f $(dirname $0))
+
+. ${SCRIPTDIR}/utils.sh
+
+# Add python3-selinux package (needed by Molecule on selinux enabled systems,
+# because Molecule is using `copy` and `file` Ansible modules to setup the
+# container).
+if lsr_compare_pythons python -eq /usr/bin/python3; then
```
tyll: IMHO this comparison is hard to understand without knowing the previous context/digging deep into the code. How about adding a function `venv_python_matches_system_python` to give this a more descriptive name?

i386x: Good point.

i386x:
I accidentally squash commits here instead of #143.

tyll:
> I accidentally squashed commits here instead of #143. Should I revert it back?

No, I do not need the fixup PRs.

i386x:
This PR reflects the changes in https://github.com/linux-system-roles/template/commit/110a58791463d851f4b7cfecde7846e4cc4cd6ec and https://github.com/linux-system-roles/template/commit/c56d638941c81b3a5f6be80ec4abf5c224bc884d#diff-0dc30956d7844f3c69f23e3fd299cee0R137-R149.

pcahyna:
@i386x why two commits ? Shouldn't be each PR for one commit? Otherwise how will you split synchronization when you synchronize template to another role?

i386x:
@pcahyna I updated my previous comment to make it more clear: this PR was really based on one commit (https://github.com/linux-system-roles/template/commit/110a58791463d851f4b7cfecde7846e4cc4cd6ec). However, those changes requested by reviewer related to `.travis/utils.sh` was squashed with https://github.com/linux-system-roles/template/commit/c56d638941c81b3a5f6be80ec4abf5c224bc884d.

PR #146, opened 2020-01-15
Onboard refactored tox & Travis CI setup and configuration vol. 4

i386x:
This is a 4th batch of changes introduced by PR #133 (changes are introduced in batches to make reviews easier). List of changes introduced in this PR:
- rename `run_pylint.py` to `custom_pylint.py` (this is what the script really is - a customized `pylint`) and move it to `.travis/` directory where all auxiliary scripts should be placed
- call `custom_pylint.py` via shell wrapper so environment variables set by `config.sh` takes effect also when `tox` runs locally

### DIFF
--- a/.travis/custom_pylint.py
+++ b/.travis/custom_pylint.py
@@ -1,7 +1,7 @@
 #                                                         -*- coding: utf-8 -*-
 # SPDX-License-Identifier: MIT
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


.travis/custom_pylint.py
```diff
@@ -1,7 +1,7 @@
 #                                                         -*- coding: utf-8 -*-
 # SPDX-License-Identifier: MIT
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019 - 2020 Red Hat, Inc.
```
tyll: I think it is more common not to use space, i.e. write it as `2019-2020`

i386x: Changed.

i386x:
This PR reflects the changes in https://github.com/linux-system-roles/template/commit/c4b3791f92ffd4b73f8e0f3dadd2cf530c404c6d.

PR #149, opened 2020-01-22
Onboard refactored tox & Travis CI setup and configuration vol. 7

i386x:
This is a 7th batch of changes introduced by PR #133 (changes are introduced in batches to make reviews easier). List of changes introduced in this PR:
- `runsyspycmd.sh` script - runs a given command only if venv Python matches the system Python (this prevents failing when some command depends on binary libraries)
- put `molecule --version` and `ansible --version` to a standalone testenv (`[testenv:molecule_version]`) so a user can display molecule and ansible version by typing `tox -e molecule_version` without running `molecule lint`

### DIFF
--- a/.travis/runsyspycmd.sh
+++ b/.travis/runsyspycmd.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+
+# Execute command in environment python only if environment python has access
+# to system python libraries, especially C bindings. The script is run with
+# these arguments:
+#
+#   $1     - path to environment python
+#   $2     - path to system python
+#   $3     - command runnable in Python (should be present in $PATH)
+#   ${@:4} - arguments passed to $3
+
+set -e
+
+ME=$(basename $0)
+SCRIPTDIR=$(readlink -f $(dirname $0))
+TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
+
+. ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
+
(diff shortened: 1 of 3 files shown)


.travis/runsyspycmd.sh
```diff
@@ -0,0 +1,40 @@
...
+
+ME=$(basename $0)
+SCRIPTDIR=$(readlink -f $(dirname $0))
+TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
+
+. ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
+
+# Run user defined hook from .travis/config.sh.
+lsr_runsyspycmd_hook "$@"
```
tyll: Is it correct to run the hook before the check for syspython == envpython?

i386x: It allows to run commands also in a cases when the script normally exits (for example, user can specify alternative actions that should be taken in a case there is no venv python to system python match). On the other hand, it is disputable whether is such a scenario real in our case, so maybe for now we can put the hook away for simplicity?

i386x: After some consideration, I removed `lsr_runsyspycmd_hook` to make wrapper more cleaner. It has no use for us at now and I doubt about its usefulness in future.

i386x:
This PR reflects the changes in https://github.com/linux-system-roles/template/commit/7b2c0908e2f927aab02b4b563384458d9c4ea94f and in https://github.com/linux-system-roles/template/commit/c56d638941c81b3a5f6be80ec4abf5c224bc884d#diff-0dc30956d7844f3c69f23e3fd299cee0R138-R149.

PR #151, opened 2020-01-31
Onboard refactored tox & Travis CI setup and configuration vol. 8

i386x:
This is an 8th batch of changes introduced by PR #133 (changes are introduced in batches to make reviews easier). List of changes introduced in this PR:
- allow to run Molecule tests against more versions of Ansible
- allow to run custom commands also for Python 3.8-dev
- remove unecessary comment

### DIFF
--- a/.travis/runtox
+++ b/.travis/runtox
@@ -1,16 +1,45 @@
 #!/bin/bash
 # SPDX-License-Identifier: MIT
 
-set -ex
+# Run tox. Additionally, if LSR_MSCENARIOS is defined, run `tox -e molecule`
+# for every scenario from LSR_MSCENARIOS and for every Ansible version from
+# LSR_ANSIBLES.
+#
+# LSR_MSCENARIOS is a space separated list of molecule scenarios.
+# LSR_ANSIBLES is a space separated list of Ansible package names with versions
+# in pip format, i.e 'ansible ansible==2.6 ansible==2.7 ansible=2.8'.
+#
+# LSR_MSCENARIOS and LSR_ANSIBLES can be set in .travis/config.sh or as
+# environment variables.
 
-SCRIPTDIR=$(dirname $0)
-CONFIG=${SCRIPTDIR}/config.sh
+ME=$(basename $0)
+SCRIPTDIR=$(readlink -f $(dirname $0))
+BANNERSIZE=90
 
-if [[ -f ${CONFIG} ]]; then
-  . ${CONFIG}
+. ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
+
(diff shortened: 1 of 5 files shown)


.travis/runtox
```diff
@@ -1,16 +1,44 @@
...
 
-if [[ -f ${CONFIG} ]]; then
-  . ${CONFIG}
+. ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
+
+lsr_banner "tox" ${BANNERSIZE}
+(set -x; tox); error_code=$?
+
+if lsr_check_python_version python -le '3.4'; then
```
tyll: This now tries to run molecule on Travis for all Python versions but then skips the run because the system libraries can only be accessed for 3.5. This adds a lot of noise to the logs. Technically it should be possible to run all molecule tests except for the molecule_test version with other Python versions (IIRC). Only the molecule_test actually uses the container image. But we probably do not want to do this, running it once is enough, isn't it?

Therefore I propose to also add the check for the python version here to reduce the noise. Or is this not possible because this is before tox?

i386x: Thanks for noting this. Actually, as Travis documentation says, `python` command here refers also to virtualenv Python, so it should be no problem to implement the check as:
```bash
if ! lsr_venv_python_matches_system_python; then
  exit $error_code
fi
```

tox.ini
```diff
@@ -263,7 +264,8 @@ max-line-length = 88
 python =
   2.6: py26,custom
   2.7: py27,flake8,pylint,custom
-  3.5: coveralls,molecule,custom
+  3.5: coveralls,custom
   3.6: py36,black,custom
   3.7: py37,custom
   3.8: py38,custom
+  3.8-dev: py38,custom
```
tyll: This does not seem to be needed, py38 already runs when the 3.8-dev version is installed:
https://travis-ci.com/linux-system-roles/network/jobs/281802888#L215

i386x: Ok, I will remove it. Running custom commands only for stable Python versions should be enough.

i386x: Done in 4b05c20.

tyll: Even without the line, the custom command will still run with the python 3.8 development version because 3.8 in this config also matches the development version AFAICS.

i386x:
This PR reflects the changes in https://github.com/linux-system-roles/template/commit/0216bef9ea674cea592173e6c9feb02a5d972a73.

tox.ini
i386x: > Even without the line, the custom command will still run with the python 3.8 development version.

I do not see the custom command in the [log](https://travis-ci.com/linux-system-roles/network/jobs/282947052). However, it should be imho enough that the custom command runs in stable Python versions only.

PR #152, opened 2020-02-04
Onboard refactored tox & Travis CI setup and configuration vol. 9

i386x:
This is a 9th and last batch of changes introduced by PR #133 (changes are introduced in batches to make reviews easier). List of changes introduced in this PR:
- add a shell wrapper (based on PR #120), that allows to a user customize a way of publishing coverage reports, particularly:
  - disabling publishing coverage reports (in a case user's project is not registered at [coveralls.io](https://coveralls.io/))
  - specifying a path to tests and tests artifacts (in a case that they are not in `tests/`)
  - specifying a mode of the wrapper (strict, debug)

**Note:** There are some problems with reporting coverage reports to [coveralls.io](https://coveralls.io/) that are trying to be investigated and solved in PR #120. I propose to continue in investigation of these problems here so publishing coverage reports can be included also to common template.


.travis/runcoveralls.sh
```diff
@@ -0,0 +1,97 @@
...
+mv ${COVERAGEFILE} ${COVERAGEFILE}.merge
+${ENVPYTHON} -m coverage combine --append
+
+MAYBE_DEBUG=""
+if [[ "${LSR_PUBLISH_COVERAGE}" == "debug" ]]; then
+  MAYBE_DEBUG=debug
+fi
+
+set -x
+${ENVPYTHON} ${COVERALLSCMD} ${MAYBE_DEBUG} -v "$@"
```
i386x: `-v` argument actually generates a lot of noise in logs. I propose to remove it if everyone is ok with it.

richm:
I think this PR should be deleted in favor of https://github.com/linux-system-roles/network/pull/178 - if there is anything different in or missing from https://github.com/linux-system-roles/network/pull/178, please open a PR against `template`

i386x:
> I think this PR should be deleted in favor of #178 - if there is anything different in or missing from #178, please open a PR against `template`

Changes in this PR are already contained in template and therefore there are also in #178 (not counting changes in this PR that are obsolete). I propose to close this when #178 will be merged.

pcahyna:
Why not close it now?

i386x:
> Why not close it now?

To find easily its test logs, if someone needs them to investigate related issues in #178 and #120 (thinking about it now, this PR is already referenced from #178, so it will not hurt to move it out from open PR list).

pcahyna:
you can find test logs for a closed PR, see #122 (click on one of the red crosses).

PR #153, opened 2020-02-19
Synchronize files from linux-system-roles/template

systemroller:
This PR propagates files from [linux-system-roles/template](https://github.com/linux-system-roles/template) which should be in sync across [linux-system-roles](https://github.com/linux-system-roles) repos. In case of changing affected files via pushing to this PR, please do not forget also to push the changes to [linux-system-roles/template](https://github.com/linux-system-roles/template) repo.

Revision: [`782da17bfe474df40851ccf96f87d9a1ecb9ce0e`](https://github.com/linux-system-roles/template/tree/782da17bfe474df40851ccf96f87d9a1ecb9ce0e)

CC: @i386x, @pcahyna

### DIFF
--- a/.travis/custom_pylint.py
+++ b/.travis/custom_pylint.py
@@ -29,7 +29,8 @@ Usage: python custom_pylint.py ARGUMENTS
 Run pylint with ARGUMENTS followed by the list of python files contained in the
 working directory and its subdirectories.  As a python file is recognized a
 file that match INCPAT.  Files and directories that match EXPAT are skipped.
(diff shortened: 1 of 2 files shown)


tyll:
@i386x this fails in the CI, could you please take a look?

tyll:
> @i386x this fails in the CI, could you please take a look?

The CI problem is unrelated to this PR but it needs to be fixed in the template as well. Current discussion is at:
https://github.com/linux-system-roles/logging/pull/87#issuecomment-589843616

PR #156, opened 2020-02-25
Travis: Disable Python 3.8-dev

tyll:
Python 3.8-dev seems to be broken on Travis, disable it for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ matrix:
     - python: 3.6
     - python: 3.7
     - python: 3.8
-    - python: 3.8-dev
 
 services:
   - docker


richm:
Does it work if you add an explicit `[testenv:py38-dev]` to `tox.ini`?

tyll:
> Does it work if you add an explicit `[testenv:py38-dev]` to `tox.ini`?

This should not matter from my understanding. py38-dev is a Travis specific specifier which just means to use a development release for Python 3.8. Tox does not really care if it is a development or final release AFAIK. Therefore from tox point of view it should be just Python 3.8. I added it to #153 to check it there, since I do not want to lose the CI results here.

tyll:
> > Does it work if you add an explicit `[testenv:py38-dev]` to `tox.ini`?
> 
> This should not matter from my understanding. py38-dev is a Travis specific specifier which just means to use a development release for Python 3.8. Tox does not really care if it is a development or final release AFAIK. Therefore from tox point of view it should be just Python 3.8. I added it to #153 to check it there, since I do not want to lose the CI results here.

It did not help, Travis failed there with the pytest error.

PR #157, opened 2020-02-29
wireless support

jaedolph:
Implementing wired and wireless 802.1x support. Work in progress: needs better validation, doco and tests


jaedolph:
Started a new branch without the wireless support (wired 802.1x only). Will delete this PR and create a new one once I have finished the tests for it.

module_utils/network_lsr/argument_validator.py
```diff
@@ -521,6 +521,28 @@ def _validate_post(self, value, name, result):
...
+        "wpa-psk",
+        "wpa-eap",
+    ]
+
+    def __init__(self):
+        ArgValidatorDict.__init__(
+            self,
+            name="wireless",
+            nested=[
+                ArgValidatorStr("ssid"),
```
tyll: Please add a 32 character limit.

module_utils/network_lsr/argument_validator.py
```diff
@@ -521,6 +521,28 @@ def _validate_post(self, value, name, result):
...
+                ArgValidatorStr(
+                    "key-mgmt", enum_values=ArgValidator_DictWireless.VALID_KEY_MGMT
+                ),
+                ArgValidatorStr("password"),
```
tyll: This has a minimum and maximum length, too, IIRC.

tyll:
[citest commit:38eecd1313e4ae889262a83f8532b831d969b505]

library/network_connections.py
```diff
@@ -667,6 +691,13 @@ def device_is_master_type(self, dev):
                 return True
         return False
 
+    @staticmethod
+    def path_to_glib_bytes(path):
+        """
+        Converts a path to a GLib.Bytes object that can be accepted by NM
+        """
+        return Util.GLib().Bytes.new("file://{}\x00".format(path).encode("utf-8"))
```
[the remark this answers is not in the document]
jaedolph: Yeah agreed. In my new branch (without wireless support) I have put this function in the utils class

library/network_connections.py
```diff
@@ -396,6 +409,17 @@ def ifcfg_create(
...
             ifcfg["ETHTOOL_OPTS"] = ethtool_options
 
+        if connection["802-1x"] is not None:
+            ifcfg["IEEE_8021X_EAP_METHODS"] = str(connection["802-1x"]["eap"])
+            ifcfg["IEEE_8021X_IDENTITY"] = str(connection["802-1x"]["identity"])
+            ifcfg["IEEE_8021X_CA_CERT"] = str(connection["802-1x"]["ca-cert"])
+            ifcfg["IEEE_8021X_PRIVATE_KEY"] = str(connection["802-1x"]["private-key"])
+            ifcfg["IEEE_8021X_CLIENT_CERT"] = str(connection["802-1x"]["client-cert"])
+
+            keys_file = "IEEE_8021X_PRIVATE_KEY_PASSWORD={}".format(
```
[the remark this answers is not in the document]
jaedolph: Will fix

library/network_connections.py
```diff
@@ -396,6 +409,17 @@ def ifcfg_create(
         if ethtool_options:
             ifcfg["ETHTOOL_OPTS"] = ethtool_options
 
+        if connection["802-1x"] is not None:
```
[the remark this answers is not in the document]
jaedolph: Sounds good. I chose 802-1x as the key as that is how it is written in the nmcli command but can change it

library/network_connections.py
```diff
@@ -356,6 +356,19 @@ def ifcfg_create(
...
+            ifcfg["ESSID"] = str(connection["wireless"]["ssid"])
+            ifcfg["KEY_MGMT"] = str(connection["wireless"]["key-mgmt"]).upper()
+            ifcfg["MAC_ADDRESS_RANDOMIZATION"] = "default"
+
+            # need to set this by default.
+            # If ethtool vars are defined,this will get overwritten
+            ifcfg["ETHTOOL_OPTS"] = "-K %s" % connection["interface_name"]
+
+            if connection["wireless"]["key-mgmt"] == "wpa-psk":
+                keys_file = "WPA_PSK={}".format(connection["wireless"]["password"])
```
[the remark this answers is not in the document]
jaedolph: Can change to % formatting

library/network_connections.py
```diff
@@ -356,6 +356,19 @@ def ifcfg_create(
                 connection["parent"], connections, idx
             )
             ifcfg["VID"] = str(connection["vlan"]["id"])
+        elif connection["type"] == "wireless":
+            ifcfg["TYPE"] = "Wireless"
+            ifcfg["ESSID"] = str(connection["wireless"]["ssid"])
+            ifcfg["KEY_MGMT"] = str(connection["wireless"]["key-mgmt"]).upper()
+            ifcfg["MAC_ADDRESS_RANDOMIZATION"] = "default"
+
+            # need to set this by default.
```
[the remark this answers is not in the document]
jaedolph: I couldn’t get the link to come up without it when using init scripts. I then figured out that if you don’t use NetworkManager, you need to set up wireless config manually in wpa_supplicant.conf. This is the main reason I am going to do a new PR without wireless as I think it needs more work (to write wpa_supplicant config files) if we want to fully support using initscripts.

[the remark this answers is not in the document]
thom311: I think it would be fine if Wi-Fi is only supported by NetworkManager (at least, for the begining). I don't think many people use initscripts to setup Wi-Fi, do they? It's only important to get a clear error message when the user tries Wi-Fi with initscripts.

module_utils/network_lsr/argument_validator.py
jaedolph: Have added verification for this now.

module_utils/network_lsr/argument_validator.py
jaedolph: From what I looked up, the password should be less than 63 bytes. Have added a check for this.

jaedolph:
@tyll @thom311 I have fixed up the wireless branch now. Has support for WPA-PSK and WPA TLS-EAP. Only works with NetworkManager.

The integration test I have pretty much copied off the NetworkManager-ci WiFi tests. I could only get it working on RHEL7. RHEL8 doesn't seem to ship with the mac80211_hwsim kernel module and I couldn't find it in any of the option kernel/kmod packages. Fedora has it available in the `kernel-modules-internal` rpm, but it crashed every time i tired to load it on Fedora31.

If anyone has any ideas on how to fix this please let me know otherwise we can just run the integration test on RHEL7 only.

README.md
```diff
@@ -614,6 +630,20 @@ network_connections:
...
+```yaml
+network_connections:
+  - name: wlan0
+    type: wireless
+    wireless:
+      ssid: "My WPA2-PSK Network"
+      key-mgmt: "wpa-psk"
+      # recommend vault encrypting the wireless password
+      # see https://docs.ansible.com/ansible/latest/user_guide/vault.html
+      password: "p@55w0rD"
```
tyll: Does it make sense to do it like this if we also rename 802.1x to authentication? @pcahyna opinions? 



```suggestion
    wireless:
      ssid: "My WPA2-PSK Network"
      key-mgmt: "wpa-psk"
    authentication:
      # recommend vault encrypting the wireless password
      # see https://docs.ansible.com/ansible/latest/user_guide/vault.html
      password: "p@55w0rD"
```

Maybe `key-mgmt` would also make sense as part of `authentication`...

jaedolph: @tyll I grouped them this way as it corresponds to how the options are laid out in NetworkManager

I.e. all the 'wireless' options correspond to an `802-11-wireless-*` setting and the '802.1x' settings all correspond to an `802-1x.*` setting

I thought this would make the role more intuitive for users that are already familiar with nmcli and also makes the validation logic simpler.

jaedolph: If we do make a generic 'authentication' key, maybe it would be good to group the settings under different subkeys for PSK and EAP? Might make the validation logic a bit less complicated.

```
authentication:
  wpa-psk:
    password: "p@55w0rD"
```

or 

```
authentication:
  wpa-eap:
      identity: myhost
      eap: tls
      private-key: /etc/pki/tls/client.key
      # recommend vault encrypting the private key password
      # see https://docs.ansible.com/ansible/latest/user_guide/vault.html
      private-key-password: "p@55w0rD"
      client-cert: /etc/pki/tls/client.pem
      ca-cert: /etc/pki/tls/cacert.pem
```

jaedolph: @tyll  @pcahyna would you like me to change this?

tyll: Thank you for your patience, @pcahyna would like to investigate regarding the possible problems with 802.1x with Jinja.
If we do 
```
authentication:
    wpa-eap:
        ...
```
It would again be a problem with `802.1x", as it would look like:

```
authentication:
    802.1x:
        ...
```

Also `wpa-eap` would also not be a valid Python identifier.
I think if we want to go with `authentication`, using something like `mode: 802.1x` would be nice:

```
authentication:
    mode: 802.1x|wpa-psk
    ....
```
Hope this would work fine regarding validation since we probably do different validation depending on the `type` key, too.

tests/playbooks/tests_wireless.yml
```diff
@@ -0,0 +1,86 @@
...
+                  ssid: "mock_wifi"
+                  key-mgmt: "wpa-psk"
+                  password: "p@55w0rD"
+        - import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: "{{ interface }}"
+                persistent_state: absent
+                state: absent
```
tyll, line 46:
```suggestion
                state: down
```

The runtime state should be `down`.

tests/playbooks/tests_wireless.yml
```diff
@@ -0,0 +1,86 @@
...
+                  ca-cert: /etc/pki/tls/cacert.pem
+      always:
+        - block:
+            - import_role:
+                name: linux-system-roles.network
+              vars:
+                network_connections:
+                  - name: "{{ interface }}"
+                    persistent_state: absent
+                    state: absent
```
tyll, line 82:
```suggestion
                    state: down
```

tests/tests_wireless_nm.yml
```diff
@@ -0,0 +1,17 @@
...
+  tasks:
+    - name: Set network provider to 'nm'
+      set_fact:
+        network_provider: nm
+
+# workaround for: https://github.com/ansible/ansible/issues/27973
+# There is no way in Ansible to abort a playbook hosts with specific OS
+# releases Therefore we include the playbook with the tests only if the hosts
+# would support it.
+# mac80211_hwsim only seems to be available and working on RHEL/CentOS 7
```
tyll, line 13: Just clarify, it is also not available on Fedora?

tyll:
FYI: #197 introduces a script to generate the shim playbook to load the NM only playbook, this will need a small adjustment in this PR

tests/tests_wireless_nm.yml
jaedolph: Fedora has it available in the kernel-modules-internal rpm, but it crashed every time i tired to load it on Fedora31

jaedolph:
> FYI: #197 introduces a script to generate the shim playbook to load the NM only playbook, this will need a small adjustment in this PR

No worries. If you merge #197 I will rebase in to this one and update the tests

tests/playbooks/tests_wireless.yml
jaedolph: fixed

tests/playbooks/tests_wireless.yml
jaedolph: fixed

tests/playbooks/tests_wireless.yml
```diff
@@ -0,0 +1,92 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- hosts: all
+  vars:
+    interface: wlan0
+  tasks:
+    # mac80211_hwsim only seems to be available and working on RHEL/CentOS
```
jaedolph, line 7: Added this workaround as `ensure_provider_tests.py` complained if this was in the tests_wireless_nm.yml playbook

jaedolph:
@tyll have updated the tests to work with the `ensure_provider_tests.py` script

jaedolph:
[citest bad]

jaedolph:
[citest bad]

tests/tasks/setup-mock-wifi.yml
```diff
@@ -0,0 +1,91 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
tyll, line 1: Please avoid dashes in the filename

tests/playbooks/tests_802_1x.yml
```diff
@@ -35,15 +35,15 @@
                     - 203.0.113.2/24
                   dhcp4: "no"
                   auto6: "no"
-                802.1x:
```
tyll, line 38: If it is not too much trouble, it would be nice to split out the 802.1x -> ieee802_1x changes into its own commit/PR.

molecule/default/Dockerfile.j2
```diff
@@ -7,9 +7,9 @@ FROM {{ item.registry.url }}/{{ item.image }}
 FROM {{ item.image }}
 {% endif %}
 
-RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
```
tyll, line 10: Is this required or an accidental change? It does not seem to affect our CI.

tests/playbooks/tests_wireless.yml
tyll: We can also adjust `ensure_provider_tests.py` to add extra run_conditions to the shim playbook, like adding a comment in the dictionary for NM tests with another key `extra_run_conditions` and then add them to the shim playbook. What do you think?

tests/playbooks/tests_802_1x.yml
jaedolph: Have done a new PR here: #202

Will rebase in to this one if it is merged

molecule/default/Dockerfile.j2
jaedolph: yes not intentional! Will fix this

tests/playbooks/tests_wireless.yml
jaedolph: Yeah that might be a less convoluted way of doing this. Maybe something like `platforms` would be better? Could default to `all` but you could specify RHEL7 and CentOS7 for this test?

tests/tasks/setup-mock-wifi.yml
jaedolph: Have made a PR to remove the dashes from all test filenames here  #203

tests/playbooks/tests_802_1x.yml
jaedolph: rebased now

tests/tasks/cleanup-mock-wifi.yml
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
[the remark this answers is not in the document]
jaedolph: fixed

tests/tasks/setup-mock-wifi.yml
jaedolph: fixed

molecule/default/Dockerfile.j2
jaedolph: Fixed. I was a bit heavy handed with the `sed` commands!

jaedolph:
[citest bad]

tests/tasks/setup_mock_wifi.yml
```diff
@@ -0,0 +1,91 @@
...
+- name: Create eap_user_file config
+  copy:
+    content: |
+      * TLS
+    dest: /etc/hostapd/hostapd.eap_user
+
+- name: Load mac80211_hwsim kernel module to mock a wifi network
+  shell: modprobe mac80211_hwsim && sleep 5
+
+- name: Restart NetworkManager and wpa_supplicant
```
tyll, line 74: Why do they need to be restarted at this point?

tests/tasks/setup_mock_wifi.yml
```diff
@@ -0,0 +1,91 @@
...
+    warn: false
+    creates: /etc/yum.repos.d/epel.repo
+  when:
+    - ansible_distribution in ['RedHat', 'CentOS']
+
+- name: Install packages required to set up mock wifi network
+  package:
+    name:
+      - hostapd
+      - NetworkManager
```
tyll, line 16: Installing NM and wpa_supplicant should probably be left to the role to test its installation checks.

tests/tasks/setup_mock_wifi.yml
```diff
@@ -0,0 +1,91 @@
...
+      - wpa_supplicant
+    state: present
+
+- name: Ensure NetworkManager is running
```
tyll, line 20: This can/should be removed to check that the role is properly starting NM.

tests/tasks/setup_mock_wifi.yml
jaedolph: I was basically just copying what was in the NM integration tests but I'm pretty sure it will work without this

tests/tasks/setup_mock_wifi.yml
jaedolph: I will try removing this

tasks/main.yml
```diff
@@ -23,6 +23,17 @@
...
+
+# If network packages have changed and wireless connections are required,
+# NetworkManager must be restarted
+- name: Restart NetworkManager
+  service:
+    name: NetworkManager
+    state: restarted
+  when:
+    - wireless_connections_defined
+    - package_install.changed  # noqa 503
```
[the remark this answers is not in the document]
jaedolph: This is for Ansible lint https://docs.ansible.com/ansible-lint/rules/rules.html#excluding-rules. It complains that we should use a handler instead of restarting the service if something is changed. A handler is not appropriate in this case as NM needs to be restarted prior to creating the wireless interfaces

tests/playbooks/tests_wireless.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: "Skip wireless test if platform not RHEL/CentOS 7"
+      meta: end_play
+      when:
+        - (ansible_distribution_major_version != '7') or
+          (ansible_distribution not in ['RedHat', 'CentOS'])
```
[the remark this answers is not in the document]
jaedolph: will remove this

tests/tasks/setup_mock_wifi.yml
jaedolph: I don't know. It was in the NM integration tests but I think we can safely remove it.

tasks/main.yml
```diff
@@ -23,6 +23,17 @@
...
     - not network_packages is subset(ansible_facts.packages.keys())
+  register: package_install
+
+# If network packages have changed and wireless connections are required,
+# NetworkManager must be restarted
+- name: Restart NetworkManager
+  service:
+    name: NetworkManager
+    state: restarted
+  when:
```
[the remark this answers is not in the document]
jaedolph: Will fix

tests/playbooks/tests_wireless.yml
jaedolph: Good point. I think your original suggestion is better, should we create a new issue for this?

tasks/main.yml
```diff
@@ -23,6 +23,17 @@
     state: present
   when:
     - not network_packages is subset(ansible_facts.packages.keys())
+  register: package_install
+
+# If network packages have changed and wireless connections are required,
+# NetworkManager must be restarted
+- name: Restart NetworkManager
```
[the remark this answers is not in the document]
jaedolph: Yeah I am not too happy with this implementation, but restarting NM prior to setting up the connections was the only way I could get wireless to work if NetworkManager-wifi was not previously installed.

I would say it's likely that the main use case for using this role to set up wireless connections would be running the role locally (e.g. on a laptop). I think in this case it would be safe to restart NM.

defaults/main.yml
```diff
@@ -20,10 +20,23 @@ network_provider_current: "{{
...
-wpa_supplicant_required: "{{ network_connections |
-      json_query('[*][ieee802_1x]') | flatten | count > 0 }}"
-_network_packages_default_802_1x: ["{% if wpa_supplicant_required
+# check if any 802.1x connections are defined
+_ieee802_1x_connections_defined: "{{
+      network_connections|json_query('[*][ieee802_1x]')
+      | flatten | count > 0 }}"
+
+# check if any wireless connections are defined
+wireless_connections_defined: "{{
```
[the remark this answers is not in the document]
jaedolph: These vars are also used in tasks/main.yml so I need them to be global

defaults/main.yml
```diff
@@ -20,10 +20,23 @@ network_provider_current: "{{
...
+      ['wireless'] in network_connections|json_query('[*][type]') }}"
+
+# NetworkManager-wireless is required for wireless connections
+_network_packages_default_wireless: ["{% if wireless_connections_defined
+      %}NetworkManager-wifi{% endif %}"]
+
+# wpa_supplicant is required if any 802.1x or wireless connections are defined
+wpa_supplicant_required: "{{
```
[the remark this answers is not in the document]
jaedolph: These vars are also used in tasks/main.yml so I need them to be global

tests/tasks/setup_mock_wifi.yml
jaedolph: Tested removing both of these but it appears hostapd requires wpa_supplicant to be running

tests/ensure_provider_tests.py
```diff
@@ -61,6 +61,7 @@
     "playbooks/tests_reapply.yml": {},
     "playbooks/tests_states.yml": {},
     "playbooks/tests_802_1x.yml": {},
+    "playbooks/tests_wireless.yml": {},
```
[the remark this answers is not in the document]
jaedolph: Sorted now

README.md
pcahyna: `wpa-eap` could be called `wpa_eap` and `802.1x` could be called `ieee8021x` . Not sure whether it is better to have a key for each mode or a `mode` key with each mode as a possible value.

jaedolph:
@tyll I tried getting the wireless tests to work without installing NetworkManager and wpa_supplicant during the setup, but it appears that either the hwsim interface or hostapd has issues with this

defaults/main.yml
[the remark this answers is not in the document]
tyll: Does Ansible protect variables with underscores nowadays? @pcahyna what do you suggest here? AFAIU, internal variables should be prefixed with underscores and this is one.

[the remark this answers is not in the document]
tyll: @richm can you maybe help here?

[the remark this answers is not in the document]
richm: I don't think Ansible works like python with variables that begin with `_` or `__` - it's really just a convention/best practices that if you see an Ansible variable that begins with `__` you should assume you cannot pass in the value from outside of the role, or use the variable outside of the role.  So - if `wireless_connections_defined` is to be used only by the `network` role, then it should begin with `__` to denote it as such (and the name of the variable should begin with `__network_` according to https://github.com/oasis-roles/meta_standards#naming-things)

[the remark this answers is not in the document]
tyll: @jaedolph did you see the answer? Would be great if you could adjust the variable names. Please let me know if you lack the time and we can also take it on.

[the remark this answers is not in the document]
jaedolph: @tyll I have fixed the wireless related vars now. I have also added a commit to prefix all the other role variables with __network

defaults/main.yml
```diff
@@ -10,44 +10,61 @@ network_provider_os_default: "{{
       'OracleLinux'
     ] and ansible_distribution_major_version is version('7', '<')
     else 'nm' }}"
+
 # If NetworkManager.service is running, assume that 'nm' is currently in-use,
 # otherwise initscripts
-network_provider_current: "{{
```
tyll, line 15: Not sure, if this is considered an internal variable or if a user might want to be able to query this. @pcahyna what's your opinion?

tasks/main.yml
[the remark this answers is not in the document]
tyll: @pcahyna @richm should we fail here or restart NM?

tasks/main.yml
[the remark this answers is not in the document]
tyll: How about adding a link to https://docs.ansible.com/ansible-lint/rules/default_rules.html to make this is easier to check in case we keep the task.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1165,6 +1208,40 @@ def _validate_post(self, value, name, result):
                         % (result["type"]),
                     )
 
+            if result["type"] == "wireless":
```
tyll, line 1211: How about moving this in a separate method like `validate_wireless()?

tasks/main.yml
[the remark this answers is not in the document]
richm: Consider adding a variable `network_nm_restart_ok` - if `true`, this means it is OK for the role to restart nm as needed.  If `false`, this means the role should emit some sort of message like "These changes require NetworkManager restart to take effect.  Please restart it when it is safe to do so."  I suppose the defaults/main.yml setting should be `network_nm_restart_ok: false`

defaults/main.yml
richm: What is the case where `network_provider_current` would be different than `network_provider` or `network_provider_os_default` (https://github.com/oasis-roles/meta_standards#providers for reference)

tyll: `network_provider_current` would be considered read-only for the user. So for example users could use an empty network configuration and then query `network_provider_current` to check what is detected. It can be differetn that the os default. For example the OS default for CentOS 7 is nm but the cloud images do not contain it as far as I know.

Not sure, if the variable is really useful for useres. I am fine with making this an internal variable but I wanted to check with you, first. If you agree with making it internal, let's do it.

tasks/main.yml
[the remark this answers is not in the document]
tyll: Using such a variable sounds fine by me. It would like to drop the `nm` infix to make it provider agnostic. How about

`network_allow_restart`?

@thom311 what is the impact of restarting NM? What is your opinion? I guess the biggest problem might be breaking VPN connections but we do not support them, yet.

defaults/main.yml
richm: I don't have an opinion as to whether to make it internal or not - I think it depends on what you think would be useful to users of the network role.

tyll: It does not seem to be special for the network role but for the same for any role with multiple providers. Since there is no clear request to have this variable not internal, I am in favor of making it internal.

defaults/main.yml
```diff
@@ -69,9 +73,9 @@ if ansible_distribution in ['RedHat', 'CentOS', 'OracleLinux'] and
 # |select() filters the list to include only values that evaluate to true
 #     (the empty string is false)
 # |list() converts the generator that |select() creates to a list
-network_packages_default_initscripts: "{{ ['ethtool']
-+ _network_packages_default_initscripts_bridge|select()|list()
-+ _network_packages_default_initscripts_network_scripts|select()|list()
+__network_packages_default_initscripts: "{{ ['ethtool']
```
tyll, line 76: If it is not too much trouble, it would be great if you put the commit to rename the variables in its own PR, so that we can merge it already. PSA: #231 will conflict with this change.

jaedolph: Have created PR here #233

tasks/main.yml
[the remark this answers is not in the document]
jaedolph: I have added an network_allow_restart variable that defaults to 'yes'. I think it would be unlikely that a user would not want NetworkManager restarted if they are configuring wireless connections but we can at least give them the option to turn it off. Probably the most common use case for this would be configuring a laptop or workstation device locally where the connection dropping out is not a huge deal

Please let me know what you think

tests/tasks/setup_mock_wifi.yml
jaedolph: didn't work when I removed this

tests/tasks/setup_mock_wifi.yml
jaedolph: didn't work when I removed this

module_utils/network_lsr/argument_validator.py
jaedolph: Have separated this out into a separate function now. Might be worth doing for the other connection types too.

module_utils/network_lsr/argument_validator.py
```diff
@@ -753,6 +753,47 @@ def __init__(self):
...
+
+        if value["key_mgmt"] == "wpa-psk":
+            if "password" in value:
+                if len(value["password"]) > 63:
+                    raise ValidationError(
+                        name, "invalid wpa password, must be 63 characters or less",
+                    )
+            else:
+                raise ValidationError(
+                    name, "must supply a password if using 'wpa-psk' key management",
```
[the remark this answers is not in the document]
jaedolph: Have added a check for this

module_utils/network_lsr/argument_validator.py
```diff
@@ -753,6 +753,47 @@ def __init__(self):
...
+                ArgValidatorStr(
+                    "key_mgmt", enum_values=ArgValidator_DictWireless.VALID_KEY_MGMT
+                ),
+                ArgValidatorStr("password", default_value=None),
+            ],
+            default_value=None,
+        )
+
+    def _validate_impl(self, value, name):
+        if len(value["ssid"]) > 32:
```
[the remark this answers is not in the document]
jaedolph: Good idea. Have implemented this now.

tasks/main.yml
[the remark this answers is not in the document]
jaedolph: Have added the link now

jaedolph:
[citest bad]

jaedolph:
@tyll one of the travis tests for this is stuck, do you know how to fix this?

tyll:
> @tyll one of the travis tests for this is stuck, do you know how to fix this?

There should only be one status, so you can ignore the stuck one. Since there is a conflict, this needs to be rebased and hopefully there will be only one status afterwards.

jaedolph:
> > @tyll one of the travis tests for this is stuck, do you know how to fix this?
> 
> 
> 
> There should only be one status, so you can ignore the stuck one. Since there is a conflict, this needs to be rebased and hopefully there will be only one status afterwards.

I will rebase and push again. Please let me know if there are any other changes you would like me to make to the pr :)

tests/playbooks/tests_wireless.yml
tyll: I added support for this in https://github.com/tyll/network/tree/extra_run_condition - can you add the commit from there to this PR to be able to use it here. I would like to have an actual test that uses it before adding it to the repo.

tests/tasks/setup_mock_wifi.yml
tyll: How about we split the test into two phases:

1) Setup a wireless profile without the hostapd maybe with state present/down and check that the profile is generated correctly
2) Setup hostapd etc but without touching wpa_supplicant or restarting NM (this should now work because the first step should already have done this
3) Change the profile to state:up to do the actual integration test

What do you think? I feel a little bit uneasy that the test replicates some steps that are actually part of the role logic. This is also the issue with the other steps here that (re)start NM.

tasks/main.yml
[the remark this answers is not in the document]
tyll: > Using such a variable sounds fine by me. It would like to drop the `nm` infix to make it provider agnostic. How about
> 
> `network_allow_restart`?
> 
> @thom311 what is the impact of restarting NM? What is your opinion? I guess the biggest problem might be breaking VPN connections but we do not support them, yet.

@thom311 please take a look here.

> I have added an network_allow_restart variable that defaults to 'yes'. I think it would be unlikely that a user would not want NetworkManager restarted if they are configuring wireless connections but we can at least give them the option to turn it off. Probably the most common use case for this would be configuring a laptop or workstation device locally where the connection dropping out is not a huge deal
> 
> Please let me know what you think

The other perspective is that laptops or workstations will probably already have NetworkManager-wifi installed, so they will not benefit from this and defaulting to `false` for `network_allow_restart` should not be an issue for most users but still be the safer default.

tests/tasks/setup_mock_wifi.yml
jaedolph: @tyll I tried implementing this but it appears the role will try and "down" the profile even if it is already down (step 1). This fails for a wireless connection unless I commented out most of the "run_action_down" function. I could add "failed_when: false" to the initial connection creation but this is probably not a great solution

I can try working on this some more but maybe it would be better to just confirm all the variables are set correctly?
i.e. checking __network_wpa_supplicant_required == true, __network_wireless_connections_defined == true or something?

On a default RHEL7 installation I believe NetworkManager and wpa_supplicant will already be installed and running so I don't think it is a huge issue if they are restarted before we run the role.

tests/playbooks/tests_wireless.yml
jaedolph: Have added this commit now. Works well :)

tasks/main.yml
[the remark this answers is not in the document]
pcahyna: Could the role determine whether it is safe to restart? If the only problem expected is with VPNs, the role could detect whether a VPN is being used and use `false` as the default value for `network_allow_restart`, otherwise use `true`.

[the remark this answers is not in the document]
tyll: NetworkManager might also forget other runtime information when being restarted, so the it would be best if users restarted the system for deterministic behavior.

Also, just for creating the profiles (persistent_state:present, no state), the wifi plugin would not be needed. It is only needed to bring the profiles up. So technically, this should be checked here, as well. I am currently more inclined not to restart NM in the role and request the user to do this. This would also simplify testing. What do you think?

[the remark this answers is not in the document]
tyll: @jaedolph would it work for you if we default restarting to false and re-consider if there is use feedback that tells us to enable it? I am just wondering how we could unblock this for now since we do not have any real user stories about how/when this would be used.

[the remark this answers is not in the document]
jaedolph: I think this is a good idea, probably safest to default to not restarting. I will have a look at this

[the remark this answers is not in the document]
tyll: I changed the default to "no" so we can merge it.

examples/wireless_wpa_psk.yml
```diff
@@ -0,0 +1,13 @@
...
+  vars:
+    network_connections:
+      - name: wlan0
+        type: wireless
+        wireless:
+          ssid: "My WPA2-PSK Network"
+          key_mgmt: "wpa-psk"
+          # recommend vault encrypting the wireless password
+          # see https://docs.ansible.com/ansible/latest/user_guide/vault.html
+          password: "p@55w0rD"
```
pcahyna, line 13: this example forgets to actually use the role (via `include_role` or similar).

jaedolph: oops, my mistake

tasks/main.yml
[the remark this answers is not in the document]
jaedolph: Thanks @tyll, sorry that I was so slow at getting to this

PR #159, opened 2020-03-03
802 1x support wired

jaedolph:
Re making this PR without the wireless support. As mentioned in previous PR, wireless support in initscripts will require management of the wpa_supplicant.conf file which requires more work. May create a PR for the wireless stuff later.


library/network_connections.py
```diff
@@ -486,6 +486,31 @@ def ifcfg_create(
...
+            if connection["802-1x"]["private-key-password-flags"]:
+                flags_str = Util.convert_passwd_flags_initscripts(
+                    connection["802-1x"]["private-key-password-flags"]
+                )
+                ifcfg["IEEE_8021X_PRIVATE_KEY_PASSWORD_FLAGS"] = flags_str
+
+            ifcfg["IEEE_8021X_CLIENT_CERT"] = str(connection["802-1x"]["client-cert"])
+
+            if connection["802-1x"]["ca-cert"]:
+                ifcfg["IEEE_8021X_CA_CERT"] = str(connection["802-1x"]["ca-cert"])
```
thom311: when you do `grep -R IEEE_8021X_ /etc/sysconfig/`, you'll see that initscripts don't support any of these. See also, `/etc/sysconfig/network-scripts/ifup-wireless`.

The role also sets `NM_CONTROLLED=no`, so even if you have NetworkManager running, `ifup` won't call NetworkManager to handle this file. Hence, it's just not going to work.

At least, I doubt any of this will work. Did you test it? I would just drop initscripts support altogether and fail if running such a playbook against initscripts provider.

module_utils/network_lsr/argument_validator.py
```diff
@@ -698,6 +699,64 @@ def _validate_post(self, value, name, result):
...
+        ArgValidatorStr._validate_impl(self, value, name)
+
+        if posixpath.isabs(value) is False:
+            raise ValidationError(
+                name, "value '%s' is not a valid posix absolute path" % (value),
+            )
+        return value
+
+
+class ArgValidator_Dict802_1X(ArgValidatorDict):
```
thom311: I think a 802-1x setting only works for type `wifi` and `ethernet`. That is, because it's only implemented for those types. In NetworkManager you cannot do 802.1x port authentication on (for example) a bridge interface.

I think `_validate_post()` should reject specifying a 802.1x setting for unsupported types.

.travis/preinstall
```diff
@@ -16,8 +16,9 @@ SCRIPTDIR=$(readlink -f $(dirname $0))
 # Add python3-selinux package (needed by Molecule on selinux enabled systems,
 # because Molecule is using `copy` and `file` Ansible modules to setup the
 # container).
+# Add gi packages so tests can run
 if lsr_venv_python_matches_system_python; then
-  LSR_EXTRA_PACKAGES='python3-selinux'
+  LSR_EXTRA_PACKAGES='python3-selinux python3-gi python-gi gir1.2-gtk-3.0 python-gi-cairo python3-gi-cairo'
```
tyll: Are really all of these packages required? python3-gi/python-gi should probably be enough.

.travis.yml
```diff
@@ -22,6 +24,7 @@ services:
 
 before_install:
   - ./.travis/preinstall
+  - sudo apt get -y install python3-gi python-gi gir1.2-gtk-3.0 python-gi-cairo python3-gi-cairo
```
tyll: This seems to be a duplicate of the packages from the preinstall script.

module_utils/network_lsr/argument_validator.py
jaedolph: Will add this to the _validate_post

.travis/preinstall
jaedolph: Even with these packages i still can’t get gi to import. Will try some other stuff. Maybe I can just mock the calls to NM() too

library/network_connections.py
jaedolph: Yep, my bad. Seems the 802.1x lines in initscripts are only used by network manager. I’m pretty sure you have to configure the 802.1x auth in wpa_supplicant.conf for it to work properly with NetworkManager disabled

jaedolph: As far as testing, I only tested wireless 802.1x but obviously not very well!

I configured 802.1x with a wireless access point running OpenWRT, but do not have hardware that supports wired 802.1x.

jaedolph:
So do we think the best thing to do would be to merge this branch with my other one and delete the initscripts parts?

I can also add some validation/doco to make it clear we are only supporting 802.1x using NetworkManager for now

library/network_connections.py
thom311: Not sure I understand. Do you mean, you also tested wireless 802.1x **with initscripts** and it worked? Interesting, I wouldn't expect that...

jaedolph: I did test with initscripts and got it to work, but I believe my test method was flawed.
One of the tests I did was this:
* Configure wireless 802.1x using the role with "network_provider: nm"
* Ping the gateway
* Stop NetworkManager. Run ifdown
* Configure wireless 802.1x using the role with "network_provider: initscripts"
* Ping the gateway

I'm pretty sure what was happening was that NetworkManager set up the wpa_supplicant process correctly and this was still running in the background when I used initscripts, giving the impression that the initscripts method was working.

I had made the incorrect assumption that all parameters defined by NetworkManager in /etc/sysconfig/network-scripts were compatible with initscripts but this is clearly not the case.

thom311: > I had made the incorrect assumption that all parameters defined by NetworkManager in /etc/sysconfig/network-scripts were compatible with initscripts but this is clearly not the case.

That is the aim in general, but only the case for a subset of parameters/options (also, because NetworkManager supports many more parameters than initscripts).

tyll:
> So do we think the best thing to do would be to merge this branch with my other one and delete the initscripts parts?

Removing initscripts sounds good. Since the wireless support is not needed, I would prefer to only add wired 802.1x support in this PR and get it completely ready. Afterwards wireless support can be added in a second PR. This will reduce the complexity of this PR and allow to finish it faster. What do you think?

> I can also add some validation/doco to make it clear we are only supporting 802.1x using NetworkManager for now

:+1:

jaedolph:
Sounds good, I will have some time tomorrow to try and finish this PR. Will remove the initscripts part and try to get the tests working on the CI/CD

tyll:
> Sounds good, I will have some time tomorrow to try and finish this PR. Will remove the initscripts part and try to get the tests working on the CI/CD

Awesome, if possible, please add a integration test like https://github.com/linux-system-roles/network/blob/master/tests/tests_ethtool_features_nm.yml - it does not need to verify that the authentication actually works, since we do not have an easy way to setup a test authentication server and if there is, we can still refine the test.
It would be great if the integration test could import the example playbook from `examples/eth-with-802-1x.yml` to ensure the example is tested. For this it might need to use a variable instead of a hardcoded interface name.

examples/eth-with-802-1x.yml
```diff
@@ -0,0 +1,27 @@
...
+---
+- hosts: network-test
+  vars:
+    network_connections:
+      - name: eth0
+        type: ethernet
+        802.1x:
+          identity: myhost
+          private-key: /etc/pki/tls/client.key
+          # recommend vault encrypting the private key password
```
tyll: Please add a link to the vault docs here.

examples/eth-with-802-1x.yml
```diff
@@ -0,0 +1,27 @@
...
+          private-key-password: "p@55w0rD"
+          client-cert: /etc/pki/tls/client.pem
+          ca-cert:  /etc/pki/tls/cacert.pem
+
+  # certs have to be deployed first
+  pre_tasks:
```
tyll: It is also possible to use `import_role` to have a task to run the role, removing the need for pre_tasks. This might be a better way of doing things to allow to easier re-use the example in another playbook. What do you think?

Example:
https://github.com/linux-system-roles/network/blob/master/tests/playbooks/tests_ethtool_features.yml#L77

jaedolph: I can change this if you like. I tend to prefer to write playbooks with pre_tasks and post_tasks so all the roles are kept together but that is just a personal preference

.travis/preinstall
jaedolph: removed these now

.travis.yml
jaedolph: gone now

examples/eth-with-802-1x.yml
jaedolph: added link

tests/unit/test_network_connections.py
```diff
@@ -197,7 +199,8 @@ def do_connections_validate_nm(self, input_connections, **kwargs):
             if "type" in connection:
                 connection["nm.exists"] = False
                 connection["nm.uuid"] = n.Util.create_uuid()
-        mode = VALIDATE_ONE_MODE_INITSCRIPTS
```
jaedolph: Pretty sure this was a bug

jaedolph:
Have fixed up some issues and added more tests for the 802.1x features (including an integration test). All tests are passing now

README.md
```diff
@@ -403,6 +403,61 @@ kernel and device, changing some features might not be supported.
...
+ * `ca-cert`
+
+    Absolute path to the PEM encoded certificate authority used to verify the EAP server.
+
+  * `system-ca-certs`
+
+    If set to `True`, NetworkManager will use the system's trusted ca certificates to verify.
+    the EAP server.
+
+    This setting will have no effect if `network_provider: initscripts` is used.
```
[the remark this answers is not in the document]
jaedolph: Oops, will delete this. Also found a couple of values that are missing "(required)" in doco

module_utils/network_lsr/argument_validator.py
jaedolph: Yeah, I thought that was a better spot to put it as it is easy to determine if the role is being run with NetworkManager or initscripts

thom311:
lgtm. I think the two patches can be combined.

tests/playbooks/tests_802_1x.yml
```diff
@@ -0,0 +1,66 @@
...
+          vars:
+            network_connections:
+              - name: "{{ interface }}"
+                state: up
+                type: ethernet
+                ip:
+                  dhcp4: "no"
+                  auto6: "no"
+                # "wait: 0" is required as the connection will fail to activate.
+                # The test environment doesn't have an actual 802.1x EAP provider.
```
tyll: The NetworkManager CI has a script to setup hostapd with veth to test this scenario.

https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci/-/blob/master/prepare/hostapd_wired.sh
https://gitlab.freedesktop.org/NetworkManager/NetworkManager-ci/-/blob/master/nmcli/features/ethernet.feature#L417

Could you use this information to setup hostapd before this playbook and clean it up in the cleanup block?

jaedolph: Thanks for the review guys.

I had at getting the NetworkManager-ci test environment working without much success. My main issue was getting traffic to go between two bridged veth interfaces. I was trying this even without any 802.1x auth and the only way i could get it working was using network namespaces:

```
ip link add veth1 type veth peer name veth1-br
ip link add veth2 type veth peer name veth2-br

ip link add br1 type bridge
ip link set br1 up

ip netns add ns1
ip netns add ns2

ip link set veth1 netns ns1
ip link set veth2 netns ns2

ip netns exec ns1 ip addr add 10.1.1.11/24 dev veth1
ip netns exec ns2 ip addr add 10.1.1.12/24 dev veth2

ip link set veth1-br up
ip link set veth2-br up

ip link set veth1-br master br1
ip link set veth2-br master br1

ip netns exec ns1 ip link set veth1 up
ip netns exec ns2 ip link set veth2 up

ip netns exec ns1 ping -c3 10.1.1.12
ip netns exec ns2 ping -c3 10.1.1.11
```

Is there anything I should be doing different to get this working without namespaces?

tyll: Where did you try this? Might there be a firewall causing problems?

jaedolph: Main thing I noticed when not using namespaces is I couldn't get ARP to go over the bridge, it just hit the bridge and didn't go anywhere...

jaedolph: Sorry, didn't see your prev message. Firewalld is stopped and I did an iptables -F, doesn't seem to make a difference.
Was trying this on a rhel7 vm

jaedolph: My method that doesn't work is this:
```
ip link add veth1 type veth peer name veth1-br
ip link add veth2 type veth peer name veth2-br

ip link add br1 type bridge
ip link set br1 up


ip addr add 10.1.1.11/24 dev veth1
ip addr add 10.1.1.12/24 dev veth2

ip link set veth1-br up
ip link set veth2-br up

ip link set veth1-br master br1
ip link set veth2-br master br1

ip link set veth1 up
ip link set veth2 up
```

tcpdump on br1 looks like this when i run `ping 10.1.1.12 -I veth1`:
```
[root@rhel7-test ~]# tcpdump -enni br1
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on br1, link-type EN10MB (Ethernet), capture size 262144 bytes
08:03:54.949716 ea:8c:4f:b2:d9:9d > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Request who-has 10.1.1.12 tell 10.1.1.11, length 28
08:03:55.951722 ea:8c:4f:b2:d9:9d > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Request who-has 10.1.1.12 tell 10.1.1.11, length 28
08:03:57.948878 ea:8c:4f:b2:d9:9d > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Request who-has 10.1.1.12 tell 10.1.1.11, length 28
08:03:58.949938 ea:8c:4f:b2:d9:9d > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Request who-has 10.1.1.12 tell 10.1.1.11, length 28
08:03:59.951731 ea:8c:4f:b2:d9:9d > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 42: Request who-has 10.1.1.12 tell 10.1.1.11, length 28
```

tyll: @vbenes do you have any idea?

jaedolph: Talked to a guy at work today and he said we should be all good to just create one interface in a namespace and leave the other in the default namespace. Then we can run the hostapd process in the secondary namespace and the ansible role can configure 802.1x on the interface in the default namespace.

Managed to get this to work:
```
ip netns exec ns1 ip addr add 10.1.1.11/24 dev veth1
ip addr add 10.1.1.12/24 dev veth2

ip link set veth1-br up
ip link set veth2-br up

ip link set veth1-br master br1
ip link set veth2-br master br1


ip netns exec ns1 ip link set veth1 up
ip link set veth2 up

ping -c3 10.1.1.11 -I veth2
```

Next step will be to try to run hostapd like this:
```
ip netns exec ns1 hostapd --arg1 --arg2
````

Comment (jaedolph) [reply tests/playbooks/tests_802_1x.yml]: Managed to get hostapd to run in a namespace and could successfully connect to it with an 802.1x interface created using the role. Made a playbook to set everything up but it needs a little more work. Will aim to add another commit to the PR tomorrow. 

Does anyone know if the test vms have the EPEL repo enabled? Otherwise I will configure this as part of the playbook.

**Comment by jaedolph** [comment]
New hostpad integration seems to work OK on EL7 but I will do some more work to get it working on EL8/Fedora

I also noticed a wierd thing on RHEL7 and RHEL8 where i cannot set 802-1x-system-ca-certs using the role or nmcli. It accepts the command `nmcli con mod eth0 802-1x-system-ca-certs yes` but when i run `nmcli con show eth0` it always shows the value of 802-1x-system-ca-certs as 'no'. 

I don't have this issue on Fedora. Could this be a bug with NetworkManager that is fixed in versions later than 1.18? Maybe something to do with how NetworkManager is packaged on RHEL?

Also, I found out today that NetworkManager does not do any verification if a ca-cert is not supplied. I found an open issue in NetworkManager regarding this: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/220

File: tests/playbooks/tests_802_1x.yml (hunk 9 above)
Comment (tyll) [reply tests/playbooks/tests_802_1x.yml]: EPEL is not enabled by default.

**Comment by tyll** [comment]
@thom311 please take a look at the nmcli issues.

**Comment by thom311** [comment]
> I also noticed a wierd thing on RHEL7 and RHEL8 where i cannot set 802-1x-system-ca-certs using the role or nmcli. It accepts the command `nmcli con mod eth0 802-1x-system-ca-certs yes` but when i run `nmcli con show eth0` it always shows the value of 802-1x-system-ca-certs as 'no'.
> 
> I don't have this issue on Fedora. Could this be a bug with NetworkManager that is fixed in versions later than 1.18? Maybe something to do with how NetworkManager is packaged on RHEL?

That certainly sounds like a bug. Possibly it's fixed on latest versions, dunno. Which were the exact vesions where you encountered the issue?


> Also, I found out today that NetworkManager does not do any verification if a ca-cert is not supplied. I found an open issue in NetworkManager regarding this: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/issues/220

If you don't configure a certificate, it doesn't validate it. It's true, that if you also don't enable 802-1x.system-ca-certs then no validation is performed. I commented on issue 220 why I think system-ca-certs isn't that useful as it sounds. So yes: configure a certificate! NetworkManager (the daemon) will allow you to do that. A GUI should preferably make it very obvious that such a configuration is a bad idea.

**Comment by jaedolph** [comment]
@thom311 I did a more thorough test and it is definitely looking like a bug with NetworkManger on RHEL7 and RHEL8. Here are some results of me running the same commands on fully patched Fedora31, RHEL8.1 and RHEL7.7 boxes:

```
[root@fedora ~]# nmcli --version
nmcli tool, version 1.20.10-1.fc31
[root@fedora ~]# NetworkManager --version
1.20.10-1.fc31
[root@fedora ~]# ip link add veth1 type veth peer name veth1p
[root@fedora ~]# nmcli con add con-name 8021x-test ifname veth1 type ethernet 802-1x.eap tls 802-1x.client-cert /etc/pki/tls/client.pem 802-1x.private-key /etc/pki/tls/client.key.nocrypt 802-1x.system-ca-certs yes
Connection '8021x-test' (cd1ab693-f845-4057-a3b2-2a201e61fe55) successfully added.
[root@fedora ~]# nmcli con show 8021x-test | grep 802-1x.system-ca-certs
802-1x.system-ca-certs:                 yes

[root@rhel8 ~]# nmcli --version
nmcli tool, version 1.20.0-5.el8_1
[root@rhel8 ~]# NetworkManager --version
1.20.0-5.el8_1
[root@rhel8 ~]# ip link add veth1 type veth peer name veth1p
[root@rhel8 ~]# nmcli con add con-name 8021x-test ifname veth1 type ethernet 802-1x.eap tls 802-1x.client-cert /etc/pki/tls/client.pem 802-1x.private-key /etc/pki/tls/client.key.nocrypt 802-1x.system-ca-certs yes
Connection '8021x-test' (78378edf-e5da-4d23-8c7c-7835b9544a9e) successfully added.
[root@rhel8 ~]# nmcli con show 8021x-test | grep 802-1x.system-ca-certs
802-1x.system-ca-certs:                 no

[root@rhel7-test ~]# ip link add veth1 type veth peer name veth1p
[root@rhel7-test ~]# nmcli con add con-name 8021x-test ifname veth1 type ethernet 802-1x.eap tls 802-1x.client-cert /etc/pki/tls/client.pem 802-1x.private-key /etc/pki/tls/client.key.nocrypt 802-1x.system-ca-certs yes
Connection '8021x-test' (7e7ffbe9-93b6-4130-9a7a-b1687e602bb1) successfully added.
[root@rhel7-test ~]# nmcli con show 8021x-test | grep 802-1x.system-ca-certs
802-1x.system-ca-certs:                 no
```

**Comment by jaedolph** [comment]
The new tests seem to be working correctly on all the versions now. I think some repos were timing out during a couple of the fedora builds though, are we able to manually run these ones again?

**Comment by jaedolph** [comment]
> lgtm. I think the two patches can be combined.

@thom311  do you want me to rebase all this into one commit?

**Comment by thom311** [comment]
> @thom311 I did a more thorough test and it is definitely looking like a bug with NetworkManger on RHEL7 and RHEL8.

@jaedolph, ah right. I forgot about this issue. Thank you for bringing this up. This was fixed by [1] (for 1.22.0), also backported to [2] (for 1.20.8).

I now also backported it to nm-1-18 branch ([3]), for future 1.18.6 minor release.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/2a4fb75d3b03d8d4391b10ad028a991dc6cf78e8
[2] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/d0572b660239cdfefae90f55c130b0babf9064a3
[3] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/73de171aac5ae89a6a3332112944728ad747d456

**Comment by thom311** [comment]
> > lgtm. I think the two patches can be combined.
> 
> @thom311 do you want me to rebase all this into one commit?

Commit [2] merely fixes stuff from the first commit [1]. These two should be squashed together (so that you end up with 3 commits on this pull-request, instead of 4).

[1] "Adding support for 802.1x wired connections (EAP-TLS only) Fixed bug in do_connections_validate_nm() function."
[2]  "Fixed mistake in the 802.1x section of the README"

**Comment by jaedolph** [comment]
> @jaedolph, ah right. I forgot about this issue. Thank you for bringing this up. This was fixed by [1] (for 1.22.0), also backported to [2] (for 1.20.8).
> 
> I now also backported it to nm-1-18 branch ([3]), for future 1.18.6 minor release.

Awesome, glad I could help

> Commit [2] merely fixes stuff from the first commit [1]. These two should be squashed together (so that you end up with 3 commits on this pull-request, instead of 4).

I will squash the second commit into the first. This will also run the tests again and hopefully the Fedora ones work this time!

**Comment by tyll** [comment]
[citest bad]

File: tests/tasks/setup-802_1x_server.yml [hunk 10]
```diff
@@ -0,0 +1,79 @@
...
+    ip link add veth2 type veth peer name veth2-br
+
+    ip link add br1 type bridge
+    ip link set br1 up
+
+    ip netns add ns1
+
+    ip link set veth1 netns ns1
+
+    ip netns exec ns1 ip addr add 10.1.1.11/24 dev veth1
```
tyll: Please use IP addresses from https://tools.ietf.org/html/rfc5737 to avoid conflicts with local networks.

```suggestion
    ip netns exec ns1 ip addr add 203.0.113.0/24 dev veth1
```

tests/tasks/setup-802_1x_server.yml
```diff
@@ -0,0 +1,79 @@
...
+    ip link set veth1-br up
+    ip link set veth2-br up
+
+    ip link set veth1-br master br1
+    ip link set veth2-br master br1
+
+    ip netns exec ns1 ip link set veth1 up
+    ip link set veth2 up
+
+    # Enable forwarding of EAP 802.1x messages through software bridge "b1".
```
tyll:
```suggestion
    # Enable forwarding of EAP 802.1x messages through software bridge "br1".
```

tests/tasks/setup-802_1x_server.yml
jaedolph: Haven't heard of this RFC before! fixed now so all ips in the 802.1x tests use the 203.0.113.2/24 subnet

tests/tasks/setup-802_1x_server.yml
jaedolph: fixed

jaedolph:
[citest bad]

jaedolph:
@tyll any other issues with this PR?

tyll:
> @tyll any other issues with this PR?

I  added a fixup commit to fix some syntax issues in the test that were missed by a faulty CI config that was fixed in master. I noticed that the current commits could be squashed together, since the changes belong together. If you like, I can do this but I want to make you aware of the changes in your branch so nothing is accidentally lost.

jaedolph:
Thanks for fixing this up @tyll. I have squashed into one commit now

jaedolph:
[citest bad]

tyll:
[citest bad]

README.md
```diff
@@ -603,6 +655,23 @@ network_connections:
       rule_append_only: yes
 ```
 
+Configuring 802.1x:
+
+```yaml
+network_connections:
+  - name: eth0
+    type: ethernet
+    802.1x:
```
pcahyna: I propose to change the name to `802_1x` or some other valid identifier, so that the name can be used in e.g. Jinja2. Dot has a special meaning.

tyll: `802_1x` is also not a valid variable name in python, since the first character cannot be a number, so to access it the `[]` needs to be used anyhow, as in:

 `{{ network_connections[0]["802.1x"]["identity"] }}` or `{{ network_connections[0]["802_1x"]["identity"] }}`

If you still think it is a problem, please open an issue for this to track this and ideally provide a better name. Also if we want to have special restrictions here, it would make sense to add a test for this in the CI or the actual code to changes like this to get in (could be discussed in the ticket to be created).

vbenes:
ieee8021x?

jaedolph:
happy to change to a different name
> ieee8021x?
This looks like a good option

tyll:
jaedolph <notifications@github.com> schrieb am Fr., 27. März 2020, 12:39:

> happy to change to a different name
>
> ieee8021x?
>
> This looks like a good option
>

I wass thinking that maybe "authentication" might be a good generic
alternative, too.

>

jaedolph:
> I wass thinking that maybe "authentication" might be a good generic alternative, too.
> […](#)

Yeah this would be a good one too. Maybe something like eap_auth would be good too?

PR #160, opened 2020-03-14
Update tests_default.yml to follow new standards

a-mere-peasant:
I've updated this to follow the new standard for tests.Please let me know if anything else needs to be done.


a-mere-peasant:
could you please tell me if something else is needed in this file ? @tyll

PR #162, opened 2020-03-15
Updating tests_bridge and tests_ethernet to be consistent with newer tests

a-mere-peasant:
I've updated these tests so they are now consistent with the new ones see [tests_ethool_feaures_nm](https://github.com/linux-system-roles/network/blob/master/tests/tests_ethtool_features_nm.yml)
Please do let me know if anything else needs to be done ,
Thanks

### DIFF
--- a/tests/tests_bridge.yml
+++ b/tests/tests_bridge.yml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
-- name: Test configuring bridges
-  hosts: all
+- hosts: all
+  name: Test configuring bridges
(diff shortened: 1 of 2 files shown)


tyll:
Thank you. The idea from the GSoC tasks was to make the tests use playbooks that explicitly select the initscripts or nm provider like it is done with tests_ethool_feaures_nm and tests_ethool_feaures_initscripts. Is it more clear now?

a-mere-peasant:
Oh , I get it now ,thanks for the clarification , i'll get started on it and close this pr

Issue #165, opened 2020-03-18
Use podman for the molecule test


RMPR:
- [X] Subscribe to the system roles mailing list and send an introduction message. Tell us which area interests you and why you would like to participate
- [X] Take a look at the integration tests and try to get them running. The are run by a custom CI tool and are inspired by the Standard Test Interface
- [X] Get the unit and linting tests running using tox
- [ ] Optional: Get the molecule tests running by running tox -e molecule (this requires being able to run docker as the current user)

When I try to run `tox -e molecule` I get 
> ERROR: Unable to contact the Docker daemon. Please refer to https://docs.docker.com/config/daemon/ for managing the daemon

Which is not surprising because I don't have docker installed, but on the flip side I have podman installed and since from the [whatis podman page](https://podman.io/whatis.html) it's said that:
> Simply put: `alias docker=podman`

It's my understanding that there's a way to run this test if the user has podman instead of docker. 
Am I right? If yes can you provide some guidance on how to work around this issue? 
Thanks.

richm:
molecule v3 supposedly has support for podman - I tried several things to get podman to work with molecule but I could never get it to work.  If you can figure out how to make it work, patches are welcome.

tyll:
> When I try to run `tox -e molecule` I get
> 
> > ERROR: Unable to contact the Docker daemon. Please refer to https://docs.docker.com/config/daemon/ for managing the daemon
> 
> Which is not surprising because I don't have docker installed, but on the flip side I have podman installed and since from the [whatis podman page](https://podman.io/whatis.html) it's said that:
> 
> > Simply put: `alias docker=podman`
> 
> It's my understanding that there's a way to run this test if the user has podman instead of docker.
> Am I right? If yes can you provide some guidance on how to work around this issue?

Is the error message from when you tried this alias with molecule? What system are you using? What happens when you try to run molecule with podman?

RMPR:
Yes, I aliased docker to podman and I also created a symlink at /usr/bin/docker because some programs relied on that:
OS: Fedora 31 Workstation.
I tried to change the line requiring docker to podman #166 but still an error, the last five lines of the stacktrace:
```
--> Sanity checks: 'docker'
ERROR: Missing Docker driver dependency. Please install via 'molecule[docker]' or refer to your INSTALL.rst driver documentation file
[...]

PR #166, opened 2020-03-19
Replace docker by podman in the installdeps

RMPR:


tyll:
Why is this a good idea? What problem does it solve?

richm:
> Why is this a good idea? What problem does it solve?

It solves the "problem" that we do not want to use docker except where strictly necessary - we want to use podman wherever possible.

tyll:
> > Why is this a good idea? What problem does it solve?
> 
> It solves the "problem" that we do not want to use docker except where strictly necessary - we want to use podman wherever possible.

So changing the requirement makes molecule use podman instead of docker? It does not seem to me that it works this way. @RMPR Nevertheless, the justification needs to be added to the commit message. Please follow https://chris.beams.io/posts/git-commit

RMPR:
I don't actually want this to be merged, I wanted to make the test fallback on podman if it's installed and docker no. See #165

tyll:
> I don't actually want this to be merged, it is just an illustration of what I tried. See #165

Thank you for the clarification. There is no need to file a PR to show this. You can just point to your fork/commit in the issue. Also, if you do not want something to be merged, create it as a draft PR and mention this in the description.

RMPR:
Okay, thanks for the pointers.

Issue #97, opened 2019-01-10
RFE: Improve teamd support


tyll:
It seems to be supported in the code but it is not documented.

tyll:
@aneagoe thank you! I will keep you update when we make progress here. I will be at [DevConf.cz](https://devconf.info/cz), [CentOS Dojo Brussels](https://wiki.centos.org/Events/Dojo/Brussels2019) and [FOSDEM](https://fosdem.org/2019/), soon. In case you are there as well and would like to chat about the role, please let me know.

tyll:
I created a branch to support setting the runner now with NM: https://github.com/tyll/network/tree/team_options

The option name is not stable, yet. We are considering to adjust the schema to match the bonding options.

tyll:
@aneagoe all the items from the TODO list in th PR are still undone plus tests are missing:

- [ ] Use a default to tx_hash in the role (NM seems to default to "tx_hash\": [\"eth\", \"ipv4\", \"ipv6\"]}}
- [ ] Add support for initscripts (this seems to require writing JSON in
the TEAM_CONFIG variable with shell quoting), example:
TEAM_CONFIG="{\"runner\": {\"name\": \"lacp\", \"tx_hash\": [\"eth\", \"ipv4\", \"ipv6\"]}}"
- [ ] Add more options
- [ ] Add documentation/example
- [ ] Add integration tests

Also we need a decision whether or not to abstract teamd/bonding settings in a more generic way. @pcahyna suggested it, I am currently not convinced that it is necessary. Also, at this point, this is not a priority, therefore someone who would like to do the coding would be good, too.

aneagoe:
@tyll Understood. With regards to MODE vs RUNNER I'd keep things separate; there is a certain divergence between bond and team (ie outside of "broadcast" there's no overlap). It could be confusing to use a generic definition.
As well, how much should be validated with regards to TEAM_CONFIG? JSON validation can be easy to do, but for example checking if certain parameters are withing accepted ranges could become complex.
Would the branch you created already work in its simplest form (ie just choosing runner)?

tyll:
Changing the runner should work when using NetworkManager but the branch does not match the requirements to be merged.
I would be fine with only doing JSON validation in the beginning and other validation on a best-effort base or not at all. @cathay4t what do you think?

cathay4t:
[...]

PR #169, opened 2020-03-22
Team documentation

a-mere-peasant:
Add documentation for existing support for team type until it is configurable

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ the value of the `master` and assumes it will be present during a real
 run. That means, in presence of an invalid `master`, `--check` may
 signal success but the actual play run fails.
 
+The `team` type uses `roundrobin` as the `runner` configuration. No further configuration is supported at the moment.
 #### `type: vlan`
 
 Similar to `master`, the `parent` references the connection profile in the ansible


README.md
```diff
@@ -220,6 +220,7 @@ the value of the `master` and assumes it will be present during a real
 run. That means, in presence of an invalid `master`, `--check` may
 signal success but the actual play run fails.
 
+The `team` type uses `roundrobin` as the `runner` configuration.
```
tyll: @thom311 is this an NM or teamd default? Manpages for both do not seem to specify a default and the role does not set the runner at all.

thom311: the role doesn't support setting an options for the team. Hence, it creates a connection profile with "team.config" empty. That means, NM will call teamd with `-c ''` ... and teamd, will indeed then use "roundrobin": https://github.com/jpirko/libteam/blob/c7237377dead39ae4a711297203bacf7edb9fa41/teamd/teamd.c#L956

thom311: I think the addition to the readme is correct. However, we should also add some options, otherwise, it it doesn't seem very useful to me.

tyll: Good point. @a-mere-peasant please add a sentence like "No further configuration is supported at the moment.". Also the the branch needs to be rebased as described in https://git-scm.com/book/en/v2/Git-Branching-Rebasing

a-mere-peasant: >Good point. @a-mere-peasant please add a sentence like "No further configuration is supported at the moment.".

Ok,I'll add this .
Should there be an indication about 
https://github.com/tyll/network/tree/team_options
for people who want to use it ? or something referencing #97 ?

tyll: Good question. In my opinion, there is no need for the references since the README is meant for users and not developers.

Issue #172, opened 2020-03-24
Port tests_ethernet.yml to use the same format as the new tests


RMPR:
I'll tackle this issue.

PR #168, opened 2020-03-20
Port tests_bridge.yml to use the same format as the new tests

RMPR:
Addressing #167.

PR #173, opened 2020-03-24
Port tests_ethernet.yml to use the same format as the new tests

RMPR:
Address #172


tyll:
This needs to handle `down-profile.yml` as well as in #168.

tyll:
Followup is in #197

PR #178, opened 2020-03-30
sync with latest template

richm:
Just use the `python` command in the virtual environment, instead
of using `envpython`.  This solves some problems where the
attempt to resolve the `envpython` to the `syspython` by following
symlinks causes python not to be able find some packages installed
in the venv.

Add support for python 3.8-dev in travis.

Do not install ansible by default in all tox venvs for pytest
tests.  If the user needs to install ansible for pytests, the
user will add `ansible` to the new `pytest_extra_requirements.txt`
For python 2.6, which requires ansible 2.6, there is a special
`pytest26_extra_requirements.txt`, and an example
`ansible26_requirements.txt` which the user will use to specify
the extra dependencies for installing ansible 2.6 on python 2.6.

There are two extra dependencies added to the list of dependencies
installed on the travis machines when using python 2.6 - libffi-dev
and libssl-dev - these are required in order to install ansible 2.6
and its extra dependencies.

For python 2.6, an older version of tox must be used, which does not
have support for `commands_pre`, so allow `tests/setup_module_utils.sh`
to be run directly from `.travis/runpytest.sh` on those platforms.
the user will select to install ansible

Use the file `.yamllint_defaults.yml` for the base `yamllint`
configuration.  User customizations go into `.yamllint.yml`, which
extends `.yamllint_defaults.yml`.  This allows the use of `yamllint`
with IDEs that do not allow specifying command line arguments
for the `yamllint` command.

Only run coveralls in travis in the py35 environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
     - python: 3.6
     - python: 3.7
     - python: 3.8
+    - python: 3.8-dev
 
 services:
   - docker
--- a/.travis/runcoveralls.sh
+++ b/.travis/runcoveralls.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+
+# Reports coverage results using coveralls. The aim of this script is to
+# provide a unified way to reporting coverage results across all linux system
+# roles projects.
+
+# The given command line arguments are passed to coveralls.
+
+# Environment variables:
+#
+#   LSR_PUBLISH_COVERAGE
+#     if the variable is empty or unset, nothing will be published; if the
+#     variable has its value set to 'strict', the reporting is performed in
+#     strict mode, so situations like missing data to be reported are treated
+#     as errors; if the value of this variable is 'debug', coveralls is run in
+#     debug mode (see coveralls debug --help); other values cause that coverage
+#     results will be reported normally
+#   LSR_TESTSDIR
+#     a path to directory where tests and tests artifacts are located; if unset
+#     or empty, this variable is set to ${TOPDIR}/tests; this path should
+#     already exists and be populated with tests artifacts before the script
+#     starts performing actions on it
+
+set -e
+
+ME=$(basename $0)
+SCRIPTDIR=$(readlink -f $(dirname $0))
+
+. ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
+
+# Publish the results only if it is desired.
+if [[ -z "${LSR_PUBLISH_COVERAGE}" ]]; then
+  lsr_info "${ME}: Publishing coverage report is not enabled. Skipping."
+  exit 0
+fi
+
+LSR_TESTSDIR=${LSR_TESTSDIR:-${TOPDIR}/tests}
+
+# Ensure we are in $LSR_TESTSDIR. It is supposed that if a user wants to submit
+# tests results, $LSR_TESTSDIR always exists.
+cd ${LSR_TESTSDIR}
+
+# For simplicity, we suppose that coverage core data file has name .coverage
+# and it is situated in $LSR_TESTSDIR. Similarly for .coveragerc.
+COVERAGEFILE='.coverage'
+COVERAGERCFILE='.coveragerc'
+
+# In case there is no $COVERAGEFILE, there is nothing to report. If we are
+# running in strict mode, treat this situation as error.
+if [[ ! -s ${COVERAGEFILE} ]]; then
+  NO_COVERAGEFILE_MSG="${COVERAGEFILE} is missing or empty"
+  if [[ "${LSR_PUBLISH_COVERAGE}" == "strict" ]]; then
+    lsr_error "${ME} (strict mode): ${NO_COVERAGEFILE_MSG}!"
+  fi
+  lsr_info "${ME}: ${NO_COVERAGEFILE_MSG}, nothing to publish."
+  exit 0
+fi
+
+# Create $COVERAGERCFILE file with a [paths] section. From the official docs:
+#
+#   The first value must be an actual file path on the machine where the
+#   reporting will happen, so that source code can be found. The other values
+#   can be file patterns to match against the paths of collected data, or they
+#   can be absolute or relative file paths on the current machine.
+#
+# So in our $COVERAGERCFILE file we make both locations to point to the
+# project's top directory.
+cat > ${COVERAGERCFILE} <<EOF
+[paths]
+source =
+    ..
+    $(readlink -f ..)
+EOF
+
+# Rename $COVERAGEFILE to ${COVERAGEFILE}.merge. With this trick, coverage
+# combine applies configuration in $COVERAGERCFILE also to $COVERAGEFILE.
+mv ${COVERAGEFILE} ${COVERAGEFILE}.merge
+python -m coverage combine --append
+
+MAYBE_DEBUG=""
+if [[ "${LSR_PUBLISH_COVERAGE}" == "debug" ]]; then
+  MAYBE_DEBUG=debug
+fi
+
+set -x
+coveralls ${MAYBE_DEBUG} "$@"
--- a/.travis/utils.sh
+++ b/.travis/utils.sh
@@ -136,14 +136,61 @@ function lsr_compare_pythons() {
 }
 
 ##
+# lsr_get_system_python
+#
+# Return the system python, or /usr/bin/python3 if nothing
+# else can be found in a standard location.
+function lsr_get_system_python() {
+  local syspython=$(command -pv python3)
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python)
+  fi
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python2)
+  fi
+  echo ${syspython:-/usr/bin/python3}
+}
+
+##
 # lsr_venv_python_matches_system_python [$1] [$2]
 #
 #   $1 - command or full path to venv Python interpreter (default: python)
 #   $2 - command or full path to the system Python interpreter
-#        (default: /usr/bin/python3)
+#        (default: /usr/bin/python3 or /usr/bin/python or /usr/bin/python2)
 #
 # Exit with 0 if virtual environment Python version matches the system Python
 # version.
 function lsr_venv_python_matches_system_python() {
-  lsr_compare_pythons ${1:-python} -eq ${2:-/usr/bin/python3}
+  local syspython="${2:-$(lsr_get_system_python)}"
+
+  lsr_compare_pythons ${1:-python} -eq $syspython
 }
+
+##
+# lsr_setup_module_utils [$1] [$2]
+#
+#   $1 - path to the ansible/module_utils/ directory in the venv
+#        assumes ansible has been installed in the venv
+#        defaults to env var $SRC_MODULE_UTILS_DIR
+#   $2 - path to the local module_utils/ directory for the role
+#        defaults to env var $DEST_MODULE_UTILS_DIR
+#
+# Exit with 0 if virtual environment Python version matches the system Python
+# version.
+function lsr_setup_module_utils() {
+  local srcdir=${1:-$SRC_MODULE_UTILS_DIR}
+  local destdir=${2:-$DEST_MODULE_UTILS_DIR}
+  if [ -n "$srcdir" -a -d "$srcdir" -a -n "$destdir" -a -d "$destdir" ]; then
+    bash $TOPDIR/tests/setup_module_utils.sh "$srcdir" "$destdir"
+  fi
+}
+
+# set TOPDIR
+ME=${ME:-$(basename $0)}
+SCRIPTDIR=${SCRIPTDIR:-$(readlink -f $(dirname $0))}
+TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
+
+# Local Variables:
+# mode: Shell-script
+# sh-basic-offset: 2
+# End:
--- a/.yamllint_defaults.yml
+++ b/.yamllint_defaults.yml
@@ -1,0 +1,14 @@
+# SPDX-License-Identifier: MIT
+---
+ignore: |
+  /.tox/
+extends: default
+rules:
+  braces:
+    max-spaces-inside: 1
+    level: error
+  brackets:
+    max-spaces-inside: 1
+    level: error
+  truthy: disable
+  document-start: disable
--- a/pytest26_extra_requirements.txt
+++ b/pytest26_extra_requirements.txt
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+
+# Write extra requirements for running pytest with python2.6 here:
+# NOTE: If your pytests need ansible, use the requirements from
+# ansible26_requirements.txt, either by listing it like this:
+# -ransible26_requirements.txt
+# or by listing the dependencies from that file explicitly
--- a/tox.ini
+++ b/tox.ini
@@ -13,15 +13,18 @@ basepython = python3
 deps =
     py{26,27,36,37,38}: pytest-cov
     py{27,36,37,38}: pytest>=3.5.1
+    py{27,36,37,38}: -rpytest_extra_requirements.txt
     py{26,27}: mock
     py26: pytest
+    py26: -rpytest26_extra_requirements.txt
 
 [base]
-system_python = /usr/bin/python3
 passenv = *
 setenv =
     PYTHONPATH = {toxinidir}/library:{toxinidir}/module_utils
     LC_ALL = C
+    SRC_MODULE_UTILS_DIR = {envsitepackagesdir}/ansible/module_utils
+    DEST_MODULE_UTILS_DIR = {toxinidir}/module_utils
 changedir = {toxinidir}/tests
 covtargets = --cov={toxinidir}/library --cov={toxinidir}/module_utils
 pytesttarget = unit
@@ -43,7 +46,7 @@ changedir = {[base]changedir}
 whitelist_externals =
     {[base]whitelist_externals}
 commands =
-    bash {[base]runpytest} {envpython} \
+    bash {[base]runpytest} \
          --durations=5 \
          {[base]covtargets} \
          --cov-report=html:htmlcov-py26 \
@@ -61,7 +64,7 @@ changedir = {[base]changedir}
 whitelist_externals =
     {[base]whitelist_externals}
 commands =
-    bash {[base]runpytest} {envpython} \
+    bash {[base]runpytest} \
          --durations=5 \
          {[base]covtargets} \
          --cov-report=html:htmlcov-py27 \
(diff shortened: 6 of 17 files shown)


.yamllint_custom.yml
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: MIT
+---
+extends: .yamllint.yml
```
tyll: IMHO it makes more sense to have a file like `.yamllint_defaults.yml` that contains the defaults and a `.yamllint.yml` that extends it. This makes sure that yamllint without special configuration picks up the custom configuration instead of only the defaults. This would be useful for running yamllint inside of IDEs/editors.

pcahyna: This is a good point.

.travis.yml
```diff
@@ -4,7 +4,7 @@ dist: xenial
 language: python
 env:
   global:
-    - LSR_ANSIBLES='ansible==2.6.* ansible==2.7.* ansible==2.8.*'
```
tyll: The Ansible versions can be updated without discussion, therefore I added then in #179

tox.ini
```diff
@@ -10,14 +10,19 @@ skip_missing_interpreters = true
 [testenv]
 basepython = python3
 # List common dependencies for Python interpreters here:
+# note that we dropped support for ansible 2.6, and ansible
+# 2.7 dropped support for py26 - so we only do ansible
+# related testing on py27 and later
 deps =
+    py{27,36,37,38}: ansible
```
tyll: This seems to be a new dependency since it is not removed from the py26 env (it does not seem to be specified, there).

richm: ansible does not work with python 2.6
ansible is needed for pytest

pcahyna: Are you sure? https://docs.ansible.com/ansible/latest/dev_guide/developing_python_3.html#minimum-version-of-python-3-x-and-python-2-x "Module-side, we support Python 3.5 or greater and Python 2.6 or greater." And this seems to be used for module code testing, not running Ansible itself.

pcahyna: (maybe I don't understand why Ansible is needed for pytest.)

richm: > Are you sure? https://docs.ansible.com/ansible/latest/dev_guide/developing_python_3.html#minimum-version-of-python-3-x-and-python-2-x "Module-side, we support Python 3.5 or greater and Python 2.6 or greater." And this seems to be used for module code testing, not running Ansible itself.

I tried it - it doesn't work - I tried installing the latest ansible 2.7.x in a py26 tox venv and it complained about the version of python.  Please try it.  NOTE that I understand what the ansible documentation says, I'm suggesting that it may be incorrect.

richm: > (maybe I don't understand why Ansible is needed for pytest.)

in lsr, pytest is (almost?) always used to do unit testing for module code, which means the unit test code needs ansible.

.yamllint_custom.yml
richm: So the presence of `_custom` in the name of the file doesn't lead you to believe that this is the file that should have local customizations?

tox.ini
pcahyna: when did it complain? I would expect it to complain when you try to run the `ansible` command, bit not when you just import the required Python helper packages in a module. Otherwise our roles would not work against RHEL 6.

richm: It complained when I tried to `pip install ansible` in the tox py26 venv.  Again, please don't take my word for it, please try it for yourself, perhaps I did something wrong.

.yamllint_custom.yml
tyll: It makes me understand that it is meant for local customomizations but the `yamllint` tool does not know this. It will only load `.yamllint.yml` by default. Therefore it would be better to have the customizations in `.yamllint.yml` so that `yamllint` can be called by IDEs without arguments specific to system roles.

richm: ah - so it is only for IDE support where the IDE cannot pass command line arguments to the `yamllint` command

tox.ini
tyll: Usually, the roles only import ansible because the module_utils are in the ansible namespace and to create an AnsibleModule object. As long as we need to support Python 2.6 for CentOS/RHEL 6, we should also tests against 2.6. So if the unit tests require ansible to test more than just the AnsibleModule object creation, this needs to be addressed or proper testing in 2.6. This might not the right place to discuss this because the unit tests in the network role do not require ansible for their tests.

.yamllint_custom.yml
tyll: yes, or just to avoid having to configure the IDE specifically for each project, for example for one-time contributors.

tox.ini
```diff
@@ -267,8 +273,8 @@ max-line-length = 88
 [travis]
 python =
   2.6: py26,custom
-  2.7: py27,flake8,pylint,custom
-  3.5: coveralls,custom
-  3.6: py36,black,yamllint,custom
+  2.7: py27,coveralls,flake8,pylint,custom
```
tyll: Is this change correct? I believe this change was part of another sync attempt and it does not make sense in the current configuration since coveralls needs to run only once.

tox.ini
pcahyna: Ah, it is already the installation via pip that fails. I see that it would be a problem, yes.
@tyll but how to import module_utils if ansible (which contains it) can't be installed via pip?

tyll:
As with the last sync attempt, it would be nicer to sync the commits for the actual changes to be able to review and approve them together with their justification from the commit message. There are changes that are good like changing the Ansible versions, some changes that need discussion like changing the yamllint config file.

pcahyna:
> As with the last sync attempt, it would be nicer to sync the commits for the actual changes to be able to review and approve them together with their justification from the commit message. There are changes that are good like changing the Ansible versions, some changes that need discussion like changing the yamllint config file.

I think we discussed with @i386x the possibility to do this (sync commit-by-commit) and determined that it was not that important because (IIRC) after the initial sync, new changes would be small. Maybe this was a wrong assumption.

richm:
> As with the last sync attempt, it would be nicer to sync the commits for the actual changes to be able to review and approve them together with their justification from the commit message. There are changes that are good like changing the Ansible versions, some changes that need discussion like changing the yamllint config file.

In the future, for template changes, I'll submit the PR to the `template` repo and the `network` repo at the same time so we can get a wider audience for reviews, and make sure the changes work for `network` before merging them for `template`.

tox.ini
tyll: The tests mock the ansible module: https://github.com/linux-system-roles/network/blob/master/tests/unit/test_nm_provider.py#L17

richm: in `kernel_settings` and `storage` and possibly other roles, `ansible` is used directly in the unit tests.  I'm not sure if `ansible` is needed by `pylint` or `flake` tests.
If it is going to be a problem to have `ansible` always installed for every python environment, then I'll figure out a way to refactor this, to make the `ansible` dependency per-tox environment, or even a per-repo dependency installed via some sort of custom requirements file.

tox.ini
richm: I'll change `coveralls` to be executed only in the 3.6 environment.

@tyll Also - there is a new `.travis/runcoveralls.sh` in the template - can this be used to replace the network
```
    bash -c 'cd ..; bash .travis/fix-coverage.sh; cd -'
    coveralls
```
?
It looks as though there is some creation and fixup of .coveragerc in template `.travis/runcoveralls.sh` but I don't know if this is the same as what network is doing, or if we need some additional work to template `.travis/runcoveralls.sh` in order to be suitable for network.

tox.ini
tyll: Do the kernel_settings and storage role target CentOS 6, too, or only newer versions? Then it would not matter for them. But the reasoning in the comment does not make sense for other roles that also target CentOS6/Python 2.6.

richm: I'm coming up with a much more comprehensive way to specify that ansible is a dependency for pytests for different platforms, which also takes into consideration that installing ansible in a tox venv on 2.6 has additional requirements, both on the ubuntu platform side in the form of additional C library dependencies, and on the `pip` side in the form of older versions of some dependencies.  This will give the role developer the flexibility to decide if `ansible` is a dependency for the pytest tests, and which version of ansible to use.

richm:
@tyll I think I have addressed all of your review feedback.  This PR now contains the changes from the latest `template` as well as the new changes from PR https://github.com/linux-system-roles/template/pull/18 which address your review feedback.

richm:
updated https://github.com/linux-system-roles/auto-maintenance/pull/8 also

.travis/utils.sh
```diff
@@ -135,15 +135,43 @@ function lsr_compare_pythons() {
...
-#        (default: /usr/bin/python3)
+#        (default: /usr/bin/python3 or /usr/bin/python or /usr/bin/python2)
 #
 # Exit with 0 if virtual environment Python version matches the system Python
 # version.
 function lsr_venv_python_matches_system_python() {
-  lsr_compare_pythons ${1:-python} -eq ${2:-/usr/bin/python3}
+  local syspython="${2:-$(lsr_get_system_python)}"
+
+  lsr_compare_pythons ${1:-python} -eq ${syspython:-/usr/bin/python3}
```
tyll: syspython will always be defined since `lsr_get_system_python()` defaults to /usr/bin/python3 already, so defaulting to /usr/bin/python3 here again seems to redundant.

.yamllint_defaults.yml
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: MIT
```
tyll: It seems to me that this is missing a change for the .yamllint file to extend the .yamllint_defaults.yml file.

richm: Good catch.  The auto-maintenance sync-template.sh script was changed to only copy `.yamllint.yml` if it didn't exist, since that is now the file used for local customizations.  Since network already had a .yamllint.yml, I had to manually copy it from template.

pytest26_extra_requirements.txt
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: MIT
+
+# Write extra requirements for running pytest with python2.6 here:
+# NOTE: If your pytests need ansible, use the requirements from
+# ansible26_requirements.txt
```
tyll: Just wondering, does this mean to copy the contents from ansible26_requirements here or does adding `-r ansible26_requirements.txt` work here?

tox.ini
```diff
@@ -159,22 +164,24 @@ deps =
 whitelist_externals =
     {[base]whitelist_externals}
 commands =
-    bash {toxinidir}/.travis/runflake8.sh {envpython} --statistics {posargs} .
+    bash {toxinidir}/.travis/runflake8.sh --exclude=.venv,.tox --statistics {posargs} .
 
 [testenv:yamllint]
 deps = yamllint
+whitelist_externals =
```
tyll: This env does not use bash, so the `whitelist_externals` does not seem to be needed here.

.travis/runcoveralls.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Reports coverage results using coveralls. The aim of this script is to
+# provide a unified way to reporting coverage results across all linux system
+# roles projects.
+
+# The given command line arguments are passed to coveralls.
+
+# Environment variables:
+#
+#   LSR_PUBLISH_COVERAGE
```
tyll: It seems that the change to set LSR_PUBLISH_COVERAGE to strict is missing. Should it be added to config.sh?

tox.ini
richm: it is needed because of the `command_pre` in `[testenv]` which applies to every section . . . so I suppose the right answer here is to remove the `command_pre`.  This means `setup_module_utils.sh` will not be done automatically by default for every test env.  I suppose I could add a `command_pre` for `setup_module_utils.sh` to every env, but that seems fragile and redundant, because in the case of network, it isn't needed for `pytest`.  So I guess the right answer is to add a hook in the scripts that might need to set up module_utils for testing, such as `.travis/runpytest.sh`, `.travis/runpylint.sh`, `.travis/runflake8.sh`, that is controlled by `.travis/config.sh` and allows the user to specify that module_utils needs to be setup for testing.

.travis/runcoveralls.sh
richm: What change would that be?  The initial commit of runcoveralls.sh to the template is here - https://github.com/linux-system-roles/template/commit/57513dac636bb8fef4e9fcc3c95066f6b704ed16#diff-9c9e55de5f8cf641950dc28851dbd328 - I don't see where the default value of `LSR_PUBLISH_COVERAGE` would have been `strict`

richm: I believe the default value in the template should be `strict`.

pytest26_extra_requirements.txt
richm: See if the comment in the pytest26_extra_requirements.txt in the latest commit answers your question

tox.ini
richm: done - see latest commit

.travis/runcoveralls.sh
richm: done

tox.ini
pcahyna: > The tests mock the ansible module: https://github.com/linux-system-roles/network/blob/master/tests/unit/test_nm_provider.py#L17

Hmm, is this approach going to be scalable and maintainable if used for all the roles that have embedded modules?

richm: > Hmm, is this approach going to be scalable and maintainable if used for all the roles that have embedded modules?

My approach is to give unit tests for modules an actual run-time environment that is as close to a production environment as possible, with a real ansible with the role module_utils/ code in the real `ansible.module_utils.*`module namespace, so that no `sys.path` munging, or module `Mock`-ing, is necessary.

That being said, I think the approach that `network` takes might be useful for certain unit test cases where you want to test a single method or small group of methods, and nothing else, and find that dragging in all of ansible is an impediment.

But I still think that unit testing with a real ansible is a good thing that we should do by default.

richm:
grr - coveralls test in 3.5 - I guess I need to become quite familiar with coveralls and figure out what to do here :-P

richm:
> grr - coveralls test in 3.5 - I guess I need to become quite familiar with coveralls and figure out what to do here :-P

@tyll - perhaps you can tell me what this is supposed to do, and how I'm supposed to fix this . . .

richm:
> > grr - coveralls test in 3.5 - I guess I need to become quite familiar with coveralls and figure out what to do here :-P
> 
> @tyll - perhaps you can tell me what this is supposed to do, and how I'm supposed to fix this . . .

Maybe I shouldn't use `strict` by default?  This is what `network` is doing currently: https://travis-ci.com/github/linux-system-roles/network/jobs/306713129
```
coveralls run-test: commands[0] | bash -c 'cd ..; bash .travis/fix-coverage.sh; cd -'
+ cat
+ mv .coverage .coverage.merge
mv: cannot stat '.coverage': No such file or directory
/home/travis/build/linux-system-roles/network/tests
coveralls run-test: commands[1] | coveralls
Submitting coverage to coveralls.io...
Coverage submitted!
Job #558.3
https://coveralls.io/jobs/60690839
___________________________________ summary ____________________________________
  custom: commands succeeded
  coveralls: commands succeeded
```
Note that there is no `.coverage` file but the current `network` `tox` task ignores that and successfully submits coverage to coveralls.io.

It seems that coveralls should be used to collect the coverage information generated by pytest on various platforms, and coverage information generated by running the molecule tests with a special ```
    ansible_python_interpreter:
      "{{ coverage }} run -p --include *ansible_module_{{ coverage_module }}.py"
```
then collect all of the coverage information from the various hosts/environments and merge them together.  But that doesn't appear to be how `tox` + `travis` is working.  For one, in `travis`, all of the pytest tests are run on possibly separate VMs.  For another, the `coveralls` test isn't run after all of the other tests have run and generated coverage information to merge and publish.

So I guess I don't know what the `tox` `coveralls` test is supposed to do, both inside and outside of `travis`.

**Comment by richm** [comment]
. . . so until I figure out what `coveralls` is supposed to do in the context of `tox` or `travis`, I have disabled `LSR_PUBLISH_COVERAGE=strict`

File: .travis/runcoveralls.sh (hunk 9 above)
Comment (i386x) [reply .travis/runcoveralls.sh]: The default value of `LSR_PUBLISH_COVERAGE` should be left empty, since afaik at now only network role is using coverage & coveralls. Setting the `LSR_PUBLISH_COVERAGE` is done in `.travis/config.sh`, so to enable coverage reporting just put `LSR_PUBLISH_COVERAGE=strict` or `LSR_PUBLISH_COVERAGE=yes` there. Just a note, I am not sure what are @tyll expectations on coverage reporting, there is a discussion in #120 about it. Similar PR with .travis/runcoveralls.sh is also #152.

**Comment by richm** [comment]
[citest bad]

File: .travis/runcoveralls.sh (hunk 9 above)
Comment (richm) [reply .travis/runcoveralls.sh]: Ok, I see.  The coveralls stuff is still WIP.  I suggest then that this PR is ok, and most of the work being done for https://github.com/linux-system-roles/network/pull/120 and https://github.com/linux-system-roles/network/pull/152 should really go into template (except the .travis/config.sh changes), then sync'd back to network (and the other roles).

Comment (tyll) [reply .travis/runcoveralls.sh]: > The default value of `LSR_PUBLISH_COVERAGE` should be left empty, since afaik at now only network role is using coverage & coveralls. Setting the `LSR_PUBLISH_COVERAGE` is done in `.travis/config.sh`, so to enable coverage reporting just put `LSR_PUBLISH_COVERAGE=strict` or `LSR_PUBLISH_COVERAGE=yes` there. Just a note, I am not sure what are @tyll expectations on coverage reporting, there is a discussion in #120 about it. Similar PR with `.travis/runcoveralls.sh` is also #152.

I guess the value for the network role should be `yes` then, because it should be enabled and (even if it is broken) and not accidentally be disabled IMHO. Please note, the value `yes` is not mentioned in the documentation for the variable, it only mentions emtpy, `strict` and `debug`. This is why I proposed `strict`.

**Comment by i386x** [comment]
> grr - coveralls test in 3.5 - I guess I need to become quite familiar with coveralls and figure out what to do here :-P

The purpose of fail in 3.5 (https://github.com/linux-system-roles/network/commit/151cefcbd08399f4c3b75d0747fc753d29552578) is missing pytest run before coverage, which ends up in missing `.coverage` file and hence the error. The correct coveralls placement should be imho
```
2.7: py27,coveralls,flake8,pylint,custom
3.5: custom
3.6: py36,coveralls,black,yamllint,custom
```

i386x:
> Coverage submitted!
> Job #558.3
> https://coveralls.io/jobs/60690839
> 
> Note that there is no `.coverage` file but the current `network` role `tox` task ignores that and successfully submits coverage to coveralls.io.

Despite on the fact that `coveralls` wrote that coverage result were submitted, there are no data displayed in https://coveralls.io/jobs/60690839 (compare with https://coveralls.io/builds/28545429).

richm:
I am removing this PR in favor of https://github.com/linux-system-roles/network/pull/181 and https://github.com/linux-system-roles/network/pull/182, and more to come once these are merged

PR #181, opened 2020-03-31
use python instead of envpython

richm:
Many of the tests and tasks invoked from `tox` would try to use
the path to the `python` executable in the tox virtualenv, and
in some cases, would try to follow the symlink if the `python`
command were a symlink, in some cases, following the symlink to
the system python, and trying to use the absolute path.  This
would fail in some cases because when using the system python
with the absolute path, the python modules installed in the
tox venv could not be found.  So instead, just use the `python`
command from the venv with no path.

also enable travis 3.8-dev testing, since this fix allows that
platform to work correctly.

### DIFF
--- a/.travis/utils.sh
+++ b/.travis/utils.sh
@@ -136,14 +136,35 @@ function lsr_compare_pythons() {
 }
 
 ##
+# lsr_get_system_python
+#
+# Return the system python, or /usr/bin/python3 if nothing
+# else can be found in a standard location.
+function lsr_get_system_python() {
+  local syspython=$(command -pv python3)
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python)
+  fi
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python2)
+  fi
+  if [[ -z "$syspython" ]]; then
+    lsr_error Could not determine system python path
+  fi
+  echo $syspython
+}
+
+##
 # lsr_venv_python_matches_system_python [$1] [$2]
 #
 #   $1 - command or full path to venv Python interpreter (default: python)
 #   $2 - command or full path to the system Python interpreter
-#        (default: /usr/bin/python3)
+#        (default: system python as determined by lsr_get_system_python())
 #
 # Exit with 0 if virtual environment Python version matches the system Python
 # version.
 function lsr_venv_python_matches_system_python() {
-  lsr_compare_pythons ${1:-python} -eq ${2:-/usr/bin/python3}
+  local syspython="${2:-$(lsr_get_system_python)}"
+
+  lsr_compare_pythons ${1:-python} -eq $syspython
 }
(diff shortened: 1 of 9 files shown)


richm:
[citest bad]

.travis/utils.sh
```diff
@@ -135,15 +135,33 @@ function lsr_compare_pythons() {
...
+  echo ${syspython:-/usr/bin/python3}
+}
+
 ##
 # lsr_venv_python_matches_system_python [$1] [$2]
 #
 #   $1 - command or full path to venv Python interpreter (default: python)
 #   $2 - command or full path to the system Python interpreter
-#        (default: /usr/bin/python3)
+#        (default: /usr/bin/python3 or /usr/bin/python or /usr/bin/python2)
```
tyll, line 159: This comment about the defaults is not accurate, since it checks python3, python and python2 before falling back to /usr/bin/python3 and they could be in different locations than /usr/bin - I believe it is actually the case for Travis that they are not in /usr/bin.

```suggestion
#        (default: system python as determined by lsr_get_system_python())
```

tyll:
If this change also fixes using python 3.8-dev from Travis, the change to enable it again could be included here as well.

.travis/utils.sh
```diff
@@ -135,15 +135,33 @@ function lsr_compare_pythons() {
...
+# else can be found in a standard location.
+function lsr_get_system_python() {
+  local syspython=$(command -pv python3)
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python)
+  fi
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python2)
+  fi
+  echo ${syspython:-/usr/bin/python3}
```
tyll, line 151: Sorry, I had to think about this. It seems to be rather odd that /usr/bin/python3 would be a correct choice without it being in PATH. If it is in PATH, it would be set already by line 144. Technically, it should probably fail in this case instead of assuming /usr/bin/python3 to be correct. But maybe this is a wrong assumption. It does not seem to me that this is that important, just wanted to raise it. I am also fine with merging it as-is, since it will probably not hurt. What is your opinion?

richm: changed to this:
```
  if [[ -z "$syspython" ]]; then
    lsr_error Could not determine system python path
  fi
  echo $syspython
```

richm:
> Please see my note, not a blocker. LGTM overall. It just needs to be rebased to master AFAICS. Thank you for adding 3.8-dev.

Fixed and rebased

.travis/utils.sh
```diff
@@ -135,15 +135,36 @@ function lsr_compare_pythons() {
...
+# else can be found in a standard location.
+function lsr_get_system_python() {
+  local syspython=$(command -pv python3)
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python)
+  fi
+  if [[ -z "$syspython" ]]; then
+    syspython=$(command -pv python2)
+  fi
+  if [[ -z "$syspython" ]]; then
```
tyll, line 151: awesome, thank you!

PR #182, opened 2020-03-31
changes to yamllint from template

richm:


richm:
[citest bad]

richm:
[citest bad]

richm:
[citest bad]

tyll:
FYI: I updated the commit message to describe the actual change instead of only the fact that it is a sync from the template.

PR #183, opened 2020-04-01
python 2.6 support for pytest and ansible 2.6

richm:
This doesn't affect network because network does not use ansible
for pyunit testing, so this commit is primarily to be in sync with
the template and other repos.

However, if you need `ansible` and `mock` for unit testing or
linting, see `pytest_extra_requirements.txt`

### DIFF
--- a/ansible_pytest_extra_requirements.txt
+++ b/ansible_pytest_extra_requirements.txt
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+
+# ansible and dependencies for all supported platforms
+ansible ; python_version > "2.6"
+ansible<2.7 ; python_version < "2.7"
+idna<2.8 ; python_version < "2.7"
+PyYAML<5.1 ; python_version < "2.7"
--- a/pytest_extra_requirements.txt
+++ b/pytest_extra_requirements.txt
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+
+# Write extra requirements for running pytest here:
+# If you need ansible then uncomment the following line:
+#-ransible_pytest_extra_requirements.txt
+# If you need mock then uncomment the following line:
+mock ; python_version < "3.0"
--- a/tox.ini
+++ b/tox.ini
@@ -13,8 +13,8 @@ basepython = python3
 deps =
     py{26,27,36,37,38}: pytest-cov
     py{27,36,37,38}: pytest>=3.5.1
-    py{26,27}: mock
     py26: pytest
+    py{26,27,36,37,38}: -rpytest_extra_requirements.txt
 
 [base]
 system_python = /usr/bin/python3
--- a/.travis/preinstall
+++ b/.travis/preinstall
@@ -20,6 +20,11 @@ if lsr_venv_python_matches_system_python; then
   LSR_EXTRA_PACKAGES='python3-selinux'
 fi
 
+# extra packages needed with python 2.6 and ansible 2.6
+if lsr_check_python_version python -lt 2.7 ; then
+  LSR_EXTRA_PACKAGES="$LSR_EXTRA_PACKAGES libffi-dev libssl-dev"
+fi
+
 . ${SCRIPTDIR}/config.sh
 
 # Install extra dependencies.


pcahyna:
Why Ansible 2.6? It is not supported anymore. I hope we are not using it. At least here, it is not being used:
https://github.com/richm/network/blob/4550afe27b038518566a29bbf5e42d9f08c4f75c/.travis.yml#L7

richm:
> Why Ansible 2.6? It is not supported anymore. I hope we are not using it. At least here, it is not being used:
> https://github.com/richm/network/blob/4550afe27b038518566a29bbf5e42d9f08c4f75c/.travis.yml#L7

Do we still need to support EL6?  Can we assume we will not support ansible<2.7 on EL6?

pcahyna:
Yes, we need to support EL6 as managed node. How is Ansible 2.6 related to EL6?

richm:
> Yes, we need to support EL6 as managed node. How is Ansible 2.6 related to EL6?

For some reason I thought we had to support ansible 2.6 on EL6?  Is this true?  Or can we assume we will not support ansible 2.6 on EL6, and only support ansible 2.7 and later on EL6?

pcahyna:
Ah, I see, Ansible 2.6 was the last release to support EL6. https://dl.fedoraproject.org/pub/epel/6/SRPMS/Packages/a/ansible-2.6.20-1.el6.src.rpm , 2.7 and later would not be supported. I am not aware of any requirement to support our roles on EL6. We need to support running them against EL6, though, which is still supported by newer Ansibles and requires the module code to be Python 2.6 compatible.

pcahyna:
> Ah, I see, Ansible 2.6 was the last release to support EL6.

I mean, supported being installed on EL6. All ansible versions support being run against EL6 as managed node.

richm:
ah, ok, i see.  So we still need CI testing with EL6, but not tox/travis testing.

pcahyna:
We need tox/travis testing with python26 for roles that embed Python modules. Module unit tests are run from travis/tox and they should be executed under Python 2.6 to ensure that the module stays 2.6 compatible.
It does not need EL6 or Ansible 2.6 though. Just Python 2.6.

tox.ini
```diff
@@ -13,8 +13,10 @@ basepython = python3
 deps =
     py{26,27,36,37,38}: pytest-cov
     py{27,36,37,38}: pytest>=3.5.1
+    py{27,36,37,38}: -rpytest_extra_requirements.txt
     py{26,27}: mock
```
tyll, line 17: I guess the mock dependency is there for a similar reason (the mock module is part of Python 3) to be able to mock ansible. Not sure if this should be split out, too. But also seems to add a lot of complexity to move all these possible customizations outside of tox.ini.

richm: Yes, I think `mock` is also an optional component for unit testing.  I suppose we could do something like this:
```
    py{3*}: -rpy3_pytest_extra_requirements.txt
    py{2*}: -rpy2_pytest_extra_requirements.txt
    py{26}: -rpy26_pytest_extra_requirements.txt
```
where the last one is renaming the current `pytest26_extra_requirements.txt` file.  (It isn't for pytest version 2.6, it is extra dependencies for pytest when using py26).

tyll: The issue I see is that the amount of files is becoming excessive. With tox.ini there was initially one nice config file but it is becoming more and more.

For this specific problem, environment markers might help: https://www.python.org/dev/peps/pep-0508/#examples - I am not sure if they work well when also supporting Python 2.x

richm: ah, I see - so if it works, have a single `-rpytest_extra_requirements.txt` which has something like this:
```
# my pytesting needs mock and ansible
mock ; python_version < "3.0"
ansible ; python_version > "2.6"
ansible<2.7 ; python_version < "2.7"
idna<2.8 ; python_version < "2.7"
PyYAML<5.1; python_version < "2.7"
```
?
If so, looks like it works as expected on python 2.6 and python 3.8 (at least, the ones that travis CI uses, on ubuntu xenial).

tyll: yes, I think this looks nicer

pytest_extra_requirements.txt
```diff
@@ -0,0 +1,5 @@
+# SPDX-License-Identifier: MIT
+
+# Write extra requirements for running pytest here:
+# If you need ansible/mock then uncomment the following line:
+#-ransible_pytest_extra_requirements.txt
```
tyll, line 5: This file needs this AFAIU:
`mock; python_version < "3.0"`

ansible_pytest_extra_requirements.txt
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: MIT
+
+# needs mock and ansible
+mock ; python_version < "3.0"
```
tyll, line 4: Mock is not a dependency for Ansible

richm: I guess we need to have 3 files:
pytest_extra_requirements.txt - in the template it is empty - projects can customize to add mock and/or ansible
ansible_pytest_extra_requirements.txt - contains ansible and its dependencies (everything except `mock` from the current ansible_pytest_extra_requirements.txt)
mock_pytest_extra_requirements.txt - contains the `mock` package from the current ansible_pytest_extra_requirements.txt

The comment in `pytest_extra_requirements.txt` will change to say something like "if you use ansible in your pytest, add -ransible_pytest_extra_requirements.txt, and if you use mock, add -rmock_pytest_extra_requirements.txt"

ok?

tyll: > I guess we need to have 3 files:
> pytest_extra_requirements.txt - in the template it is empty - projects can customize to add mock and/or ansible

Yes, this file would be empty in the template but contain `mock` for the network role.

> ansible_pytest_extra_requirements.txt - contains ansible and its dependencies (everything except `mock` from the current ansible_pytest_extra_requirements.txt)

Sounds good.

> mock_pytest_extra_requirements.txt - contains the `mock` package from the current ansible_pytest_extra_requirements.txt

I do not think we need a file just for mock. Any role that needs mock can just add it to the `pytest_extra_requirements.txt` file.

> The comment in `pytest_extra_requirements.txt` will change to say something like "if you use ansible in your pytest, add -ransible_pytest_extra_requirements.txt, and if you use mock, add -rmock_pytest_extra_requirements.txt, and if you use both, add both -ransible_pytest_extra_requirements.txt and -rmock_pytest_extra_requirements.txt"

This only needs to mention the ansible convenience file.

PR #184, opened 2020-04-01
allow more overrides of template code

richm:
NOTE: I don't think this commit directly affects network, since
network doesn't use any of these - this PR is more for parity
with template and other repos.

Centralize the setting of `ME` and `TOPDIR` in utils.sh

Add support for `RUN_FLAKE8_EXTRA_ARGS` and `RUN_BLACK_EXTRA_ARGS`

### DIFF
--- a/.travis/runflake8.sh
+++ b/.travis/runflake8.sh
@@ -10,6 +10,9 @@
 #
 #   RUN_FLAKE8_DISABLED
 #     if set to an arbitrary non-empty value, flake8 will be not executed
+#
+#   RUN_FLAKE8_EXTRA_ARGS
+#     any extra command line arguments to provide e.g. --ignore=some,errs
 
 set -e
 
@@ -25,4 +28,4 @@ if [[ "${RUN_FLAKE8_DISABLED}" ]]; then
 fi
 
 set -x
-python -m flake8 "$@"
+python -m flake8 ${RUN_FLAKE8_EXTRA_ARGS:-} "$@"
--- a/.travis/runpytest.sh
+++ b/.travis/runpytest.sh
@@ -14,7 +14,6 @@ set -e
 
 ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
-TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
 
 . ${SCRIPTDIR}/utils.sh
 
--- a/.travis/runblack.sh
+++ b/.travis/runblack.sh
@@ -19,6 +19,9 @@
 #
 #   RUN_BLACK_DISABLED
 #     if set to an arbitrary non-empty value, black will be not executed
+#
+#   RUN_BLACK_EXTRA_ARGS
+#     extra cmd line args to pass to black
 
 set -e
 
@@ -61,4 +64,5 @@ set -x
 python -m black \
   --include "${INCLUDE_ARG:-${RUN_BLACK_INCLUDE:-${DEFAULT_INCLUDE}}}" \
   --exclude "${EXCLUDE_ARG:-${RUN_BLACK_EXCLUDE:-${DEFAULT_EXCLUDE}}}" \
+  ${RUN_BLACK_EXTRA_ARGS:-} \
   "${OTHER_ARGS[@]}"
--- a/.travis/runsyspycmd.sh
+++ b/.travis/runsyspycmd.sh
@@ -12,7 +12,6 @@ set -e
 
 ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
-TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
 
 . ${SCRIPTDIR}/utils.sh
 . ${SCRIPTDIR}/config.sh
--- a/.travis/utils.sh
+++ b/.travis/utils.sh
@@ -168,3 +168,13 @@ function lsr_venv_python_matches_system_python() {
 
   lsr_compare_pythons ${1:-python} -eq $syspython
 }
+
+# set TOPDIR
+ME=${ME:-$(basename $0)}
+SCRIPTDIR=${SCRIPTDIR:-$(readlink -f $(dirname $0))}
+TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
+
+# Local Variables:
+# mode: Shell-script
+# sh-basic-offset: 2
+# End:


.travis/runpylint.sh
```diff
@@ -17,6 +17,7 @@ set -e
 ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
 
+. ${SCRIPTDIR}/utils.sh
```
tyll, line 20: This change does not seem to make sense, since nothing is using utils.sh in this file.

.travis/runpytest.sh
```diff
@@ -14,9 +14,9 @@ set -e
 
 ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
-TOPDIR=$(readlink -f ${SCRIPTDIR}/..)
 
 . ${SCRIPTDIR}/utils.sh
+. ${SCRIPTDIR}/config.sh
```
tyll, line 19: The config does not seem to be user here, is it?

.travis/runflake8.sh
```diff
@@ -10,6 +10,8 @@
 #
 #   RUN_FLAKE8_DISABLED
 #     if set to an arbitrary non-empty value, flake8 will be not executed
+#   RUN_FLAKE8_IGNORE
```
tyll, line 13: The black wrapper uses a generic "EXTRA_ARGS" variable and this a specific "IGNORE" variable. It seems using "RUN_FLAKE8_EXTRA_ARGS" here would make the template more consistent and provide a similar benefit.

.travis/runpylint.sh
richm: I would have to check the python script custom_pylint.py to see if it is using any environment variables defined in utils.sh to be sure (I don't think so) - so this is primarily for the sake of consistency - every script in `.travis/` includes `utils.sh` so that in the future we can control the common behavior of all scripts in one central location, and every script can rely on having a common set of environment variables/functions/behaviors.  Similarly with `config.sh` below.  I suppose your argument might be "well, we don't need it now, why not just add it in the future when we do need it?" and I suppose, yes, that seems like a reasonable argument, but for the sake of consistency, why not just do it now?

.travis/runpytest.sh
richm: Similar to the above, this is primarily for the sake of consistency.

.travis/runflake8.sh
richm: do you mean there should be simply `RUN_FLAKE8_EXTRA_ARGS` which would allow users to provide `--ignore value,list` to runflake8.sh, and get rid of `RUN_FLAKE8_IGNORE`?

richm: Or, add a `RUN_FLAKE8_EXTRA_ARGS` in addition to `RUN_FLAKE8_IGNORE`, similar to how black has special purpose `RUN_BLACK_INCLUDE` and `RUN_BLACK_EXCLUDE` (because they are more commonly used and cover the vast majority of the cases where the user wants to customize the behavior?) and `RUN_BLACK_EXTRA_ARGS` for everything else?

.travis/runpylint.sh
tyll: IMHO it is bad style to have unnecessary code because it makes it harder to understand the code (in this case someone would need to check if utils.sh has an effect on runpylist.sh or not), makes it harder to change utils.sh (now more files needs to be checked if they might be broken by the changes) and it has a (probably low) performance impact. Static linters for other languages usually warn about unused imports, for example for Python. By not including unnecessary code, you also have consistency for a rule like everything includes only what it actually needs.

I will not block this PR if you have prefer to add this nonetheless.

richm: ok, now I remember - the function `lsr_setup_module_utils` is needed in order to run `pylint` in some repos like `storage` - so `runpylint.sh` must source both `utils.sh` in order to pick up the definition of the shell function `lsr_setup_module_utils`, and `config.sh` in order to run the shell function.  https://github.com/linux-system-roles/storage/blob/master/.travis/config.sh#L50

I guess this isn't a problem for network because you have your own way of solving this problem.

.travis/runpytest.sh
richm: similar to above with pylint -  the function `lsr_setup_module_utils` is needed in order to run pytest - so runpytest.sh must source both utils.sh in order to pick up the definition of the shell function lsr_setup_module_utils, and config.sh in order to run the shell function

.travis/runflake8.sh
richm: I'm assuming it is the latter

tyll: I guess I would also remvoe `RUN_BLACK_IN/EXCLUDE` and only allow enabling/disabling and extra args. But it seems the challenge with black is that it needs non-empty default values for include/exclude, therefore the specific variables are needed.

.travis/runpylint.sh
tyll: I see. Is this good practice is shell programming? It seems bad to me that sourcing a file called `config.sh` is intended to have side effects by running code from a different file that is not even sourced in `config.sh`. IMHO cleaner approaches would be to

a)
set a flag in config.sh that indicates that the `lsr_setup_module_utils` function should be called and explicitly call where needed

or

b) 
make `config.sh` source `utils.sh` to get the `lsr_setup_module_utils` function (this would at least make it a little bit clearer that it is running code from `utils.sh`
(This does not solve the side effect of sourcing).

What do you think?

.travis/runflake8.sh
richm: > needs non-empty default values for include/exclude

needs in what sense?  That they are required?  They don't appear to be mandatory arguments:
```
Usage: black [OPTIONS] [SRC]...

  The uncompromising code formatter.

Options:
...
  --include TEXT                  A regular expression that matches files and
                                  directories that should be included on
                                  recursive searches.  An empty value means
                                  all files are included regardless of the
                                  name.  Use forward slashes for directories
                                  on all platforms (Windows, too).  Exclusions
                                  are calculated first, inclusions later.
                                  [default: \.pyi?$]

  --exclude TEXT                  A regular expression that matches files and
                                  directories that should be excluded on
                                  recursive searches.  An empty value means no
                                  paths are excluded. Use forward slashes for
                                  directories on all platforms (Windows, too).
                                  Exclusions are calculated first, inclusions
                                  later.  [default: /(\.eggs|\.git|\.hg|\.mypy
                                  _cache|\.nox|\.tox|\.venv|\.svn|_build|buck-
                                  out|build|dist)/]
```

.travis/runpylint.sh
richm: with option b) - each script in `.travis/` would have the following:
```
. ${SCRIPTDIR}/utils.sh # OPTIONAL - only if the script requires the shell functions/env. vars.
. ${SCRIPTDIR}/config.sh # OPTIONAL - only if the script needs to allow the project to override behavior
```
so it may be that a script in `.travis/` will have both
```
. ${SCRIPTDIR}/utils.sh
. ${SCRIPTDIR}/config.sh
```
AND the project needs to override behavior in the script by calling a shell function defined in utils.sh:
```
${SCRIPTDIR}/config.sh:
. ${SCRIPTDIR}/utils.sh
lsr_setup_module_utils
```
So utils.sh will be sourced twice.  If this is what you intended, then this can work - I'll have to add "protection" to utils.sh so that it can either be sourced multiple times, or changed so that sourcing it a second or subsequent time will be a no-op.

.travis/runflake8.sh
tyll: The template uses non-empty default values:

```
DEFAULT_INCLUDE='^[^.].*\.py$'
DEFAULT_EXCLUDE='/(\.[^.].*|tests/roles)/'
```
From the value of the exclude I assume this is because `/tests/roles` might contain a symlink back to `/` causing an infinite loop.

.travis/runpylint.sh
tyll: Actually (a) is my favorite, (b) is a compromise because it is still a little bit odd that the config contains code/has side effects. But, yes, a guarding mechanism would be needed.

.travis/runflake8.sh
richm: So they weren't really mandatory, but they were convenient, or were otherwise useful to have as separate environment variables, since they could have been implemented using `RUN_BLACK_EXTRA_ARGS`

So for the purpose of setting a good coding style precedent for this and future code (as opposed to being consistent with runblack.sh), I should get rid of `RUN_FLAKE8_IGNORE` and instead allow this and other overrides to be done solely via `RUN_FLAKE8_EXTRA_ARGS`?

tyll: > So they weren't really mandatory, but they were convenient, or were otherwise useful to have as separate environment variables, since they could have been implemented using `RUN_BLACK_EXTRA_ARGS`

To implement it correctly, the shell script would need to parse `RUN_BLACK_EXTRA_ARGS` to check if they include `--include ...`. Because if it is missing, the script would need to add `--include "^[^.].*\.py$+ä+*"` to make it the default. For `--ignore` for flake8, the default would be not to include a a command-line flag since the defaults are in tox.ini.

> So for the purpose of setting a good coding style precedent for this and future code (as opposed to being consistent with runblack.sh), I should get rid of `RUN_FLAKE8_IGNORE` and instead allow this and other overrides to be done solely via `RUN_FLAKE8_EXTRA_ARGS`?

This is my suggestion. At least if there is a role that needs it.

tyll:
[citest bad]

PR #185, opened 2020-04-01
support testing/linting module_utils code

richm:
This does not directly affect network because network has its own
solution for dealing with testing/linting module_utils code, but
it gets network to feature/code parity with the template and
other repos.

Defines a shell function in utils.sh - lsr_setup_module_utils -
that allows configuring a tox venv installed ansible to have
the repo module_utils code in the default pythonpath for the
venv, which allows testing and linting the module_utils code
without having to otherwise munge or mock the pythonpath.

Projects needing this functionality can set in `.travis/config.sh`
the variable `RUN_PYTEST_SETUP_MODULE_UTILS` to setup
module_utils/ for pytest, and `RUN_PYLINT_SETUP_MODULE_UTILS`
to setup module_utils/ for pylint.

### DIFF
--- a/.travis/runpylint.sh
+++ b/.travis/runpylint.sh
@@ -12,12 +12,24 @@
 
 # The given command line arguments are passed to custom_pylint.py.
 
+# Environment variables:
+#
+#   RUN_PYLINT_SETUP_MODULE_UTILS
+#     if set to an arbitrary non-empty value, the environment will be
+#     configured so that linting of the module_utils/ code will be run
+#     correctly
+
 set -e
 
 ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
 
 . ${SCRIPTDIR}/config.sh
+
+if [[ "${RUN_PYLINT_SETUP_MODULE_UTILS}" ]]; then
+  . ${SCRIPTDIR}/utils.sh
+  lsr_setup_module_utils
+fi
 
 set -x
 python ${SCRIPTDIR}/custom_pylint.py "$@"
--- a/tests/setup_module_utils.sh
+++ b/tests/setup_module_utils.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+
+set -euo pipefail
+
+if [ -n "${DEBUG:-}" ] ; then
+    set -x
+fi
+
+if [ ! -d "${1:-}" ] ; then
+    echo Either ansible is not installed, or there is no ansible/module_utils
+    echo in $1 - Skipping
+    exit 0
+fi
+
+if [ ! -d "${2:-}" ] ; then
+    echo Role has no module_utils - Skipping
+    exit 0
+fi
+
+# we need absolute path for $2
+absmoddir=$( readlink -f "$2" )
+
+# clean up old links to module_utils
+for item in "$1"/* ; do
+    if lnitem=$( readlink "$item" ) && test -n "$lnitem" ; then
+        case "$lnitem" in
+            *"${2}"*) rm -f "$item" ;;
+        esac
+    fi
+done
+
+# add new links to module_utils
+for item in "$absmoddir"/* ; do
+    case "$item" in
+        *__pycache__) continue;;
+        *.pyc) continue;;
+    esac
+    bnitem=$( basename "$item" )
+    ln -s "$item" "$1/$bnitem"
+done
--- a/.travis/config.sh
+++ b/.travis/config.sh
@@ -20,6 +20,7 @@
 #       - RUN_PYLINT_INCLUDE
 #       - RUN_PYLINT_EXCLUDE
 #       - RUN_PYLINT_DISABLED
+#       - RUN_PYLINT_SETUP_MODULE_UTILS
 #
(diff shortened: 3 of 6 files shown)


tests/setup_module_utils.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
tyll, line 1: This script is missing license information

tests/setup_module_utils.sh
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: MIT
+
+set -euo pipefail
+
+if [ -n "${DEBUG:-}" ] ; then
+    set -x
+fi
+
+if [ ! -d "${1:-}" ] ; then
```
tyll, line 10: This is already checked in `utils.sh`, therefore this log message will not appear if the script is already skipped because of the check in `utils.sh`. This does not seem to be intentional.

tests/setup_module_utils.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo Either ansible is not installed, or there is no ansible/module_utils
+    echo in $1 - Skipping
+    exit 0
+fi
+
+if [ ! -d "${2:-}" ] ; then
+    echo Role has no module_utils - Skipping
```
tyll, line 17: Same problem as the check for `$1`.

.travis/runpylint.sh
```diff
@@ -12,12 +12,24 @@
 
 # The given command line arguments are passed to custom_pylint.py.
 
+# Environment variables:
+#
+#   RUN_PYLINT_SETUP_MODULE_UTILS
+#     if set to an arbitrary non-empty value, the environment will be
```
tyll, line 18: just a remark: Requesting a specific value such as `yes` makes it easier to allow other values in the future, too. For example the coveralls setting allows yes, strict and debug. Therefore it might make sense to decide on a specific value for boolean true here.

richm: The problem is that every other use of a boolean valued variable in the scripts assumes `-z "${VAR}"` is `false` and `-n "${VAR}"` or `[[ "${VAR}" ]]` is `true`.  If I make `RUN_PYLINT_SETUP_MODULE_UTILS` and `RUN_PYTEST_SETUP_MODULE_UTILS` have an actual values `no/yes` or `false/true` or something like that in this PR, that will be inconsistent with all other usage of boolean valued variables.
Note that `LSR_PUBLISH_COVERAGE` is not a boolean valued variable, and it doesn't allow `yes` as a value (or at least, once https://github.com/linux-system-roles/network/pull/186 is merged, `yes` will not be a valid value).

tyll: IIRC, "yes" was the proposed value for `LSR_PUBLISH_COVERAGE` in one of the other discussions. Just to clarify: If you prefer an arbitrary value as "true", I will not block this PR for this.

tests/setup_module_utils.sh
richm: I wanted to allow for this script being called separately in the future, in which case I wanted to leave the argument checking in.  It will also catch the case where someone makes an inadvertent breaking change to `tox.ini` where the environment variables are set.  So, I would prefer to leave this as-is.

tests/setup_module_utils.sh
richm: I wanted to allow for this script being called separately in the future, in which case I wanted to leave the argument checking in.  It will also catch the case where someone makes an inadvertent breaking change to `tox.ini` where the environment variables are set.  So, I would prefer to leave this as-is.

.travis/runpylint.sh
richm: > IIRC, "yes" was the proposed value for `LSR_PUBLISH_COVERAGE` in one of the other discussions.

It may have been, but https://github.com/linux-system-roles/network/pull/186 changes that - with that PR, the valid values are `strict`, `debug`, or `normal` (`normal` replaces `yes` as the value that means "run coverage normally").  And 186 maintains the convention in the test code that the empty value (`-z ${LSR_PUBLISH_COVERAGE}`) means `off` or `false` (as opposed to having a specific `off` or `false` value).

> Just to clarify: If you prefer an arbitrary value as "true", I will not block this PR for this.

I "prefer" it in the sense that, while I recognize that it is not ideal, I don't want to make a change which will be confusing and inconsistent, and I also don't want this PR to include a change to every other util script to change the behavior of every other boolean valued variable.

richm:
[citest bad]

tests/setup_module_utils.sh
tyll: I do not follow how this will catch anything if `tox.ini` is wrong because utils.sh contains this check:

```bash
  if [ -n "$srcdir" -a -d "$srcdir" -a -n "$destdir" -a -d "$destdir" ]; then
￼    bash $TOPDIR/tests/setup_module_utils.sh "$srcdir" "$destdir"
￼  fi
```

So of `$srcdir` is not a directory, this script will not be called and there will also be no error message. Nevertheless, I guess in this case the CI should fail anyhow since this is just a helper to get the CI running and without it it should fail for the roles that enable it.

PR #186, opened 2020-04-01
update to latest template code

richm:
These changes don't directly affect network, or don't actually
change the behavior of existing code that network uses (coveralls),
but are intended to sync with the latest template code.

Adds `LSR_PUBLISH_COVERAGE=normal` to `.travis/config.sh` because
network wants to publish coverage normally (as opposed to `strict`
or not at all).


tox.ini
```diff
@@ -168,13 +167,13 @@ commands = yamllint .
...
-whitelist_externals = bash
 deps =
     coveralls
 changedir = {[base]changedir}
+whitelist_externals =
+    {[base]whitelist_externals}
 commands =
-    bash -c 'cd ..; bash .travis/fix-coverage.sh; cd -'
-    coveralls
+    bash {toxinidir}/.travis/runcoveralls.sh {posargs}
```
tyll, line 176: Without changes to config.sh, this change will disable coveralls for the network role AFAIU. Even though it is currently broken, it should not be silently disabled.

.travis/runcoveralls.sh
```diff
@@ -0,0 +1,88 @@
...
+# Reports coverage results using coveralls. The aim of this script is to
+# provide a unified way to reporting coverage results across all linux system
+# roles projects.
+
+# The given command line arguments are passed to coveralls.
+
+# Environment variables:
+#
+#   LSR_PUBLISH_COVERAGE
+#     if the variable is empty or unset, nothing will be published; if the
```
tyll, line 13: This help is missing the info about what to do to enable coveralls in non-strict and non-debug mode. Currently it seems to do this for any value but it would be better to do this for a certain value or set of values only.

tox.ini
richm: I thought `.travis/runcoveralls.sh` was supposed to do the same thing that
```
cd ..; bash .travis/fix-coverage.sh; cd -
coveralls
```
does.  If not, then @i386x what was the purpose of `.travis/runcoveralls.sh`?  @i386x did you have some other purpose/functionality intended for `.travis/runcoveralls.sh`?

richm: > Without changes to config.sh,

What changes need to be made to `network/.travis/config.sh` to enable coveralls, and can I include them in this PR?

.travis/runcoveralls.sh
richm: @tyll I don't know anything about `runcoveralls.sh` - I'm just trying to sync what @i386x added to the template with other repos - at this point I would rather drop it as it seems it is not fully baked and doesn't really work.

tox.ini
tyll: This change is there to allow disabling it (which is the default), so it needs something like `LSR_PUBLISH_COVERAGE=true` or `LSR_PUBLISH_COVERAGE=yes` not sure if there is a specific style for booleans in the template wrappers.

tox.ini
```diff
@@ -168,13 +167,13 @@ commands = yamllint .
...
 envdir = {toxworkdir}/env-{env:TRAVIS_PYTHON_VERSION:coveralls}
 passenv = TRAVIS TRAVIS_*
-whitelist_externals = bash
 deps =
     coveralls
 changedir = {[base]changedir}
+whitelist_externals =
+    {[base]whitelist_externals}
 commands =
-    bash -c 'cd ..; bash .travis/fix-coverage.sh; cd -'
```
tyll, line 176: The file `.travis/fix-coverage.sh` should be removed as part of this PR.

richm:
[citest bad]

tyll:
There is a conflict now that needs to be resolved.

richm:
[citest bad]

PR #191, opened 2020-04-15
Integration tests: Fix coverage collection

tyll:
Update coverage helpers for changes in newer Ansible versions (tested
with 2.9.6) and with using module_utils.


pcahyna:
Would this be useful to have in the role template?

richm:
> Would this be useful to have in the role template?

Yes.

tyll:
> > Would this be useful to have in the role template?
> 
> Yes.

Currently, the files contain specific code for the network role, it is not documented and it lacks integration with the CI. IMHO it might make more sense to move this feature to its own project that the CI uses. If we only integrate it with the test-harness, it does not need any changes to the template, I guess.

PR #193, opened 2020-04-16
Library/NM: Always rollback checkpoint on failure

### LINKED ISSUE #154: Checkpoint is not reverted when role aborts with log_error()
`log_error()` does not only log but also exists the module. In this case, the role does not yet revert the checkpoint. This needs to be changed.

tyll:
If there is a NetworkManager checkpoint, roll it back on any failure.
Otherwise future invocations of the role fail until the checkpoint timed
out.

closes #154

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1474,6 +1474,7 @@ class RunEnvironmentAnsible(RunEnvironment):
         RunEnvironment.__init__(self)
         self._run_results = []
         self._log_idx = 0
+        self.on_failure = None
 
         from ansible.module_utils.basic import AnsibleModule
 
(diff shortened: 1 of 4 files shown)


tyll:
I tested this manually with:
`ansible-playbook -i rhel82-cloud, tests/tests_ethtool_features_initscripts.yml tests/tests_ethtool_features_nm.yml tests/tests_states.yml tests/tests_states.yml`

tyll:
@richm @pcahyna just for statistics, the CI is failing again for infrastructure problems, for example:

```
fatal: [/cache/Fedora-Cloud-Base-31-1.9.x86_64.qcow2]: FAILED! => {"changed": false, "msg": "Failed to download packages: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-31&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]", "results": []}
```

Issue #188, opened 2020-04-09
Support state:down persistent_state:absent for non-existent profile


tyll:
A configuration like

```yaml
network_connections:
   - name: non_existent
     state: down
     persistent_state: absent
```

should succeed instead of fail to allow idempotency/repeating this config to remove a profile (if it exists) without failing in later ansible runs.

Tasks:
- [x] Create an integration test
- [ ] Fix this for NM
- [ ] fix this for initscripts

Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1822777

a-mere-peasant:
is something like this okay for the integration test ? 
```
---
- name: testing state down persistent_state absent for non existent profile
  hosts: all
  vars:
    network_connections:
       - name: non_existent
         state: down
         persistent_state: absent

  roles:
    - linux-system-roles.network
```
would fixing this mean not generating errors or ignoring them like #70 states ?
I think there should at least be a warning if in case there is a typo

tyll:
@a-mere-peasant in general, it looks good. Using the `import_role` might be nicer to clarify that just calling the role it is the test. What do you think?

tyll:
The test should probably be added to `test_states.yml`.

a-mere-peasant:
Clarifying that calling the role is the test is sounds great.

 The tests in `tests_states.yml`use include_role instead of import_role ,so should I just use that for consistency ? 

Also should i include tasks for asserting the device and profile absent ?

tyll:
> Clarifying that calling the role is the test is sounds great.
> 
> The tests in `tests_states.yml`use include_role instead of import_role ,so should I just use that for consistency ?

yes, please.

> Also should i include tasks for asserting the device and profile absent ?

yes, sounds good. Thank you!

PR #200, opened 2020-04-20
Add a test for downing non existent profile

a-mere-peasant:


a-mere-peasant:
This test is added to address the issue #188

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,16 @@
             state: down
     - include_tasks: tasks/assert-device_absent.yml
     - include_tasks: tasks/assert-profile_absent.yml
+
+    #downing non_existent profile
```
tyll, line 51: Please write this as a user story as in the examples above.

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,16 @@
...
+    - name: Set down non_existent profile
```
tyll, line 52: Please run this by yourself and check if it works. The CI also complains about it (Travis).

a-mere-peasant: It seems there was a linting error ( an extra blank line at the end ). I had checked if the playbook was running as expected .

tyll: What does expected mean for you? Does it work or does it fail?

a-mere-peasant: It fails with this error message:
```
[WARNING]: [001] <error> #0, state:down persistent_state:absent,
'non_existent': Connection not found on system and 'type' not specified
fatal: [localhost]: FAILED! => {"changed": false, "msg": "error: Connection not found on system and 'type' not specified"}
```
which I think should happen as the connection does not exist and the role does not know that it is to be downed . Should something else happen ?

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,16 @@
...
+    # I can set down and remove a non existent profile
+    - name: Set down non_existent profile
+      include_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: non_existent
+            state: down
+            persistent_state: absent
+    - include_tasks: tasks/assert-device_absent.yml
```
tyll, line 61: These tasks are actually not working as intended here, since they use the variable `interface` and therefore do not check `non_existent` here.

tests/playbooks/tests_states.yml
tyll: This sounds correct. However, this means we cannot merge this as-is because it will break the CI. Therefore this also needs the fix for NM or some extra Ansible code that checks that this fails and ensures that the playbook will still succeed.

a-mere-peasant:
The playbook now ignores this task failing and succeeds, should I add a comment that this is only till the issue gets fixed ?or leave it as is ?

tyll:
> The playbook now ignores this task failing and succeeds, should I add a comment that this is only till the issue gets fixed ?or leave it as is ?

Good idea. Today I learned about [this pattern](https://github.com/linux-system-roles/storage/blob/master/tests/tests_disk_errors.yml#L17) to specify that a failure is expected, can you use it here, too? Thanks:

a-mere-peasant:
> > The playbook now ignores this task failing and succeeds, should I add a comment that this is only till the issue gets fixed ?or leave it as is ?
> 
> Good idea. Today I learned about [this pattern](https://github.com/linux-system-roles/storage/blob/master/tests/tests_disk_errors.yml#L17) to specify that a failure is expected, can you use it here, too? Thanks:

I was unable to apply a similar pattern ,there were some errors occuring while chekcing the `ansible_failed_task.name`,so I'm using this alternate way . Does this seem to be fine ?
```
    - name: Verify that the play fails with the expected error message
      block:
        - name: Set down non existent profile
          include_role:
          vars:
            network_connections:
              - name: non_existent
                type: ethernet
                state: down
                persistent_state: absent

        - name: 'UNREACH'
          fail:
            msg: 'Downing a non_existent profile did not fail, discard next message'

      rescue:
        - debug:
            msg: 'Downing a non existent profile failed'
```

richm:
> I was unable to apply a similar pattern ,there were some errors occuring while chekcing the ansible_failed_task.name,so I'm using this alternate way . Does this seem to be fine ?

You didn't quite apply the similar pattern correctly.  Please compare what you did to https://github.com/linux-system-roles/storage/blob/master/tests/tests_disk_errors.yml#L33-L41  HINT: you may have to define more vars than just `network_provider_current`

richm:
See also https://lists.fedorahosted.org/archives/list/systemroles@lists.fedorahosted.org/message/3C5L2CGJVXNYYBJ7GMKT47NAVW3QCR4F/ for a potentially better way to do this.

a-mere-peasant:
> > The playbook now ignores this task failing and succeeds, should I add a comment that this is only till the issue gets fixed ?or leave it as is ?
> 
> Good idea. Today I learned about [this pattern](https://github.com/linux-system-roles/storage/blob/master/tests/tests_disk_errors.yml#L17) to specify that a failure is expected, can you use it here, too? Thanks:

This method suffers from this issue https://github.com/ansible/ansible/issues/49942 and the corresponding test would look something like this 
```
    - name: Verify that the play fails with the expected error message
      block:
        - name: Set down non existent profile
          include_role:
          vars:
            network_connections:
              - name: non_existent
                type: ethernet
                state: down
                persistent_state: absent

        - name: 'UNREACH'
          fail:
            msg: 'This should be unreachable'

      rescue:
        - name: Check that we failed in the role
          assert:
            that:
              - ansible_failed_task.name != 'UNREACH'
            msg: "Role has not failed when it should have"
          vars:
            network_provider_current: "{{
                'nm' if 'NetworkManager.service' in ansible_facts.services and
                    ansible_facts.services['NetworkManager.service']['state'] == 'running'
                    else 'initscripts'
                }}"
            network_connections:
              - name: non_existent
                type: ethernet
                state: down
                persistent_state: absent
            wpa_supplicant_required: "{{ network_connections |
                 selectattr('ieee802_1x', 'defined') | list | count > 0 }}"
            _network_packages_default_gobject_packages: ["python{{
                  ansible_python['version']['major'] | replace('2', '')}}-gobject-base"]
            _network_packages_default_802_1x: ["{% if wpa_supplicant_required
                  %}wpa_supplicant{% endif %}"]
            _network_packages_default_initscripts_bridge: ["{%
            if ['bridge'] in network_connections|json_query('[*][type]') and
               ansible_distribution in ['RedHat', 'CentOS', 'OracleLinux'] and
               ansible_distribution_major_version is version('7', '<=')
            %}bridge-utils{% endif %}"]
            _network_packages_default_initscripts_network_scripts: ["{%
            if ansible_distribution in ['RedHat', 'CentOS', 'OracleLinux'] and
                ansible_distribution_major_version is version('7', '<=')
             %}initscripts{% else %}network-scripts{% endif %}"]
            network_service_name_default_nm: NetworkManager
            network_packages_default_nm: "{{ ['ethtool', 'NetworkManager']
                  + _network_packages_default_gobject_packages|select()|list()
                  + _network_packages_default_802_1x|select()|list()}}"
            network_service_name_default_initscripts: network
            network_packages_default_initscripts: "{{ ['ethtool']
            + _network_packages_default_initscripts_bridge|select()|list()
            + _network_packages_default_initscripts_network_scripts|select()|list()
            }}"
            _network_provider_setup:
              nm:
                service_name: "{{ network_service_name_default_nm }}"
                packages: "{{ network_packages_default_nm }}"
              initscripts:
                service_name: "{{ network_service_name_default_initscripts }}"
                packages: "{{ network_packages_default_initscripts }}"

            network_packages: "{{
                _network_provider_setup[network_provider]['packages'] }}"
            network_service_name: "{{
                _network_provider_setup[network_provider]['service_name'] }}"
            network_provider: "{{ network_provider_current }}"
```
@richm suggested a better way to do this and also the issue with using `ansible_failed_test`  [here](https://lists.fedorahosted.org/archives/list/systemroles@lists.fedorahosted.org/message/MKW25OX4RKM6HCCRYDVIPQIXWSE2NNGH/)  (Thanks a lot for the help)and I will be using the same for this test .

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,32 @@
...
+                state: down
+                persistent_state: absent
+
+        - name: 'UNREACH'
+          set_fact:
+            __network_test_failed: false
+
+      rescue:
+        - debug:
+            msg: 'The role has failed'
```
richm, line 72: I don't think you need the `debug` here anymore.  Do you need the `rescue` block at all, or is the `rescue` block that resets the failed task state for Ansible to continue with subsequent tasks?  Try it and see if it works without the `rescue` block.  If you do need the `rescue` block, then perhaps something like this:
```
       rescue:
         set_fact:
           __network_test_failed: true
```

a-mere-peasant: > I don't think you need the `debug` here anymore. Do you need the `rescue` block at all, or is the `rescue` block that resets the failed task state for Ansible to continue with subsequent tasks? 

The `rescue` block is needed so that the failed task is reset and the playbook continues as usual since we need the playbook to succeed even when this task fails (which it will until #188 is fixed) . 

> Try it and see if it works without the `rescue` block. If you do need the `rescue` block, then perhaps something like this:
>
> ```
>        rescue:
>          set_fact:
>            __network_test_failed: true
> ```
Wouldn't setting the rescue block as this mean that the `always` block will succeed and always show  ` success_msg: "The role failed when it should have" ` and we wouldn't know whether the role failed on the correct task ?

richm: > > I don't think you need the `debug` here anymore. Do you need the `rescue` block at all, or is the `rescue` block that resets the failed task state for Ansible to continue with subsequent tasks?
> 
> The `rescue` block is needed so that the failed task is reset and the playbook continues as usual since we need the playbook to succeed even when this task fails (which it will until #188 is fixed) .

Ok.

> 
> > Try it and see if it works without the `rescue` block. If you do need the `rescue` block, then perhaps something like this:
> > ```
> >        rescue:
> >          set_fact:
> >            __network_test_failed: true
> > ```
> 
> Wouldn't setting the rescue block as this mean that the `always` block will succeed and always show `success_msg: "The role failed when it should have"` and we wouldn't know whether the role failed on the correct task ?

I'm not sure what you mean.  And perhaps my terminology/variable naming is confusing.  Maybe something like this:
```yaml
    - name: Verify that the play fails with the expected error message
      block:
        - name: Set down non existent profile
          include_role:
          vars:
            network_connections:
              - name: non_existent
                type: ethernet
                state: down
                persistent_state: absent

        - name: This task should never execute
          set_fact:
            __network_role_failed: false

      rescue:
        - name: Role correctly reported failure
          set_fact:
            __network_role_failed: true

      always:
        - assert:
            that:  __network_role_failed
```
That is, what this `block/rescue/always` block is testing is that the role fails as expected.   And, you can test this, by making the `include_role` task succeed.

> we wouldn't know whether the role failed on the correct task ?

It depends on what you mean.  If by "correct task" you mean some specific task inside the network role, then we have no way to know that since `ansible_failed_task` is not working.  If by "correct task" you mean the `include_role` task in the `block` then yes, we know that this task failed because it is the only task in the `block`, so it is the only task which will go into the `rescue` block upon failure.

If this is not acceptable - if we really need to know exactly which network role task failed, then you'll have to implement something using `ansible_failed_test` that checks the error message and other context available.

a-mere-peasant: >It depends on what you mean. If by "correct task" you mean some specific task inside the network role, then we have no way to know that since ansible_failed_task is not working. If by "correct task" you mean the include_role task in the block then yes, we know that this task failed because it is the only task in the block, so it is the only task which will go into the rescue block upon failure.

My bad , I was still thinking about the previous configuration with the `UNREACH` task which was to fail if the `include_role` task did not . I'll change this accordingly .

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,32 @@
...
+
+      rescue:
+        - name: Role has failed when it should have
+          set_fact:
+            __network_test_failed: true
+
+      always:
+        - assert:
+            that: __network_test_failed
+            fail_msg: "The role did not fail when it shoud have"
```
tyll, line 78:
```suggestion
            fail_msg: "The role did not fail when it should have"
```

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,32 @@
             state: down
     - include_tasks: tasks/assert_device_absent.yml
     - include_tasks: tasks/assert_profile_absent.yml
+
+    #test case (downing a non existent profile)
+    #I can set down and remove a non existent profile
+    #This will be rescued until issue 188 is fixed
```
tyll, line 53:
```suggestion
    # test case (downing a non existent profile)
    # I can set down and remove a non existent profile
    # This will be rescued until issue #188 is fixed
```

tyll:
[citest commit:50ef8ddfa2c2446fe80c02750ed0910ca9ef81be]

tyll:
@a-mere-peasant please rebase on master

tests/playbooks/tests_states.yml
```diff
@@ -47,3 +47,32 @@
...
+    # This will be rescued until issue 188 is fixed
+    - name: Verify that the play fails with the expected error message
+      block:
+        - name: Set down non existent profile
+          include_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: non_existent
+                type: ethernet
```
tyll, line 62: Does this work without the type?

tyll:
[citest commit:3afbe81b9617f313db173f8666a17e09a0dfc2fc]

tests/playbooks/tests_states.yml
a-mere-peasant: Yes.
Really sorry I forgot to remove this , I was using this to test rescue block in the beginning and I didn't change it when I changed the rest . I'll fix this right away .

tyll:
[citest commit:2b7e16a839ff58771fe6632940b5020d9bfe070d]

pcahyna:
@richm @a-mere-peasant 

> > Good idea. Today I learned about [this pattern](https://github.com/linux-system-roles/storage/blob/master/tests/tests_disk_errors.yml#L17) to specify that a failure is expected, can you use it here, too? Thanks:
> 
> This method suffers from this issue [ansible/ansible#49942](https://github.com/ansible/ansible/issues/49942)

I determined a similar method which avoids that issue and the ugly workaround that you would have to make otherwise to keep `ansible_failed_task.name` from causing you problems, look at https://github.com/linux-system-roles/storage/pull/88 .

Setting a fact works too, but I preferred to avoid it, because a fact can not be unset and pollutes the namespace, which could sometimes bite you - a fact is essentially a global variable.

Issue #198, opened 2020-04-19
Inconsistent filenames in tests/tasks


jaedolph:
Most filenames in the role seem to use underscores except those in `tests/tasks` e.g. `assert-device_absent.yml`. It would be more consistent to change these filenames to use underscores rather than hyphens. This would also make the filenames valid python identifiers as discussed here: #189

PR #203, opened 2020-04-21
Rename files to be valid python identifiers

jaedolph:
As identified here:  #198


tests/get_total_coverage.sh
```diff
@@ -20,13 +20,13 @@ tox -e py26,py27,py36,py37 -- --cov-append
```
jaedolph, line 20: Couldn't find this file referenced anywhere. Do you know where it gets called from?

tyll: This is only called manually. It needs to be integrated into a CI or test-harness needs to be changed to call `get_coverage.sh` or something else to create coverage data. But nothing happened in this direction, yet.

jaedolph: Ah right, should be ok to change the name then? I have changed all the references to the other coverage scripts it references (e.g. `get_coverage.sh`)

jaedolph:
[citest bad]

tests/get_total_coverage.sh
tyll: Probably not needed but it does not hurt, too.

jaedolph:
[citest bad]

tyll:
FYI: This has a conflict that needs to be resolved

jaedolph:
> FYI: This has a conflict that needs to be resolved

Fixed now

jaedolph:
[citest bad]

Issue #29, opened 2018-03-07
[RFE] refactor logging of network module


thom311:
Currently, the module collects all logging statements, and at the end returns them as "warnings", so that they are shown by ansible. Obviously, these are not really warnings, but rather debug information.

Instead, the logging messages should be returned in a different json field that is ignored by ansible. Then, the `tasks/main.yml` should have a follow-up `debug` task that prints the returned variable.

I guess, in the failure case, the `network_connections` task must run ignoring failures to reach the `debug` statement. Then, a follow up task should check whether the `network_connections` task failed and abort.

tyll:
Ansible now supports blocks and the messages could be shown in the rescue or always part in case of failures. Reference: https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html

tyll:
Actually in case of a failure, the module can just output the logs as warnings as currently in addition to putting them into a different json field. No need for a rescue block IMHO.  The debug statement might be nice nonetheless.

PR #207, opened 2020-04-26
library: Separate debug and info logs from warnings

elvgarrui:
Logs are now separed by severity level. Warnings and Failures are treated like
before, but debug and info logs are saved into a new json parameter called
"info_debug" that is later shown by a new created task.

In order to achieve a better readability, debug and info logs are shown as
warnings in case of failure like before, and also in "info_debug". If there is
not any failure, debug and info logs are never shown as warnings.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1559,24 +1559,32 @@ class RunEnvironmentAnsible(RunEnvironment):
                 c["persistent_state"],
             )
             prefix = prefix + (", '%s'" % (c["name"]))
-        for r in rr["log"]:
-            yield (r[2], "[%03d] %s %s: %s" % (r[2], LogLevel.fmt(r[0]), prefix, r[1]))
+        for severity, msg, idx in rr["log"]:
+            yield (
+                idx,
+                "[%03d] %s %s: %s" % (idx, LogLevel.fmt(severity), prefix, msg),
+                severity,
+            )
 
-    def _complete_kwargs(self, connections, kwargs, traceback_msg=None):
-        if "warnings" in kwargs:
-            logs = list(kwargs["warnings"])
-        else:
-            logs = []
-
+    def _complete_kwargs(self, connections, kwargs, traceback_msg=None, fail=False):
+        warning_logs = kwargs.get("warnings", [])
+        debug_logs = []
         loglines = []
         for res in self._run_results:
             for idx, rr in enumerate(res):
                 loglines.extend(self._complete_kwargs_loglines(rr, connections, idx))
-        loglines.sort(key=lambda x: x[0])
-        logs.extend([x[1] for x in loglines])
+        loglines.sort(key=lambda log_line: log_line[0])
+        for idx, log_line, severity in loglines:
+            debug_logs.append(log_line)
+            if fail:
+                warning_logs.append(log_line)
+            elif severity >= LogLevel.WARN:
+                warning_logs.append(log_line)
         if traceback_msg is not None:
-            logs.append(traceback_msg)
-        kwargs["warnings"] = logs
+            warning_logs.append(traceback_msg)
+        kwargs["warnings"] = warning_logs
+        stderr = "\n".join(debug_logs) + "\n"
+        kwargs["stderr"] = stderr
         return kwargs
 
     def exit_json(self, connections, changed=False, **kwargs):
@@ -1595,7 +1603,7 @@ class RunEnvironmentAnsible(RunEnvironment):
         kwargs["msg"] = msg
         kwargs["changed"] = changed
         self.module.fail_json(
-            **self._complete_kwargs(connections, kwargs, traceback_msg)
+            **self._complete_kwargs(connections, kwargs, traceback_msg, fail=True)
         )
 
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -64,6 +64,10 @@
     ignore_errors: "{{ network_ignore_errors | default(omit) }}"
     force_state_change: "{{ network_force_state_change | default(omit) }}"
     connections: "{{ network_connections | default([]) }}"
+  register: __network_connections_result
+
+- name: Show debug messages
+  debug: var=__network_connections_result
 
 - name: Re-test connectivity
   ping:
--- a/tests/playbooks/tests_ethernet.yml
+++ b/tests/playbooks/tests_ethernet.yml
@@ -37,6 +37,8 @@
           address: 192.0.2.1/24
   roles:
     - linux-system-roles.network
+  tasks:
+    - include_tasks: tasks/assert_output_in_stderr_without_warnings.yml
 
 - hosts: all
   tasks:
--- a/tests/playbooks/tests_reapply.yml
+++ b/tests/playbooks/tests_reapply.yml
@@ -50,7 +50,7 @@
         - name: Assert that reapply is found in log output
           assert:
             fail_msg: Reapply not found in log output
-            that: "{{ 'connection reapplied' in test_module_run.warnings[2] }}"
+            that: "{{ 'connection reapplied' in test_module_run.stderr }}"
       always:
         - block:
             # Use internal module directly for speedup
--- a/tests/tasks/assert_output_in_stderr_without_warnings.yml
+++ b/tests/tasks/assert_output_in_stderr_without_warnings.yml
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: "Assert that warnings is empty"
+  assert:
+    that:
+      - "'warnings' not in __network_connections_result"
+    msg: "There are unexpected warnings"
+- name: "Assert that there is output in stderr"
+  assert:
+    that:
+      - "'stderr' in __network_connections_result"
+    msg: "There are no messages in stderr"
--- a/tests/tests_default.yml
+++ b/tests/tests_default.yml
@@ -4,3 +4,9 @@
   hosts: all
   roles:
     - linux-system-roles.network
+  tasks:
+    - name: Test warning and info logs
+      assert:
+        that:
+          - "'warnings' not in __network_connections_result"
+        msg: "There are warnings"


elvgarrui:
This PR refers to issue #29

library/network_connections.py
```diff
@@ -1589,7 +1600,7 @@ def fail_json(
         kwargs["msg"] = msg
         kwargs["changed"] = changed
         self.module.fail_json(
-            **self._complete_kwargs(connections, kwargs, traceback_msg)
+            **self._complete_kwargs(connections, kwargs, traceback_msg, True)
```
tyll, line 1603: To make it more clear what `True` refers to:

```suggestion
            **self._complete_kwargs(connections, kwargs, traceback_msg, fail=True)
```

library/network_connections.py
```diff
@@ -1554,23 +1554,34 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
             )
             prefix = prefix + (", '%s'" % (c["name"]))
         for r in rr["log"]:
-            yield (r[2], "[%03d] %s %s: %s" % (r[2], LogLevel.fmt(r[0]), prefix, r[1]))
+            yield (r[2], "[%03d] %s %s: %s" % (r[2], LogLevel.fmt(r[0]), prefix, r[1]), r[0])
```
tyll, line 1557: This fails in the CI because the line is too long. Please re-format it with the `black` tool.

library/network_connections.py
```diff
@@ -1554,23 +1554,34 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
             )
             prefix = prefix + (", '%s'" % (c["name"]))
         for r in rr["log"]:
```
tyll, line 1556: It would be great if you could change this to something like

`for loglevel, b, c in rr["log"]` with good names for loglevel, b, c. Then these names could be used instead of `r[0]` etc in the next line to make it easier to understand. Hope you still know what these fields are.

library/network_connections.py
```diff
@@ -1554,23 +1554,34 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
...
+                    debug_logs.append(x[1])
+            else:
+                if x[2] in (LogLevel.DEBUG, LogLevel.INFO):
+                    debug_logs.append(x[1])
+                else:
+                    warning_logs.append(x[1])
         if traceback_msg is not None:
-            logs.append(traceback_msg)
-        kwargs["warnings"] = logs
+            warning_logs.append(traceback_msg)
```
tyll, line 1582: The initial issue was not so clear. I would prefer to:

- Return logs with warning or worse as warnings
- In case the module fails, return all logs as warnings
- Also return all logs (including warnings) in a new field. I was taking a look at [Ansibles commont return values](https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#common-return-values) and I guess we could use `stderr` for log messages. @pcahyna what do you think about this?

@elvgarrui what's your opinion?

tasks/main.yml
```diff
@@ -64,6 +64,10 @@
     ignore_errors: "{{ network_ignore_errors | default(omit) }}"
     force_state_change: "{{ network_force_state_change | default(omit) }}"
     connections: "{{ network_connections | default([]) }}"
+  register: result
```
tyll, line 67:
```suggestion
  register: network_connections_result
```

library/network_connections.py
elvgarrui: As far as I have seen this clarification makes sense, thanks!! Using standard names usually is the better option.

library/network_connections.py
elvgarrui: I've change this for a better readability, thanks for the tip!

library/network_connections.py
elvgarrui: Done!

library/network_connections.py
elvgarrui: I've fixed it now

tasks/main.yml
elvgarrui: Done, thanks!

library/network_connections.py
elvgarrui: I tried using `stderr` as new field but it seems like is not compatible. Instead, I could use `stderr_lines`, which is in the documentation said to keep the logs on a string format (as they are now). 

I'm not sure if this is the better option since the documentation says that the purpose of `stderr_lines` is, citing them: "When `stderr` is returned we also always provide this field which is a list of strings, one item per line from the original." - and there is no `stderr` field now, AFAIK.

The error that was returned when trying to assign the logs to stderr was: `AttributeError: 'list' object has no attribute 'splitlines'`

tyll: `stderr` needs to be a string with the log lines separated by newlines, so `stderr = "\n".join(logs) + "\n"` should probably work.

library/network_connections.py
```diff
@@ -1553,24 +1553,35 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
...
+            warning_logs = []
 
+        debug_logs = []
         loglines = []
         for res in self._run_results:
             for idx, rr in enumerate(res):
                 loglines.extend(self._complete_kwargs_loglines(rr, connections, idx))
         loglines.sort(key=lambda x: x[0])
-        logs.extend([x[1] for x in loglines])
+        for x in loglines:
```
tyll, line 1575: Please use names for `x[1]` etc here, too

library/network_connections.py
```diff
@@ -1553,24 +1553,35 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
...
+            debug_logs.append(x[1])
+            if fail:
+                warning_logs.append(x[1])
+            elif x[2] not in (LogLevel.DEBUG, LogLevel.INFO):
```
tyll, line 1579:
```suggestion
            elif x[2] < LogLevel.WARNING:
```
This allows to add other intermediate loglevels in the future since all log levels are numeric:
https://docs.python.org/3/library/logging.html#logging-levels

elvgarrui: It was >= instead of <, because we want to save as warning those logs with equal or higher priority than `WARN`. I've fixed it now!

library/network_connections.py
elvgarrui: Fixed, thanks!

library/network_connections.py
elvgarrui: It works now, thanks!

library/network_connections.py
tyll: Good catch :-)

tyll:
Awesome, looks very good, thank you. I will check it out tomorrow. It would be great if you could also add some integration tests.

Ideas:

- for example in https://github.com/linux-system-roles/network/blob/master/tests/tests_default.yml, an assert for network_connections_result.warnings to be empty (not sure if this would have any output in stderr)
- and in https://github.com/linux-system-roles/network/blob/master/tests/playbooks/tests_ethernet.yml to assert that are are no warnings and that there is some output in stderr.
- Not sure if there is a good test case to trigger a warning and to check that - this might be something that could be better tested once we have better capabilities to test with pytest.

library/network_connections.py
pcahyna: @tyll `stderr` is described as "Some modules execute command line utilities or are geared for executing commands directly (raw, shell, command, etc), this field contains the error output of these utilities.", I think you are abusing it, because this does not seem to be the case here.
Why do you need to return logs back actually? Is there a prior art for this in other Ansible modules? Who will read them in an automated setting? (Concerning warnings, see  https://blog.schlomo.schapiro.org/2015/04/warning-is-waste-of-my-time.html .) Can't you just send them to syslog on the managed host?
But if you really need to return them, I think you can just add another field to the result dict, instead of abusing stderr. Here is an example of a module that adds plenty of such fields: https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html

tyll: > @tyll `stderr` is described as "Some modules execute command line utilities or are geared for executing commands directly (raw, shell, command, etc), this field contains the error output of these utilities.", I think you are abusing it, because this does not seem to be the case here.

Do you see a specific problem with using this here? It does not conflict with any future development and stderr is meant to status output. To me it sounds like it should be intuitively clear for users what it means.

> Why do you need to return logs back actually? Is there a prior art for this in other Ansible modules? Who will read them in an automated setting? (Concerning warnings, see https://blog.schlomo.schapiro.org/2015/04/warning-is-waste-of-my-time.html .) Can't you just send them to syslog on the managed host?

These are valid questions. Personally, I found the info useful but that was not as a regular user. Nevertheless, this seems to be out of scope for the current issue at hand. I guess we can track/consider further enhancements. It seems that they could also be added to the common standard.

1) Consider removing the amount of log levels (this needs further discussion): I am not sure that I fully agree with the blog. IMHO warnings have also a place to tell about stuff that will become a problem/error in the future. Also, there might be things where we do not have a common agreement about how the tool should behave (like restarting NM)

2) Log to syslog, too. Sounds fine with me but can be added in a later step IMHO. Are there any recommendations/best practices to do this as part of Ansible?

> But if you really need to return them, I think you can just add another field to the result dict, instead of abusing stderr. Here is an example of a module that adds plenty of such fields: https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html

What is the advantage of doing this?

tyll:
@elvgarrui could you please take a look into tests for this as I wrote in my previous comment?

library/network_connections.py
tyll: @pcahyna please take a look at my questions. I am fine with merging this once there are tests and am open to follow-up with further improvements if we can agree on them. Especialy syslog sounds fine just want to know more about it.

elvgarrui:
@tyll I've made modifications in the tests in order to achieve both points. It seems to work, but I'm not sure if I should have written the task inside of the `test_default.yml` or in a separate file inside `/tasks`. In fact, I must say I'm not sure about what is the difference between the tests in the main folder and the ones in the `/playbook` folder.

tyll:
> @tyll I've made modifications in the tests in order to achieve both points. It seems to work, but I'm not sure if I should have written the task inside of the `test_default.yml` or in a separate file inside `/tasks`. In fact, I must say I'm not sure about what is the difference between the tests in the main folder and the ones in the `/playbook` folder.

Thanks, the idea is this:

`./tests/playbooks` contains the actual test files
`./tests/tests_*.yml` contains shims to run the test once for nm and once for initscripts or only nm/initscripts if it would not work for both providers
`tests_default.yml` is special because it is the one test that runs with the default provider for the platform without explicitly setting it and it is also excpected to exists by the molecule testing setup that runs as part of the tox checks.
`./tests/tasks` contains task snippets that are used in multiple tests to avoid having the same code in multiple places

What else is unclear? Happy to explain.

tyll:
Just realized, that there is more content in `tests/playbooks`. It contains helper playbooks as well, since for example in `tests/playbooks/tests_ethernet.yml`, the test steps consist of different playbooks (for example, this is what the run_tasks.yml playbook is for). This is because at the time the test was written, Ansible did not support importing roles as a task. This is now what is done in newer tests: https://github.com/linux-system-roles/network/blob/master/tests/playbooks/tests_ethtool_features.yml

tyll:
There is a yamllint error in one of the files, please see the CI output (or try it with `tox -e yamllint`).

tests/playbooks/tests_ethernet.yml
```diff
@@ -60,3 +60,8 @@
       vars:
         state: absent
     - include_tasks: tasks/assert_device_absent.yml
+
+- name: Test warning and info logs
+  hosts: all
+  tasks:
+    - include_tasks: tasks/assert_output_logs.yml
```
tyll, line 67: Please add this to the "Test static interface up" playbook to ensure that the output from there is tested.

tests/tasks/assert_output_logs.yml
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
tyll, line 1: Maybe give this file a better name that describes the intended outcome like `assert_output_in_stderr_without_warnings.yml`- what do you think?

elvgarrui: I tried to keep it short and clear, but this name is much more accurate, thanks!

tests/playbooks/tests_ethernet.yml
elvgarrui: Ok, thanks! This was one of the things that I was not sure I was making right. In a future test implementation, how should I decide where to add the new task?

elvgarrui:
@tyll I've fixed the issues you commented.
Thanks for your previous explanation! It has make thing clearer for me.

Furthermore, I do have one last question:
Is it recommended to run every integration test before submitting new code (I guess it is recommended) and if so, is there any command or shortcut to run all playbooks?

library/network_connections.py
```diff
@@ -1559,24 +1559,36 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
...
+                idx,
+                "[%03d] %s %s: %s" % (idx, LogLevel.fmt(severity), prefix, msg),
+                severity,
+            )
 
-    def _complete_kwargs(self, connections, kwargs, traceback_msg=None):
+    def _complete_kwargs(self, connections, kwargs, traceback_msg=None, fail=False):
         if "warnings" in kwargs:
-            logs = list(kwargs["warnings"])
+            warning_logs = list(kwargs["warnings"])
```
tyll, line 1571: Just realize that this if could probably be simplified as:

```python
warning_logs = list(kwargs.get("warnings", []))
```

Not sure, why the `list()` is there - @thom311 do you know?

tasks/main.yml
```diff
@@ -64,6 +64,10 @@
     ignore_errors: "{{ network_ignore_errors | default(omit) }}"
     force_state_change: "{{ network_force_state_change | default(omit) }}"
     connections: "{{ network_connections | default([]) }}"
+  register: network_connections_result
```
tyll, line 67: Another thought, can you please rename the variable to `__network_connections_result` to indicate (for now), that it is internal? This makes it more clear that we do not promise anything about it.

tyll:
> Is it recommended to run every integration test before submitting new code (I guess it is recommended) and if so, is there any command or shortcut to run all playbooks?

Yes, this would be great. I use this wrapper when I need to run all scripts locally:

https://github.com/tyll/tools/blob/master/run-standard-test

It assumes that there are images stored in `~/cloud-images`. I don't do this all the time but try to estimate which tests might be affected.

library/network_connections.py
elvgarrui: I'm not sure either, it passed the tests with this change and without the `list()`

tasks/main.yml
elvgarrui: Ok, fixed!

library/network_connections.py
```diff
@@ -1559,24 +1559,36 @@ def _complete_kwargs_loglines(self, rr, connections, idx):
...
+        for severity, msg, idx in rr["log"]:
+            yield (
+                idx,
+                "[%03d] %s %s: %s" % (idx, LogLevel.fmt(severity), prefix, msg),
+                severity,
+            )
 
-    def _complete_kwargs(self, connections, kwargs, traceback_msg=None):
+    def _complete_kwargs(self, connections, kwargs, traceback_msg=None, fail=False):
         if "warnings" in kwargs:
```
tyll, line 1570: You  can remove the full if/else block with just

```python
warning_logs = kwargs.get("warnings", [])
```

It works regardless of whether "warnings" is in kwargs.

elvgarrui: You are right, I didn't see that, thanks!

tyll:
[citest commit:fcd14486255c758770227d324c5ed3b3f90fe30c]

tyll:
The CI fails in the reapply test, it needs to be ported to the new log output. Please take a look.

tyll:
[citest commit:030c29c44edc3658d3c44ab34ec5c364fce5b429]

PR #211, opened 2020-05-01
use selectattr instead of json_query

richm:
Some tests were getting the following error:
```
Unexpected templating type error occurred on ({{
     network_connections |
       json_query('[*][ieee802_1x]') | flatten | count > 0 }}): 'NoneType'
       object is not iterable"
```
In general it is better to avoid the use of `json_query` and use
the older built-in jinaj2 filters such as `selectattr`, unless you
need to perform a complex query of complex json data.  In this
case, not sure what the problem was with the input json, but
using `selectattr` seems to have fixed it.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -22,7 +22,7 @@ network_provider: "{{ network_provider_current }}"
 
 # wpa_supplicant is required if any ieee802_1x connections are defined
 wpa_supplicant_required: "{{ network_connections |
-      json_query('[*][ieee802_1x]') | flatten | count > 0 }}"
+      selectattr('ieee802_1x', 'defined') | list | count > 0 }}"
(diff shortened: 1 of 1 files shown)


richm:
@a-mere-peasant

a-mere-peasant:
I can confirm this , was getting this templating error while running a test for  #200 which is why I have used the same while writing this test https://github.com/linux-system-roles/network/pull/200#issuecomment-623049722 and using `selectattr`  fixed the problem. Thanks again for all the help .

tyll:
There is another instance of `json_query()` in the file, can/should we get rid of it, too?

richm:
> There is another instance of `json_query()` in the file, can/should we get rid of it, too?

Yes, although if it is currently working, and it appears to be working, then it isn't a high priority.

PR #212, opened 2020-05-03
Add support for domain-suffix-match 802.1x option

jaedolph:
Role now supports validating the domain name of the EAP server certificate.
Regenerated the CA certificate as the private key for the original CA has been lost.
Updated test certificates to include a domain name in the CN so the domain-suffix-match
feature can be tested.


jaedolph:
>     * Overall coverage decreased (**-0.03%**) to **38.746%**

I believe the coveralls test is failing due to our workaround with the pygobject libraries. The code I added in the network_connections.py file are not tested in the travis environment, but I can see they are covered when I run the tests on my machine with the pygobject libraries installed.

tyll:
> > ```
> > * Overall coverage decreased (**-0.03%**) to **38.746%**
> > ```
> 
> I believe the coveralls test is failing due to our workaround with the pygobject libraries. The code I added in the network_connections.py file is not tested in the travis environment, but I can see they are covered when I run the tests on my machine with the pygobject libraries installed.

Thank you for the info. I guess it would be better to run (at least some) unit tests in containers for supported OSes or install pygobject in travis (test PR in #211). Since the unit tests also run test-harness, maybe we can report the coverage from there, too.

PR #215, opened 2020-05-05
Cleanup travix/tox config

tyll:


.travis/config.sh
```diff
@@ -39,7 +39,7 @@
 #
 #       - RUN_FLAKE8_DISABLED
 #       - RUN_FLAKE8_EXTRA_ARGS
-if [[ "$(python2 -c "import sys; print(sys.version_info.major)")" == "2" ]]
+if [[ "$(python -c "import sys; print(sys.version_info.major)")" == "2" ]]
```
richm, line 42: You should use the shell function `lsr_check_python_version` https://github.com/linux-system-roles/template/blob/master/.travis/utils.sh#L105 e.g.
```
if lsr_check_python_version python -lt 3.0
```
Then the problem is that `utils.sh` is not sourced by all .travis scripts, so you may need to conditionally source it:
```
type -f lsr_check_python_version > /dev/null 2>&1 || . ${SCRIPTDIR}/utils.sh
```

tyll: thx, changed in #217

PR #219, opened 2020-05-06
tasks: Check directly if NetworkManager is running

### LINKED ISSUE #218: Service facts seem to fail on Fedora 32
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-network-pull-linux-system-roles_network-200-2b7e16a-fedora-32-20200506-161452/artifacts/ansible.log

```
TASK [linux-system-roles.network : Check which services are running] ***********
task path: /tmp/tmpj7fr713o/tests/roles/linux-system-roles.network/tasks/main.yml:4
fatal: [/cache/Fedora-Cloud-Base-32-1.6.x86_64.qcow2]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": false}
```

Ideally, we should get rid of the service_facts module and implement a custom module that checks the system and determines what we need to know. This might be reported at some point. However, it would be great to fix Fedora 32 rather, soon, regardless of this.

tyll:
The service_facts module broke for Fedora 32. Instead of relying on it,
just check directly if NetworkManager is running to determine if it is
the current provider.

closes #218

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,9 +1,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # get service facts, used in defaults/main.yml
 ---
-- name: Check which services are running
-  service_facts:
-  no_log: true
+# If NetworkManager is running, assume that 'nm' is currently in-use,
+# otherwise initscripts
+- name: Get current provider
+  shell: |
+    for f in /proc/*/exe;
+    do [[ "$(readlink -f "${f}")" == /usr/sbin/NetworkManager ]] \
+      && echo nm && break
+    done || echo -n initscripts
+  register: __network_current_provider
+  when: true
 
 # needed for ansible_facts.packages
 - name: Check which packages are installed
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,13 +10,7 @@ network_provider_os_default: "{{
       'OracleLinux'
     ] and ansible_distribution_major_version is version('7', '<')
     else 'nm' }}"
-# If NetworkManager.service is running, assume that 'nm' is currently in-use,
-# otherwise initscripts
-__network_provider_current: "{{
-    'nm' if 'NetworkManager.service' in ansible_facts.services and
-        ansible_facts.services['NetworkManager.service']['state'] == 'running'
-        else 'initscripts'
-    }}"
+__network_provider_current: "{{ __network_current_provider.stdout }}"
 # Default to the auto-detected value
 network_provider: "{{ __network_provider_current }}"
 


tasks/main.yml
```diff
@@ -1,9 +1,14 @@
...
-- name: Check which services are running
-  service_facts:
-  no_log: true
+# If NetworkManager is running, assume that 'nm' is currently in-use,
+# otherwise initscripts
+- name: Get current provider
+  shell: |
+    for f in /proc/*/exe;
+      do [[ "$(readlink -f "${f}")" == /usr/sbin/NetworkManager ]] && echo nm && break
+    done || echo -n initscripts
```
richm, line 10: hmm - what about this:
```
command: systemctl is-enabled NetworkManager.service
```
that will tell you if the service is enabled.  If you want to find out if it is active:
```
command: systemctl is-active NetworkManager.service
```
man systemctl
```
       is-active PATTERN...
           Check whether any of the specified units are active (i.e. running).
           Returns an exit code 0 if at least one is active, or non-zero
           otherwise. Unless --quiet is specified, this will also print the
           current unit state to standard output.
```

pcahyna:
Sorry, I am against reimplementing pieces of Ansible in shell code. If Ansible is broken for Fedora 32, we have to disable it in the CI.

tasks/main.yml
tyll: This was my initial idea, too, however there is no systemctl on CentOS 6, so it would require more checks...

tyll: Also I tried to avoid too many extra tools to make it work in containers, too. Currently, it only depends on coreutils.

tyll:
> Sorry, I am against reimplementing pieces of Ansible in shell code.

The module is now broken for the second time since it was being used in this role. Also the timesync role uses a shell module for a similar approach: https://github.com/linux-system-roles/timesync/blob/master/library/timesync_provider.sh
What is the actual problem here with this patch? The code is a lot simpler than the Ansible code. Once we can drop support for CentOS 6, we can easily adjust it to use systemctl.

> If Ansible is broken for Fedora 32, we have to disable it in the CI.

Fedora 32 was released recently, so it is too late to ignore it in the CI. Leaving the role intentionally broken for Fedora users is very bad in my opinion.

pcahyna:
>  Leaving the role intentionally broken for Fedora users is very bad in my opinion.

It is primarily the Ansible module that is broken for Fedora users, the role being broken is just a corollary. I filed https://bugzilla.redhat.com/show_bug.cgi?id=1832625 and a PR, which will help everyone, not just the network role users.

pcahyna:
If the goal is to check whether NetworkManager is running, isn't a good option to ask NetworkManager directly using dbus, instead of checking the existence of the process or asking the init/service system?

pcahyna:
```sh
    dbus-send --system --print-reply \
        --dest=org.freedesktop.DBus \
        /org/freedesktop/DBus \
        org.freedesktop.DBus.GetNameOwner \
        string:"org.freedesktop.NetworkManager" >/dev/null 2>&1

```
(from `/etc/sysconfig/network-scripts/network-functions`)

tyll:
> ```shell
>     dbus-send --system --print-reply \
>         --dest=org.freedesktop.DBus \
>         /org/freedesktop/DBus \
>         org.freedesktop.DBus.GetNameOwner \
>         string:"org.freedesktop.NetworkManager" >/dev/null 2>&1
> ```
> 
> (from `/etc/sysconfig/network-scripts/network-functions`)
> 
> https://github.com/fedora-sysv/initscripts/blob/3ac46fd2fd5222917efb02eaa6abd9eae90086b0/network-scripts/network-functions#L285

This adds a dependency on dbust-tools which is not installed by default and actually seems to be missing as a dependency for network-scripts, too.

tyll:
Filed a Fedora bug for network-scripts: https://bugzilla.redhat.com/show_bug.cgi?id=1832897

tyll:
The problem with `service_facts` also masks another problem with a test in the CI, see https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-network-pull-linux-system-roles_network-219-1ecd966-fedora-32-20200506-214729/artifacts/test.log.html

tyll:
> If the goal is to check whether NetworkManager is running, isn't a good option to ask NetworkManager directly using dbus, instead of checking the existence of the process or asking the init/service system?

IMHO, checking for the process is good enough. What's the problem that you see with this check?

tasks/main.yml
pcahyna: Note that the approach in timesync also uses `systemctl` or `service` ... is it causing real problems? How would starting and enabling the service work in a container that does not have `systemctl` or `service`?

tyll: Using systemctl means more complexity:

1. If systemctl is not installed -> initscripts
2. If systemctl is installeed
2.1. If NetworkManager is running -> nm
2.2 If NetworkManager is not running -> initscripts

The note about avoiding extra dependencies is why I chose that approach instead of using `ps` or `grep` etc.

However, it would be great if you describe the problems with the current approach, so we can find a solution for the problem instead of discussing alternative solutions without a clear explanation what problem they solve that the proposed solution does not.

pcahyna:
> > Sorry, I am against reimplementing pieces of Ansible in shell code.
> 
> The module is now broken for the second time since it was being used in this role.

It definitely then warrants a separate discussion on how to prevent it in the future, thanks for bringing it up.

> Also the timesync role uses a shell module for a similar approach: https://github.com/linux-system-roles/timesync/blob/master/library/timesync_provider.sh

Note that I also advocated using `service_facts` at that time if you look at the history of that module. But the situation is not analogous because `service_facts` did not provide the functionality needed (at least not at that time) and the logic is more complex (also it is encapsulated in a module which is where complex shell(/python/whatever programming language) logic properly belongs instead of "one-liners" in the playbook, and it uses documented interfaces like `systemctl` and `service`.)

> What is the actual problem here with this patch? The code is a lot simpler than the Ansible code. 

Simplicity must be in the eye of the beholder, because to me accessing a data structure which already holds the information that we need is much simpler than invoking a complex shell command riddled with implicit implementation assumptions on how things work internally (are you sure that NetworkManager will always be at `/usr/sbin/NetworkManager` or that it will be the running executable or that `shell` is actually invoking bash [*] ?). The use of `shell` is also contrary to Ansible and ours best practices (https://github.com/oasis-roles/meta_standards/pull/21) which advise to use dedicated modules and if not possible, prefer `command` over `shell`. Application to our situation: the dedicated module is `service_facts`, or we can ask a documented command which knows the answer (I suggested `dbus-send` but something like `nmcli --terse --fields state general` could be also a viable option, maybe even for the `network-scripts` package problem).

Concerning more specific problems: how will this work in check mode or how will it avoid reporting changes when the role does not need to make any? What does `when: true` mean (literally a truism)?

> Once we can drop support for CentOS 6, we can easily adjust it to use systemctl.

It is not just CentOS 6, it it RHEL 6 too, right? Which is years ahead (EUS etc.)

> > If Ansible is broken for Fedora 32, we have to disable it in the CI.
> 
> Fedora 32 was released recently, so it is too late to ignore it in the CI.

Quite to the contrary, it is too early to add it, since Ansible itself does not seem to have Fedora 32 in their CI, if I understand correctly https://github.com/ansible/ansible/pull/69222 . (There is a nice expressions in French: "aller plus vite que la musique".)

[*] while this role supports only distributions where those assumptions may be to some extent warranted, there is interest on extending our roles to support other distributions, so this sets a bad example for others.

tasks/main.yml
pcahyna: Are you actually worried about NetworkManager running on EL 6? From 1. it seems that you are not and if this is the case, it simplifies things, because if `systemctl` is not installed, it will fail, just as in case `systemctl is-active` is invoked on newer systems with NetworkManager not running.

tyll: > Are you actually worried about NetworkManager running on EL 6? From 1. it seems that you are not and if this is the case, it simplifies things, because if `systemctl` is not installed, it will fail, just as in case `systemctl is-active` is invoked on newer systems with NetworkManager not running.

No, I don't care about NetworkManager running on EL 6. What is the full pattern with `systemctl is-active` that you propose, maybe this will clarify things. For, the problem with 1. was that it would fail the command module while 2.2 would report a different return code. But it seems that a different return code fails the module as well.

Something like this seems to work, too. Do you like this better? Do you prefer a construct with failing module that is ignored and then the failure is checked?
```
ansible -i rhel6, -m shell -a "systemctl is-enabled NetworkManager.service &>/dev/null  && echo nm || echo initscripts" all
```

tyll:
> > What is the actual problem here with this patch? The code is a lot simpler than the Ansible code.
> 
> Simplicity must be in the eye of the beholder, because to me accessing a data structure which already holds the information that we need is much simpler than invoking a complex shell command riddled with implicit implementation assumptions on how things work internally (are you sure that NetworkManager will always be at `/usr/sbin/NetworkManager` or that it will be the running executable or that `shell` is actually invoking bash [*] ?). The use of `shell` is also contrary to Ansible and ours best practices ([oasis-roles/meta_standards#21](https://github.com/oasis-roles/meta_standards/pull/21)) which advise to use dedicated modules and if not possible, prefer `command` over `shell`. Application to our situation: the dedicated module is `service_facts`, or we can ask a documented command which knows the answer (I suggested `dbus-send` but something like `nmcli --terse --fields state general` could be also a viable option, maybe even for the `network-scripts` package problem).

The complexity is in the `service_facts` module that optimizes for a different use-case (getting the information for all services) while the role is only interested in one service. Using `busctl --system call org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus GetNameOwner s org.freedesktop.NetworkManager >/dev/null 2>&1` would work for me, too. 

> Concerning more specific problems: how will this work in check mode or how will it avoid reporting changes when the role does not need to make any? What does `when: true` mean (literally a truism)?

Good points, I will take a look into this.

> > > If Ansible is broken for Fedora 32, we have to disable it in the CI.
> > 
> > 
> > Fedora 32 was released recently, so it is too late to ignore it in the CI.
> 
> Quite to the contrary, it is too early to add it, since Ansible itself does not seem to have Fedora 32 in their CI, if I understand correctly [ansible/ansible#69222](https://github.com/ansible/ansible/pull/69222) . (There is a nice expressions in French: "aller plus vite que la musique".)

In my opinion, the beat comes from Fedora not from Ansible. So when Fedora is released, the role should be ready instead of starting to develop against it. This is what Fedora Rawhide or Branched are there for.

> [*] while this role supports only distributions where those assumptions may be to some extent warranted, there is interest on extending our roles to support other distributions, so this sets a bad example for others.

I am flexible to adjust and adapt when this happens. I do not see a need to adjust to potential requirements that might not be valid anymore when this will happen.

Issue #206, opened 2020-04-24
Change ethtool features to use underscore instead of dashes


tyll:

tyll:
It would be great to introduce a marker for deprecated values ideally with extra information about what should be used instead to be able to do use this information also in the script from #208 

Possible ideas:

Reference by string:
```python
                ArgValidatorBool("esp-hw-offload", default_value=None, deprecated_by="esp_hw_offload"),
                ArgValidatorBool("esp_hw_offload", default_value=None),
```

or 

Reference by object:
```python
def __Init__self(self):
    esp_hw_offload=ArgValidatorBool("esp_hw_offload", default_value=None),
     ArgValidatorDict.__init__(
        self,
        name="features",
        nested=[
            ArgValidatorBool("esp-hw-offload", default_value=None, deprecated_by=esp_hw_offload),
            esp_hw_offload
```
I guess the string makes it easier in case a value moves to a different subtree of the dictionary. Other ideas are welcome, too.

PR #223, opened 2020-05-10
library: Change ethtool features to use underscores

elvgarrui:
Ethtool features should use underscores instead of dashes. A
warning shows in case dashes used, and it fails if underscore and dashes are
mixed.

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -199,6 +199,12 @@ class ArgValidatorBool(ArgValidator):
         raise ValidationError(name, "must be an boolean but is '%s'" % (value))
 
 
+class ArgValidatorDeprecated:
+    def __init__(self, name, deprecated_by):
+        self.name = name
+        self.deprecated_by = deprecated_by
+
+
 class ArgValidatorDict(ArgValidator):
     def __init__(
         self,
@@ -222,26 +228,33 @@ class ArgValidatorDict(ArgValidator):
             items = list(value.items())
         except AttributeError:
             raise ValidationError(name, "invalid content is not a dictionary")
-        for (k, v) in items:
-            if k in seen_keys:
-                raise ValidationError(name, "duplicate key '%s'" % (k))
-            seen_keys.add(k)
-            validator = self.nested.get(k, None)
-            if validator is None:
-                raise ValidationError(name, "invalid key '%s'" % (k))
+        for (setting, value) in items:
             try:
-                vv = validator._validate(v, name + "." + k)
+                validator = self.nested[setting]
+            except KeyError:
+                raise ValidationError(name, "invalid key '%s'" % (setting))
+            if isinstance(validator, ArgValidatorDeprecated):
+                setting = validator.deprecated_by
+                validator = self.nested.get(setting, None)
+            if setting in seen_keys:
+                raise ValidationError(name, "duplicate key '%s'" % (setting))
+            seen_keys.add(setting)
+            try:
+                validated_value = validator._validate(value, name + "." + setting)
             except ValidationError as e:
                 raise ValidationError(e.name, e.error_message)
-            result[k] = vv
-        for (k, v) in self.nested.items():
-            if k in seen_keys:
+            result[setting] = validated_value
+        for (setting, validator) in self.nested.items():
+            if setting in seen_keys or isinstance(validator, ArgValidatorDeprecated):
                 continue
-            if v.required:
-                raise ValidationError(name, "missing required key '%s'" % (k))
-            vv = v.get_default_value()
-            if not self.all_missing_during_validate and vv is not ArgValidator.MISSING:
-                result[k] = vv
+            if validator.required:
+                raise ValidationError(name, "missing required key '%s'" % (setting))
+            default_value = validator.get_default_value()
+            if (
+                not self.all_missing_during_validate
+                and default_value is not ArgValidator.MISSING
+            ):
+                result[setting] = default_value
         return result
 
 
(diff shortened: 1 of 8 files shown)


elvgarrui:
I've been trying to get this issue done for the last couple of days and I have several questions regarding how to continue working on this issue. I list them all briefly on this comment:

1. Right now, I don't see the warning for dashes I created as an output in any test and I'm not really sure why. I think it is because I need to catch the warning it on `library/network_connections.py`, maybe on the function `_check_ethtool_setting_support`? But since that only applies to NM and not initscript I'm not totally sure about it, I'd like to have some extra opinion on this matter. @tyll 

2. Regarding the fact that dashes and underscores cannot be together, I also think it should be checked in the same function in network_connections as in point 1 (exactly when it checks if the specified_features are in the supported feature list, line 2018 right now). The problem here is that some old ethtool properties contains both dashes and underscores so I'm confused about how to make an efficient comparison of that.

Thanks for the attention and sorry for the long post!

tyll:
For the warnings: you need to log them with loglevel warn. For the other items, I will take a look later or tomorrow. Thanks.

module_utils/network_lsr/argument_validator.py
```diff
@@ -93,6 +96,10 @@ def get_default_value(self):
             return self.default_value
 
     def validate(self, value):
+        if self.deprecated_by:
+            msg = f"{self.name} is deprecated by {self.deprecated_by}."
+            "Please, consider using underscores instead of dashes."
```
tyll, line 101:
```suggestion
```

The dashes/underscore message is

tyll:
> I've been trying to get issue #206 done for the last couple of days and I have several questions regarding how to continue working on this issue. I list them all briefly on this comment:
> 
> 1. Right now, I don't see the warning for dashes I created as an output in any test and I'm not really sure why. I think it is because I need to catch the warning it on `library/network_connections.py`, maybe on the function `_check_ethtool_setting_support`? But since that only applies to NM and not initscript I'm not totally sure about it, I'd like to have some extra opinion on this matter. @tyll

My idea would be to address it in:
https://github.com/linux-system-roles/network/blob/4753c70c55ef27297d4795ffa33a333a762cfddc/library/network_connections.py#L1648

with something like:

```python
with warnings.catch_warnings() as validation_warnings:
   c = ....
for warn_message in validation_warnings:
    self.log_warn(...)
```

The same might need to be done when `self._connection_validator.validate_connection_one()` is called. What do you think?
For reference:
https://docs.python.org/3/library/warnings.html#testing-warnings

> 2. Regarding the fact that dashes and underscores cannot be together, I also think it should be checked in the same function in network_connections as in point 1 (exactly when it checks if the specified_features are in the supported feature list, line 2018 right now). The problem here is that some old ethtool properties contains both dashes and underscores so I'm confused about how to make an efficient comparison of that.

This check should be part of the validation. It is enough to check this only for the settings that are deprecated. So if a deprecated setting is specified and also the non-deprecated one, the validation should fail. What's your opinion?

module_utils/network_lsr/argument_validator.py
```diff
@@ -93,6 +96,10 @@ def get_default_value(self):
             return self.default_value
 
     def validate(self, value):
+        if self.deprecated_by:
+            msg = f"{self.name} is deprecated by {self.deprecated_by}."
```
tyll, line 100: Unfortunately, the module needs to work with Python 2.6, as well. Therefore this syntax does not work. You can probably use:
`"{0.name} is deprecated by {0.deprecated_by}".format(self)`

module_utils/network_lsr/argument_validator.py
elvgarrui: Fixed this, thanks!

module_utils/network_lsr/argument_validator.py
elvgarrui: I think this is fixed now too, thanks!

elvgarrui:
@tyll 
> 
> ```python
> with warnings.catch_warnings() as validation_warnings:
>    c = ....
> for warn_message in validation_warnings:
>     self.log_warn(...)
> ```

The log warn function needs the idx of the connection, which is only known later. Should I try to pass it as part of the warning string? I cannot think of a better solution, but there should be one.

Most importantly, I fixed the unit testing and it works with the underscore validators, but now the integration testing (`playbook/test_ethtool_features.py`) fails. I was able to find out that if I comment line 750:
```ArgValidatorBool("tx_tcp_segmentation", default_value=None)```
it works again. I've been trying hard to understand why this is happening but I am out of ideas. Any suggestion?

File: tests/unit/test_network_connections.py [hunk 3]
```diff
@@ -69,56 +69,95 @@ def pprint(msg, obj):
 VALIDATE_ONE_MODE_NM = ARGS_CONNECTIONS.VALIDATE_ONE_MODE_NM
 
 ETHTOOL_FEATURES_DEFAULTS = {
+    "esp_hw_offload": None,
```
tyll, line 72: Actually, the defaults should only contain the values with dashes (or only the values with underscores) but not both. So this change masks an implementation problem

tyll:
> Most importantly, I fixed the unit testing and it works with the underscore validators, but now the integration testing (`playbook/test_ethtool_features.py`) fails. I was able to find out that if I comment line 750:
> `ArgValidatorBool("tx_tcp_segmentation", default_value=None)`
> it works again. I've been trying hard to understand why this is happening but I am out of ideas. Any suggestion?

The problem is that code is missing to ensure that the argument validator only returns the values with dashes as this is what the Ansible module (`library/network_connections.py`) expects. As an alternative, the validator could only return values with underscores. Then the module needs to be adjusted. The initscript provider needs dashes and the NM provider underscores. #234 already implements some other ethtool settings using only underscores, so changing the validator to return only underscores seems to be the right approach.

After taking a look at the code, it seems to me that the code for the [`ArgValidatorDict._validate_impl()`](https://github.com/linux-system-roles/network/blob/master/module_utils/network_lsr/argument_validator.py#L218) method needs to be updated. There needs to be check that if something is specified that is deprecated, that the value of `deprecated_by` is used as the key for the result dictionary. Also it needs to be checked that the regular value and the deprecated value are not specified at the same time.

I guess it would be good to prepare a few unit tests for at least one feature (`tx_tcp_segmentation` is a good candidate) and test that:

- setting only `tx_tcp_segmentation` to `yes` (and keep the rest as default) will set this to `True` after arg validation
- setting only `tx-tcp-segmentation` to `yes` will will set `tx_tcp_segmentation` to `True` after arg validation
- setting both `tx-tcp-segmentation` and `tx_tcp_segmentation` will raise a `ValidationError`

and also create corresponding integration tests that extend the current one to try both values and check that there are warning when `tx-tcp-segmentation` is used or a failure when both are used. Once you did it for one value, you can hopefully adjust it to easily parameterize it for all values (in the unit tests). I hope this is feasible.

What do you think?

tyll:
> @tyll
> 
> > ```python
> > with warnings.catch_warnings() as validation_warnings:
> >    c = ....
> > for warn_message in validation_warnings:
> >     self.log_warn(...)
> > ```
> 
> The log warn function needs the idx of the connection, which is only known later. Should I try to pass it as part of the warning string? I cannot think of a better solution, but there should be one.

How would you pass it as part of the warning string?
The only solution that I see is to move the deprecation check to `_validate()` and adjust it to also return all warnings and then collect them separate to the Python warnings.

module_utils/network_lsr/argument_validator.py
```diff
@@ -549,58 +556,227 @@ def __init__(self):
             self,
             name="features",
             nested=[
-                ArgValidatorBool("esp-hw-offload", default_value=None),
-                ArgValidatorBool("esp-tx-csum-hw-offload", default_value=None),
-                ArgValidatorBool("fcoe-mtu", default_value=None),
+                ArgValidatorBool("esp_hw_offload", default_value=None),
+                ArgValidatorBool(
+                    "esp-hw-offload", default_value=None, deprecated_by="esp_hw_offload"
```
tyll, line 561: @elvgarrui another thought: Maybe the necessary changes can be simplified by removing `default_value=None` from the items that are deprecated. Maybe it even makes sense to fail when `default_value` and `deprecated_by` are both specified.

tyll: > @elvgarrui another thought: Maybe the necessary changes can be simplified by removing `default_value=None` from the items that are deprecated. Maybe it even makes sense to fail when `default_value` and `deprecated_by` are both specified.

Actually it seems everything needs to have a default value so it might be more complicated to change this.

tyll:
@elvgarrui a note regarding renaming variables: It would be great to put those changes in their own commit to make it easier to review the change.

module_utils/network_lsr/argument_validator.py
tyll: Another thought: We can probably introduce a new class `ArgValidatorDeprecated` and use that here instead of `ArgValidatorBool` for the deprecated values since they are not used for anything and do not need a default value.

module_utils/network_lsr/argument_validator.py
```diff
@@ -549,62 +563,174 @@ def __init__(self):
...
+                ArgValidatorDeprecated(
+                    "tx-checksum-ipv6", deprecated_by="tx_checksum_ipv6",
+                ),
+                ArgValidatorBool("tx_checksum_ip_generic", default_value=None),
+                ArgValidatorDeprecated(
+                    "tx-checksum-ip-generic", deprecated_by="tx_checksum_ip_generic",
+                ),
+                ArgValidatorBool("tx_checksum_sctp", default_value=None),
+                ArgValidatorDeprecated(
+                    "tx-checksum-sctp", deprecated_by="tx_chacksum_sctp",
```
tyll, line 642:
```suggestion
                    "tx-checksum-sctp", deprecated_by="tx_checksum_sctp",
```

elvgarrui: Fixed it, thanks!

tyll:
[citest bad]
Network failure:
fatal: [/cache/rhel-guest-image-7.7-261.x86_64.qcow2]: FAILED! => {"changed": true, "cmd": ["yum", "install", "-y", "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"], "delta": "0:00:00.334276", "end": "2020-06-09 08:36:16.328234", "msg": "non-zero return code", "rc": 1, "start": "2020-06-09 08:36:15.993958", "stderr": "Cannot open: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm. Skipping.\nError: Nothing to do", "stderr_lines": ["Cannot open: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm. Skipping.", "Error: Nothing to do"], "stdout": "Loaded plugins: search-disabled-repos", "stdout_lines": ["Loaded plugins: search-disabled-repos"]}

tests/unit/test_network_connections.py
```diff
@@ -2475,6 +2476,49 @@ def test_full_state_absent_defaults(self):
...
+            "ethtool": expected_ethtool,
+            "interface_name": "5",
+            "persistent_state": "present",
+            "state": None,
+            "type": "ethernet",
+        }
+        self.check_one_connection_with_defaults(input_connection, expected_connection)
+
+    def test_set_ethtool_feature(self):
+        # When passing the name of an non-deprecated ethtool feature, their
```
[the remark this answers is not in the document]
elvgarrui: True! I'm sorry for these. Comments are fixed now.

tests/unit/test_network_connections.py
```diff
@@ -2475,6 +2476,49 @@ def test_full_state_absent_defaults(self):
             },
         )
 
+    def _test_ethtool_changes(self, input_features, expected_features):
+        expected_ethtool = copy.deepcopy(ETHTOOL_FEATURES_DEFAULTS)
+        expected_ethtool.update(expected_features)
+        expected_ethtool = {"features": expected_ethtool}
```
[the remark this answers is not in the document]
elvgarrui: I clarified this with better variable names. thanks!

tests/unit/test_network_connections.py
```diff
@@ -2475,6 +2476,49 @@ def test_full_state_absent_defaults(self):
             },
         )
 
+    def _test_ethtool_changes(self, input_features, expected_features):
```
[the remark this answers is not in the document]
elvgarrui: Done!

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -79,6 +79,68 @@
...
+                var: __network_connections_result
+            - assert:
+                that:
+                  - "{{ 'duplicate key' in __network_connections_result.msg }}"
+                  - "{{ __network_connections_result.changed == false }}"
+        - name: Check that no changes from the block were applied
+          debug:
+            var: __network_connections_result
+        - assert:
+            that: "{{ __network_connections_result.changed == false }}"
```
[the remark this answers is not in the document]
elvgarrui: I've changed this as you have mentioned here, thanks! It makes more sense now.

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -79,6 +79,68 @@
...
+                        tx-tcp-segmentation: "no"
+              register: __network_connections_result
+          rescue:
+            - name: Avoid duplicate key
+              debug:
+                var: __network_connections_result
+            - assert:
+                that:
+                  - "{{ 'duplicate key' in __network_connections_result.msg }}"
+                  - "{{ __network_connections_result.changed == false }}"
```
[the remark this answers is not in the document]
elvgarrui: Fixed!

module_utils/network_lsr/argument_validator.py
```diff
@@ -549,62 +562,174 @@ def __init__(self):
...
+                ArgValidatorBool("tx_vlan_stag_hw_insert", default_value=None),
+                ArgValidatorDeprecated(
+                    "tx-vlan-stag-hw-insert", deprecated_by="tx_vlan_stag_hw_insert",
+                ),
             ],
         )
         self.default_value = dict(
-            [(k, v.default_value) for k, v in self.nested.items()]
+            [
+                (k, v.default_value)
```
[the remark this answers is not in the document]
elvgarrui: Fixed, thanks!

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -79,6 +79,68 @@
...
+        - name:
+          debug:
+            var: ethtool_features.stdout_lines
+        - name: Assert device features
+          assert:
+            that:
+              - >-
+                  'tx-tcp-segmentation: on' in
+                  ethtool_features.stdout_lines | map('trim')
+        - name: Change tx_tcp_segmentation and tx-tcp-segmentation.
```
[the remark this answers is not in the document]
elvgarrui: Fixed. Thanks!

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -79,6 +79,68 @@
...
+                    ip:
+                      dhcp4: "no"
+                      auto6: "no"
+                    ethtool:
+                      features:
+                        tx_tcp_segmentation: "no"
+                        tx-tcp-segmentation: "no"
+              register: __network_connections_result
+          rescue:
+            - name: Avoid duplicate key
```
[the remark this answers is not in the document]
elvgarrui: Done!

module_utils/network_lsr/argument_validator.py
```diff
@@ -549,62 +562,174 @@ def __init__(self):
             self,
             name="features",
             nested=[
-                ArgValidatorBool("esp-hw-offload", default_value=None),
-                ArgValidatorBool("esp-tx-csum-hw-offload", default_value=None),
-                ArgValidatorBool("fcoe-mtu", default_value=None),
+                ArgValidatorBool("esp_hw_offload", default_value=None),
+                ArgValidatorDeprecated(
```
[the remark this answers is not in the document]
elvgarrui: I created this test. I used a simple assert instead of the already defined AssertTrue function to see a more complete fail message. Thanks!

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -79,6 +79,68 @@
...
+              debug:
+                var: __network_connections_result
+            - assert:
+                that:
+                  - "{{ 'duplicate key' in __network_connections_result.msg }}"
```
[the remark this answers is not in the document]
elvgarrui: Done, thanks!

tests/unit/test_network_connections.py
```diff
@@ -2475,6 +2476,74 @@ def test_full_state_absent_defaults(self):
...
+        """
+        Test that for each validator in
+        ArgValidator_DictEthtoolFeatures.nested there is another non-deprecated
+        validator that has the name from the deprecated_by attribute"
+        """
+        validators = (
+            network_lsr.argument_validator.ArgValidator_DictEthtoolFeatures().nested
+        )
+        for name, validator in validators.items():
+            if type(validator) is network_lsr.argument_validator.ArgValidatorDeprecated:
```
tyll, line 2544: You are using isinstance in the other code and it seems to be correct here, too:

```suggestion
            if isinstance(validator, network_lsr.argument_validator.ArgValidatorDeprecated):
```

tests/unit/test_network_connections.py
```diff
@@ -2475,6 +2476,74 @@ def test_full_state_absent_defaults(self):
             },
         )
 
+    def _test_ethtool_changes(self, input_features, expected_features):
+        """
+        When passing a dictionary 'input_features' each feature and their value
```
tyll, line 2481:
```suggestion
        When passing a dictionary 'input_features' with each feature and their value
```

tyll:
> Awesome, I only left minor remarks. One thing that I missed before is that the README needs to be updated to only mention the new values with underscores now:
> [linux-system-roles/network#ethtool](https://github.com/linux-system-roles/network#ethtool)

Also missed the example. FYI: I updated both.

elvgarrui:
> Also missed the example. FYI: I updated both.

Ok, I've only updated the unit test then.

tyll:
[citest bad]

PR #225, opened 2020-05-11
Prepare tests for Fedora 32

tyll:
Fix some issues found when working on #219

### DIFF
--- a/tests/playbooks/tests_checkpoint_cleanup.yml
+++ b/tests/playbooks/tests_checkpoint_cleanup.yml
@@ -18,6 +18,10 @@
     - linux-system-roles.network
   tasks:
     - block:
+        # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1832897
+        - package:
+            name: dbus-tools
+            state: present
         # create test profile
         - include_role:
             name: linux-system-roles.network
--- a/tests/playbooks/tests_checkpoint_cleanup.yml
+++ b/tests/playbooks/tests_checkpoint_cleanup.yml
@@ -54,14 +54,16 @@
               dropped.
             that: error_trigger.failed
         # yamllint disable-line rule:line-length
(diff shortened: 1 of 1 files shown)


tests/playbooks/tests_checkpoint_cleanup.yml
```diff
@@ -18,6 +18,10 @@
     - linux-system-roles.network
   tasks:
     - block:
+        # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1832897
+        - package:
+            name: dbus-tools
+            state: present
```
richm, line 24: Is this needed if not using initscripts?  I can't tell if this playbook includes `nm` specific tasks, or perhaps it is needed in either case?
Is this package needed and provided on EL6 through EL8?  Or is this playbook only executed on certain platforms?

tyll: This playbook does not run on EL6. The test uses both initscripts and nm to trigger an error condition but this only works when dbus-tools is present. initscripts does not really need dbus-tools it is only needed when NM is also running to tell NM not to touch an interface.

tests/playbooks/tests_checkpoint_cleanup.yml
```diff
@@ -67,6 +73,9 @@
                   - name: "{{ interface }}"
                     persistent_state: absent
                     state: down
+            - file:
+                dest: /etc/sysconfig/network-scripts/ifcfg-cptstbr
+                state: absent
```
pcahyna, line 78: Is this also a workaround for some bug? otherwise lgtm.

tyll: It is the quick cleanup for the initscripts provider. But now that I look at it again, I noticed it is missing the `{{ interface }}` expansion.

PR #227, opened 2020-05-13
library: all the imports must be on the top

ffmancera:
According to PEP8 [1], the import are always put at the top of the file,
just after any module comments and docstrings, and before module globals
and constants.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -2,13 +2,17 @@
 # -*- coding: utf-8 -*-
 # SPDX-License-Identifier: BSD-3-Clause
 
+import errno
 import functools
 import os
+import re
+import shlex
 import socket
 import time
 import traceback
 
 # pylint: disable=import-error, no-name-in-module
+from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.network_lsr import MyError
 
 # pylint: disable=import-error
@@ -107,7 +111,6 @@ class SysUtil:
             out = Util.check_output(["ethtool", "-P", ifname])
         except MyError:
             return None
-        import re
 
         m = re.match("^Permanent address: ([0-9A-Fa-f:]*)\n$", out)
         if not m:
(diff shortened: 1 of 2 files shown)


ffmancera:
[citest commit:f15c5bf648e0c56f11f2003cb4f4555759def266]

tyll:
[citest bad]

tyll:
> [citest commit:f15c5bf648e0c56f11f2003cb4f4555759def266]

FYI: This is only needed for PRs from external contributors.

Are you planning to fix the import order for everything? Currently it is not checked by the CI because there are some special cases. It might be possible to ignore them individually and enable the check: https://github.com/linux-system-roles/network/blob/master/tox.ini#L186
https://github.com/linux-system-roles/network/blob/master/tox.ini#L186

ffmancera:
> > [citest commit:f15c5bf648e0c56f11f2003cb4f4555759def266]
> 
> FYI: This is only needed for PRs from external contributors.
> 

Oh I didn't know that. Thanks!

> Are you planning to fix the import order for everything? Currently it is not checked by the CI because there are some special cases. It might be possible to ignore them individually and enable the check: https://github.com/linux-system-roles/network/blob/master/tox.ini#L186
> https://github.com/linux-system-roles/network/blob/master/tox.ini#L186

Yes it would be nice to do it for all the cases and ignore the exceptions individually. I think this improves the code readability. What do you think? Thanks!

tyll:
> > Are you planning to fix the import order for everything? Currently it is not checked by the CI because there are some special cases. It might be possible to ignore them individually and enable the check: https://github.com/linux-system-roles/network/blob/master/tox.ini#L186
> > https://github.com/linux-system-roles/network/blob/master/tox.ini#L186
> 
> Yes it would be nice to do it for all the cases and ignore the exceptions individually. I think this improves the code readability. What do you think? Thanks!

Sure, in the long term it makes sense. Not sure how much effort it would be at this point. Would you like to take a look? If it is too much effort, I am fine with merging the PR as-is.

ffmancera:
Rebased, thanks!

PR #231, opened 2020-05-14
ethtool: use GPERMADDR instead the ethtool command line tool

ffmancera:
This patch implements the ETHTOOL_GPERMADDR command in order to retrieve
the permanent address from ethtool instead using command line tool.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -34,7 +34,7 @@ _network_packages_default_gobject_packages: ["python{{
       ansible_python['version']['major'] | replace('2', '')}}-gobject-base"]
 
 network_service_name_default_nm: NetworkManager
-network_packages_default_nm: "{{ ['ethtool', 'NetworkManager']
+network_packages_default_nm: "{{['NetworkManager']
       + _network_packages_default_gobject_packages|select()|list()
       + _network_packages_default_802_1x|select()|list()}}"
 
@@ -56,8 +56,8 @@ if ansible_distribution in ['RedHat', 'CentOS', 'OracleLinux'] and
 # |select() filters the list to include only values that evaluate to true
 #     (the empty string is false)
 # |list() converts the generator that |select() creates to a list
-network_packages_default_initscripts: "{{ ['ethtool']
-+ _network_packages_default_initscripts_bridge|select()|list()
+network_packages_default_initscripts: "{{
+_network_packages_default_initscripts_bridge|select()|list()
 + _network_packages_default_initscripts_network_scripts|select()|list()
 }}"
 
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -13,18 +13,20 @@ import traceback
 
 # pylint: disable=import-error, no-name-in-module
 from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.network_lsr import ethtool
 from ansible.module_utils.network_lsr import MyError
 
-# pylint: disable=import-error
 from ansible.module_utils.network_lsr.argument_validator import (
     ArgUtil,
     ArgValidator_ListConnections,
     ValidationError,
 )
 
-# pylint: disable=import-error
 from ansible.module_utils.network_lsr.utils import Util
 from ansible.module_utils.network_lsr import nm_provider
+
+# pylint: enable=import-error, no-name-in-module
+
 
 DOCUMENTATION = """
 ---
@@ -107,15 +109,7 @@ class SysUtil:
 
     @staticmethod
     def _link_read_permaddress(ifname):
-        try:
-            out = Util.check_output(["ethtool", "-P", ifname])
-        except MyError:
-            return None
-
-        m = re.match("^Permanent address: ([0-9A-Fa-f:]*)\n$", out)
-        if not m:
-            return None
-        return Util.mac_norm(m.group(1))
+        return ethtool.get_perm_addr(ifname)
 
     @staticmethod
     def _link_infos_fetch():
--- a/module_utils/network_lsr/ethtool.py
+++ b/module_utils/network_lsr/ethtool.py
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+import array
+import struct
+import fcntl
+import socket
+
+from .utils import Util
+
+ETHTOOL_GPERMADDR = 0x00000020
+SIOCETHTOOL = 0x8946
+MAX_ADDR_LEN = 32
+IFNAMESIZ = 16
+
+
+def get_perm_addr(ifname):
+    """
+    Return the Permanent address value for the specified interface using the
+    ETHTOOL_GPERMADDR ioctl command.
+
+    Please for further documentation, see:
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L734
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L1388
+    https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c#n4172
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        if len(ifname) > IFNAMESIZ:
+            return None
+
+        ecmd = array.array(
+            "B",
+            struct.pack(
+                "II%is" % MAX_ADDR_LEN,
+                ETHTOOL_GPERMADDR,
+                MAX_ADDR_LEN,
+                b"\x00" * MAX_ADDR_LEN,
+            ),
+        )
+        ifreq = struct.pack("%isP" % IFNAMESIZ, ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        try:
+            res = ecmd.tobytes()
+        except AttributeError:  # tobytes() is not available in python2
+            res = ecmd.tostring()
+        _, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
+        perm_addr = Util.mac_ntoa(perm_addr[:size])
+    except IOError:
+        perm_addr = None
+    finally:
+        sock.close()
+
+    return perm_addr
--- a/module_utils/network_lsr/utils.py
+++ b/module_utils/network_lsr/utils.py
@@ -2,9 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # vim: fileencoding=utf8
 
-import os
 import socket
-import subprocess
 import sys
 import uuid
 
@@ -24,19 +22,6 @@ class Util:
             if pred is None or pred(v):
                 return v
         return default
-
-    @staticmethod
-    def check_output(argv):
-        # subprocess.check_output is python 2.7.
-        with open("/dev/null", "wb") as DEVNULL:
-            env = os.environ.copy()
-            env["LANG"] = "C"
-            p = subprocess.Popen(argv, stdout=subprocess.PIPE, stderr=DEVNULL, env=env)
-            # FIXME: Can we assume this to always be UTF-8?
-            out = p.communicate()[0].decode("UTF-8")
-            if p.returncode != 0:
-                raise MyError("failure calling %s: exit with %s" % (argv, p.returncode))
-        return out
 
     @staticmethod
     def path_to_glib_bytes(path):
@@ -275,7 +260,8 @@ class Util:
     def mac_ntoa(mac):
         if mac is None:
             return None
-        return ":".join(["%02x" % c for c in mac])
+        # bytearray() is needed for python2 compatibility
+        return ":".join(["%02x" % c for c in bytearray(mac)])
 
     @staticmethod
     def mac_norm(mac_str, force_len=None):
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -2503,10 +2503,9 @@ class TestNM(unittest.TestCase):
 
 
 class TestUtils(unittest.TestCase):
-    def test_check_output(self):
-        res = Util.check_output(["echo", "test"])
-        self.assertEqual(res, "test\n")
-        self.assertRaises(n.MyError, Util.check_output, ["false"])
+    def test_mac_ntoa(self):
+        mac_bytes = b"\xaa\xbb\xcc\xdd\xee\xff"
+        self.assertEqual(Util.mac_ntoa(mac_bytes), "aa:bb:cc:dd:ee:ff")
 
     def test_convert_passwd_flags_nm(self):
         test_cases = [
@@ -2528,10 +2527,9 @@ class TestUtils(unittest.TestCase):
 
 class TestSysUtils(unittest.TestCase):
     def test_link_read_permaddress(self):
-        # Manipulate PATH to use ethtool mock script to avoid hard dependency on
-        # ethtool
-        os.environ["PATH"] = TESTS_BASEDIR + "/helpers:" + os.environ["PATH"]
-        self.assertEqual(SysUtil._link_read_permaddress("lo"), "23:00:00:00:00:00")
+        self.assertEqual(SysUtil._link_read_permaddress("lo"), "00:00:00:00:00:00")
+        self.assertEqual(SysUtil._link_read_permaddress("fakeiface"), None)
+        self.assertEqual(SysUtil._link_read_permaddress("morethansixteenchars"), None)
 
 
 if __name__ == "__main__":
--- a/tests/unit/helpers/ethtool
+++ b/tests/unit/helpers/ethtool
@@ -1,6 +1,0 @@
-#! /bin/bash
-
-if [ "${1}" == "-P" ] && [ "${2}" != "" ]
-then
-    echo "Permanent address: 23:00:00:00:00:00"
-fi


ffmancera:
There is an issue with python 2.6. I am fixing it..

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+ETHTOOL_GPERMADDR = 0x00000020
+SIOCETHTOOL = 0x8946
+
+
+def get_perm_addr(ifname):
+    """
+    Return the Permanent address value for the specified interface using the
+    ETHTOOL_GPERMADDR ioctl command. The address is returned as string in
+    python2 so it is needed to convert it to a bytearray and convert it again
+    to a hex string again.
```
thom311, line 19: what does this mean? I think the function should consistently either return the address as string or byte array, regardless of the python version.

tests/unit/test_network_connections.py
```diff
@@ -2528,10 +2523,7 @@ def test_convert_passwd_flags_nm(self):
 
 class TestSysUtils(unittest.TestCase):
     def test_link_read_permaddress(self):
-        # Manipulate PATH to use ethtool mock script to avoid hard dependency on
-        # ethtool
-        os.environ["PATH"] = TESTS_BASEDIR + "/helpers:" + os.environ["PATH"]
```
tyll, line 2533: Please remove the helper tool, too

tyll:
Please remove ethtool from https://github.com/linux-system-roles/network/blob/master/defaults/main.yml#L37 and also from the list of packages for initscripts.

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Please for further documentation, see:
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L734
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L1388
+    https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c#n4172
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
```
tyll, line 29: is utf-8 correct here? Can the interfaces contain other characters that ASCII?

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        res = ecmd.tostring()
+        _, _, perm_addr = struct.unpack("II6s", res)
+        if hasattr(perm_addr, "hex"):
+            perm_addr = perm_addr.hex()
+        else:
+            perm_addr = binascii.hexlify(perm_addr)
+            if not perm_addr:
```
tyll, line 40: Why is this if here? When does this trigger?

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+                perm_addr = "00:00:00:00:00:00"
+        perm_addr = ":".join(f + s for f, s in zip(perm_addr[::2], perm_addr[1::2]))
```
tyll, line 42: How does this work if `perm_addr` is already `00:00:00:00:00:00`.?

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        ecmd = array.array("B", struct.pack("II6s", ETHTOOL_GPERMADDR, 6, b"\x00" * 6))
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        res = ecmd.tostring()
+        _, _, perm_addr = struct.unpack("II6s", res)
+        if hasattr(perm_addr, "hex"):
```
tyll, line 36: Since the encoding is complicated because of Python2/Python3 differences, please add this to its own function and add unit tests for this.

module_utils/network_lsr/ethtool.py
ffmancera: That comment is outdated. I fixed that. My bad, the function is returning the address as string always.

module_utils/network_lsr/ethtool.py
ffmancera: According to https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?id=1f420318bda3cc62156e89e1b56d60cc744b48ad#n827 it seems from the kernel is a char[] so interfaces cannot contain other characters that ASCII. I will do the proper encoding.

thom311: the code you linked there shows the opposite.

Like most kernel names (interface names, filenames(!!)) these names are in no defined encoding at all... except, that they must be '\0' terminated and that certain ASCII characters are treated special (`'.'`, `'/'`). That means, you cannot use any arbitrary encoding (e.g. UTF-32 would blow up), but as long as you use an encoding that honours this minimal requirements (like some extension to ASCII (UTF-8, ISO/IEC 8859)), kernel doesn't care.

Of course, since somebody has to type these "names" into a keyboard and print them on the screen, having not defined encoding, is a serious problem. The general recommendation is of course to only use ASCII or valid UTF-8.

For that reason, NetworkManager exposed interface names as ANSI-C backslash encoded strings, that are really binary. E.g. on `nm_device_get_iface()`. Try `ip link add name $'d\xccf\\c' type dummy` and then look at `ip link` and `nmcli device` output.

On the other end, when you configure a NetworkManager profile, you can only use UTF-8 name. See https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/01667694ca75cb3dff137a28313548a6cc4258e7/shared/nm-glib-aux/nm-shared-utils.c#L4748

Treating this as UTF-8 is IMO the best you can do. If the caller provides anything else, it's her problem.

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        ecmd = array.array("B", struct.pack("II6s", ETHTOOL_GPERMADDR, 6, b"\x00" * 6))
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        res = ecmd.tostring()
```
pcahyna, line 34: will be removed in Python 3.9: https://docs.python.org/3/library/array.html#array.array.tostring, OTOH `tobytes` does not seem to be available in 2.7.

module_utils/network_lsr/ethtool.py
tyll: Nice, `ip link add dev 😎 type veth peer 🌞` works, too.

thom311: Also interesting... by default `journalctl` will not print binary data. That means, if you create a non-UTF-8 interface name, NM will log that binary name, and journal will only say `[143B blob data]`. Pass `--all` to journalctl to see it. Of course, then the output of journalctl is no longer valid UTF-8.

module_utils/network_lsr/ethtool.py
ffmancera: I modified it to 00000000. this way the join works :-)

module_utils/network_lsr/ethtool.py
ffmancera: When the mac address is 00:00:00:00:00:00 on python2.6, there is no "tobytes" method so we are getting a string instead. When transforming it to string it is an empty string which is wrong. There is no other solution for this, or at least, I was not able to find it.

module_utils/network_lsr/ethtool.py
ffmancera: I will move them to its own function. Thanks!

module_utils/network_lsr/ethtool.py
ffmancera: I am using `tostring` because is the way to make this on python2 and python3. As python is now deprecated, it means they are not going to make them compatible in future versions. This will be a problem for projects that are currently supporting python2 and python3.

There is no solution for this, IMO.

module_utils/network_lsr/ethtool.py
ffmancera: > Nice, `ip link add dev 😎 type veth peer 🌞` works, too.

I do not want a world where people name interfaces with emojis :-)

module_utils/network_lsr/ethtool.py
tyll, line 34:
```suggestion
        try:
            res = ecmd.tobytes()
        except AtrributeError:  # tobytes() is not available in Python 2
            res = ecmd.tostring()
```

ffmancera: What do you think about `res = ecmd.tobytes() if hasattr(ecmd, "tobytes")` else ecmd.tostring()`? This way we do not need the try except.

module_utils/network_lsr/ethtool.py
tyll: I do not follow, when is `not perm_addr` True?

module_utils/network_lsr/ethtool.py
tyll: Why is it bad to use try/except?

ffmancera: I agree that try/except avoids putting too much information in a single line.

module_utils/network_lsr/ethtool.py
ffmancera: Whops. You are right. I think I missed something then. Thanks for pointing this out. :-)

defaults/main.yml
```diff
@@ -34,8 +34,8 @@ _network_packages_default_gobject_packages: ["python{{
       ansible_python['version']['major'] | replace('2', '')}}-gobject-base"]
 
 network_service_name_default_nm: NetworkManager
-network_packages_default_nm: "{{ ['ethtool', 'NetworkManager']
-      + _network_packages_default_gobject_packages|select()|list()
```
tyll, line 38: `NetworkManager` got lost, here.

ffmancera: Oops. My bad. Fixing it, sorry.

library/network_connections.py
```diff
@@ -13,6 +13,10 @@
 
 # pylint: disable=import-error, no-name-in-module
 from ansible.module_utils.basic import AnsibleModule
+# pylint: disable=import-error, no-name-in-module
+from ansible.module_utils.network_lsr import ethtool
+
```
tyll, line 18: This newline should be after AnsibleModule and not here.

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,54 @@
...
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        ecmd = array.array("B", struct.pack("II6s", ETHTOOL_GPERMADDR, 6, b"\x00" * 6))
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        try:
+            res = ecmd.tobytes()
+        except AttributeError:
+            # tobytes() is not available in python2
```
tyll, line 35: Please put the comment on the except line or before it.

library/network_connections.py
ffmancera: I added it after AnsibleModule but I think it is also needed there because is separating that line from: 
```
# pylint: disable=import-error, no-name-in-module                               
from ansible.module_utils.network_lsr import MyError   
```

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,54 @@
...
+        sock.close()
+
+    return perm_addr
+
+
+def _decode_perm_addr(perm_addr_bytes):
+    if hasattr(perm_addr_bytes, "hex"):
+        perm_addr = perm_addr_bytes.hex()
+    else:
+        perm_addr = binascii.hexlify(perm_addr_bytes)
```
tyll, line 51: hexlify seems to be available in all Python versions, can't you use it all the time?

ffmancera: No I cannot use it all the time. `binascii.hexlify()` is returning a bytes representation instead of str in python3 [1]. In order to make this work I would need to parse the bytes representation into a string. This is what hex() does.

Using it all the time would generate another conflict.. so I don't think we should use it all the time.

[1] https://docs.python.org/3.6/library/binascii.html

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,54 @@
...
+    perm_addr = ":".join(f + s for f, s in zip(perm_addr[::2], perm_addr[1::2]))
```
tyll, line 52: Please consider renaming `f` and `s` to  `first_nibble` and `second_nibble`

ffmancera: Done, thanks!

module_utils/network_lsr/ethtool.py
ffmancera: Done!

tests/unit/test_network_connections.py
```diff
@@ -2503,10 +2504,13 @@ def test_path_to_glib_bytes(self):
...
 
 class TestUtils(unittest.TestCase):
-    def test_check_output(self):
-        res = Util.check_output(["echo", "test"])
-        self.assertEqual(res, "test\n")
-        self.assertRaises(n.MyError, Util.check_output, ["false"])
+    def test_decode_perm_addr(self):
+        if sys.version_info >= (3, 0):
+            mac_bytes = b"\xaa\xbb\xcc\xdd\xee\xff"
+        else:  # bytes is an alias for string type in python2
```
tyll, line 2510: I don't follow this comment. Can't you use b"..." in both python versions since this is what array.array.tostring() returns?

library/network_connections.py
tyll: You can actully just use one pylint comment for the full block and enable the check afterwards with
`# pylint: enable=import-error, no-name-in-module`

tests/unit/test_network_connections.py
ffmancera: You are right, ugh sorry for missing python2 stuff :-) Thanks!

module_utils/network_lsr/ethtool.py
tyll: I see, can you please use try/except here, too, and mention this in a comment that mentions Python 2. Thanks.

tyll: After looking at `Util.mac_ntoa()`, can't you use something like `":".join(["%02x" % byte for byte in perm_addr_bytes])` or actually decode the MAC using `Util.mac_ntoa()` in the module?

ffmancera: I cannot do that. `":".join(["%02x" % byte for byte in perm_addr_bytes])` only works for bytes representation and there is no such thing in python2. Here is the error that you get on python2 using it:

```
    @staticmethod
    def mac_ntoa(mac):
        if mac is None:
            return None
>       return ":".join(["%02x" % c for c in mac])
E       TypeError: %x format: a number is required, not str
```

ffmancera: The `mac_ntoa()` requires a byte representation as input. This function is being used in the validation because the address is being converted to byte representation before using `mac_aton()` [1]. In this case, we do have different types in python2 (str) and python3 (bytes).

[1] https://github.com/linux-system-roles/network/blob/master/module_utils/network_lsr/argument_validator.py#L302

tyll: Ok, thank you. So the problem is that `struct.unpack()` returns `str` in Python2 and `bytes` in Python3. It seems that bytearray helps here:

```
$ python3 -c 'print(":".join(["%02x" % byte for byte in bytearray(b"abcdef")]))'
61:62:63:64:65:66
$ python2 -c 'print(":".join(["%02x" % byte for byte in bytearray("abcdef")]))'
61:62:63:64:65:66
```

What do you think?

ffmancera: Oh thanks! That works :-). I've update the commit

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,47 @@
...
+        ecmd = array.array("B", struct.pack("II6s", ETHTOOL_GPERMADDR, 6, b"\x00" * 6))
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
+
+        fcntl.ioctl(sockfd, SIOCETHTOOL, ifreq)
+        try:
+            res = ecmd.tobytes()
+        except AttributeError:  # tobytes() is not available in python2
+            res = ecmd.tostring()
+        _, _, perm_addr = struct.unpack("II6s", res)
+        perm_addr = _decode_perm_addr(perm_addr)
```
tyll, line 37: You can use `Util.mac_ntoa()` directly, here. Does it make sense to add the `bytearray()` wrap direclty in `mac_ntoa()`?

tests/unit/test_network_connections.py
```diff
@@ -2503,10 +2504,9 @@ def test_path_to_glib_bytes(self):
 
 
 class TestUtils(unittest.TestCase):
-    def test_check_output(self):
-        res = Util.check_output(["echo", "test"])
-        self.assertEqual(res, "test\n")
-        self.assertRaises(n.MyError, Util.check_output, ["false"])
+    def test_decode_perm_addr(self):
+        mac_bytes = b"\xaa\xbb\xcc\xdd\xee\xff"
+        self.assertEqual(ethtool._decode_perm_addr(mac_bytes), "aa:bb:cc:dd:ee:ff")
```
tyll, line 2509: Please test `Util.mac_ntoa()` here.

module_utils/network_lsr/ethtool.py
ffmancera: Yes I think it makes sense. Done, thanks!

tests/unit/test_network_connections.py
ffmancera: Done, thanks! :-)

library/network_connections.py
```diff
@@ -107,15 +112,11 @@ def _link_read_address(ifname):
 
     @staticmethod
     def _link_read_permaddress(ifname):
-        try:
-            out = Util.check_output(["ethtool", "-P", ifname])
-        except MyError:
-            return None
+        perm_addr = ethtool.get_perm_addr(ifname)
+        if perm_addr:
+            perm_addr = Util.mac_norm(perm_addr)
```
tyll, line 117: the normalization is not needed anymore, since ethtool already returns the normalized form.

module_utils/network_lsr/utils.py
```diff
@@ -275,7 +260,7 @@ def mac_aton(mac_str, force_len=None):
     def mac_ntoa(mac):
         if mac is None:
             return None
-        return ":".join(["%02x" % c for c in mac])
+        return ":".join(["%02x" % c for c in bytearray(mac)])
```
tyll, line 263: please mention that bytearray is needed for python2 compatibility.

ffmancera: Done, thanks!

library/network_connections.py
ffmancera: Removed, thanks :-)

library/network_connections.py
ffmancera: Thanks, done!

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,43 @@
...
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L734
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L1388
+    https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c#n4172
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        ecmd = array.array("B", struct.pack("II6s", ETHTOOL_GPERMADDR, 6, b"\x00" * 6))
+        ifreq = struct.pack("16sP", ifname, ecmd.buffer_info()[0])
```
tyll, line 29: struct will only use 16 bytes from ifname but ifname might be bigger in theory. Then it might return the wrong MAC address. How about checking the length of ifname after encoding and returning None if it is too long?

module_utils/network_lsr/utils.py
```diff
@@ -25,19 +23,6 @@ def first(iterable, default=None, pred=None):
                 return v
         return default
 
-    @staticmethod
```
tyll, line 28: awesome, that we can remove this!

tests/unit/test_network_connections.py
```diff
@@ -2528,10 +2527,7 @@ def test_convert_passwd_flags_nm(self):
 
 class TestSysUtils(unittest.TestCase):
     def test_link_read_permaddress(self):
-        # Manipulate PATH to use ethtool mock script to avoid hard dependency on
-        # ethtool
-        os.environ["PATH"] = TESTS_BASEDIR + "/helpers:" + os.environ["PATH"]
-        self.assertEqual(SysUtil._link_read_permaddress("lo"), "23:00:00:00:00:00")
+        self.assertEqual(SysUtil._link_read_permaddress("lo"), "00:00:00:00:00:00")
```
tyll, line 2530: Please add tests for a non-existing interface and an invalid interface (more than 16 characters), too.

ffmancera: Done, thanks!

module_utils/network_lsr/utils.py
ffmancera: Yes!

module_utils/network_lsr/ethtool.py
ffmancera: Done thanks!

module_utils/network_lsr/ethtool.py
```diff
@@ -0,0 +1,46 @@
...
+    Please for further documentation, see:
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L734
+    https://github.com/torvalds/linux/blob/master/include/uapi/linux/ethtool.h#L1388
+    https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c#n4172
+    """
+    try:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sockfd = sock.fileno()
+        ifname = ifname.encode("utf-8")
+        if len(ifname) > 16:
```
tyll, line 28: Please define a constant for this `IFNAMSZ` and use it here and for the struct format strings. Please also use one for the perm address size and check that this will also work for infiniband devices that have a longer address. Thanks.

ffmancera: Done, thanks! It will work with infiniband devices.

ffmancera:
> Awesome, thank you!

Thanks for all the advices :-)

Issue #229, opened 2020-05-13
RFE: 802.1x support ca-path setting


tyll:
Add support for the ca-path NM setting:
https://developer.gnome.org/NetworkManager/stable/nm-settings.html#nm-settings.property.802-1x.ca-path

tyll:
@jaedolph are you interested in doing this?

jaedolph:
@tyll sure thing, I'll have a look at it this weekend

jaedolph:
Have implement this in a new PR, but the actual ca-path setting itself doesn't appear to work in NetworkManager. Have raised a bug here: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/448 

@thom311 do you have any ideas?

PR #232, opened 2020-05-17
Add ca_path option for 802.1x connections

jaedolph:
Addresses issue #229


library/network_connections.py
```diff
@@ -1004,6 +1004,11 @@ def connection_create(self, connections, idx, connection_current=None):
                     Util.path_to_glib_bytes(connection["ieee802_1x"]["ca_cert"]),
                 )
 
+            if connection["ieee802_1x"]["ca_path"]:
+                s_8021x.set_property(
+                    NM.SETTING_802_1X_CA_PATH, connection["ieee802_1x"]["ca_path"],
```
jaedolph, line 1009: As mentioned here:  https://github.com/linux-system-roles/network/issues/229#issuecomment-629760274 NetworkManager doesn't actually set this property properly. Fairly sure this is a bug.

tests/playbooks/tests_802_1x.yml
```diff
@@ -89,6 +89,50 @@
...
           shell: ping -c1 203.0.113.1
+        - import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: "{{ interface }}"
+                persistent_state: absent
+                state: absent
+        - name: >-
+            TEST: 802.1x profile with unencrypted private key and ca_path
```
jaedolph, line 100: The only reason this test passes, is because NetworkManager will not do any certificate validation by default

tyll:
[citest bad]

tests/playbooks/tests_802_1x.yml
```diff
@@ -89,6 +89,50 @@
...
+            msg: "##################################################"
+        - name: Create directory for ca_path test
+          file:
+            path: "/etc/pki/tls/my_ca_certs"
+            state: directory
+            mode: 0755
+        - name: Copy cacert to ca_path
+          copy:
+            src: "cacert.pem"
+            dest: "/etc/pki/tls/my_ca_certs/cacert.pem"
```
tyll, line 111: You need to run `ln -s "cacert.pem" "$(openssl x509 -hash -noout -in "cacert.pem")".0` in that directory to hash the file - it is a one-liner to avoid installing c_rehash which requires perl, see: https://stackoverflow.com/questions/25889341/what-is-the-equivalent-of-unix-c-rehash-command-script-on-linux

tyll:
I have some changes for this locally that I need to commit properly to fix the certificate hashing and add a check for NM support. Will try to publish them tomorrow.

library/network_connections.py
tyll: NetworkManager got fixed upstream. I adjusted the module/tests to check of support and fail if it is requested but missing.

library/network_connections.py
```diff
@@ -2084,6 +2089,25 @@ def run_action_present(self, idx):
                 % (con_cur.get_id(), con_cur.get_uuid()),
             )
 
+        if (
```
tyll, line 2092: @thom311 please take a look it check if this is the right way to figure out if `ca_path` is working with NetworkManager.

tyll:
@jaedolph I adjusted your commit. It would be great if you could take a look.

tests/playbooks/tests_802_1x.yml
```diff
@@ -3,6 +3,18 @@
 - hosts: all
   vars:
     interface: 802-1x-test
+    # Fixed versions/NVRs:
+    # 1.25.2
+    # NetworkManager-1.24.2-1.fc33
+    # NetworkManager-1.22.14-1.fc32
+    # NetworkManager-1.20.12-1.fc31
+    # 1.18.8
+    __NM_capath_ignored_NVRs:
```
jaedolph, line 12: Should this be prefixed with __network too?

tyll: I considered this but then it becomes harder to keep the line length. Since it is a variable used in a test and not exposed directly from the role, I don't think it needs to use the prefix.

tests/playbooks/tests_802_1x.yml
```diff
@@ -88,7 +100,80 @@
...
+                    ieee802_1x:
+                      identity: myhost_capath
+                      eap: tls
+                      private_key: /etc/pki/tls/client.key.nocrypt
+                      client_cert: /etc/pki/tls/client.pem
+                      private_key_password_flags:
+                        - not-required
+                      ca_path: /etc/pki/tls/my_ca_certs
+            - fail:
+                name: Check if role should have failed
```
jaedolph, line 162: The ansible logs show this error:
```
TASK [fail] ********************************************************************
task path: /tmp/tmpf1of1lym/tests/playbooks/tests_802_1x.yml:161
fatal: [/cache/CentOS-7-x86_64-GenericCloud-1907.qcow2c]: FAILED! => {"changed": false, "msg": "Invalid options for fail: name"}
```
From the ansible doco it looks like the "name" option is not supported https://docs.ansible.com/ansible/latest/modules/fail_module.html

Also I don't think this will properly check if the ca_path test silently fails. Pretty sure if this "fail" tasks runs, the playbook will still run the "rescue" section. And because the "fail" task in the rescue section has opposite conditions to the "fail" task in the "block" it will never report a failure for the "ca_path test did not fail but {{ __network_NM_NVR.stdout }} known to be broken" case

tyll: yes, good catch, thanks!

tyll: I will have to add something like discussed here: https://github.com/linux-system-roles/network/pull/200

tests/playbooks/tests_802_1x.yml
```diff
@@ -88,7 +100,89 @@
...
+                      private_key: /etc/pki/tls/client.key.nocrypt
+                      client_cert: /etc/pki/tls/client.pem
+                      private_key_password_flags:
+                        - not-required
+                      ca_path: /etc/pki/tls/my_ca_certs
+            - name: trigger failure in case the role did not fail
+              fail:
+                msg: after test
+              when:
+                - __network_NM_NVR.stdout in __NM_capath_ignored_NVRs
```
pcahyna, line 165: can't you omit this condition? The possibility that `__network_NM_NVR.stdout` is not in `__NM_capath_ignored_NVRs` is handled in the rescue block below, isn't it? So you could fail here always. (You would have to move the ping task below to the rescue block.)

tests/playbooks/tests_802_1x.yml
```diff
@@ -88,7 +100,89 @@
...
+                msg: "ca_path test did not fail but
+                    {{ __network_NM_NVR.stdout }} known to be broken.
+                    Investigate or update the versions."
+              when:
+                - __network_NM_NVR.stdout in __NM_capath_ignored_NVRs
+                - ansible_failed_result.msg == 'after test'
+            - name: Check if role failed unexpectedly
+              assert:
+                that:
+                  - ansible_failed_result.msg != 'after test'
```
pcahyna, line 180: a list entry marker `- ` seems to be missing on the line below. Or maybe you meant `fail_msg` (an option for `assert`) instead of
```yaml
fail:
  msg:
```
(a separate `fail` task).

tyll: I think I meant to remove the fail and use the msg for the assert.

tests/playbooks/tests_802_1x.yml
tyll: Moving the test to the rescue block seems to be counter-intuitive. How about:

```
block:
    import_role
    set_fact: role_success
    ping
rescue:
    ignore when expected failure
always:
    assert that not role_success when expected failure
```

This seems to be a little bit more straight-forward.

I guess we could skip the rescue block if we can express this in Ansible:

```
assert that (role_success | default False) == expected failure
```

`expected failure` is `__network_NM_NVR.stdout in __NM_capath_ignored_NVRs` 

so maybe this is the solution. I guess instead of using set_fact it might be possible to register `role_result` and check this.

tyll: This seems to be working:

```yaml
---
- hosts: localhost
  vars:
    fail_trigger: false
    expected_failure: false
  tasks:
    - block:
        - assert:  # this would be the import_role
            that: not fail_trigger 
          register: role_result
        - debug:  # this would be the ping
            var: role_result
      always:
        - assert:
            that: role_result.failed == expected_failure  # expected_failure would be __network_NM_NVR.stdout in __NM_capath_ignored_NVRs
            msg: "Role {{ role_result.failed and 'failed' or 'did not fail' }}
                but was expected
                {{ expected_failure and 'to fail' or 'not to fail' }}"
```

pcahyna: If moving the test (I assume you mean the `command: ping` test) to the rescue block seems to be counter-intuitive, can you keep the `ping` in the `block` and move it between the `import_role:` and the `fail:`? The logic is that either the role fails (maybe that's expected, this will be checked in `rescue`) and the `ping` will be skipped, or it succeeds, and in this case the `ping` should definitely succeed, so it can be safely executed just after the role invocation.
(It may be needed to check in `rescue:` if the failed task was not `ping` by looking at for example `ansible_failed_result.cmd`: https://docs.ansible.com/ansible/latest/modules/command_module.html#return-cmd )

pcahyna: > This is my new approach:

Weird. Do I understand correctly that you get a proper failure in `always:` but despite this failure the playbook reports a success i.e. the failed status is cleared somewhere?? If so, this means that `always` does not behaves as expected.

tyll: > > This is my new approach:
> 
> Weird. Do I understand correctly that you get a proper failure in `always:` but despite this failure the playbook reports a success i.e. the failed status is cleared somewhere?? If so, this means that `always` does not behaves as expected.

Yes, this is what I see, too.

Also, the approach has another problem, it will mask any failure after the `set_fact`.

pcahyna: @tyll I propose this:
```yaml
        - block:
            - import_role:
              vars:
                network_connections:
                  - name: "{{ interface }}"
                    interface_name: veth2
                    state: up
                    type: ethernet
                    ip:
                      address:
                        - 203.0.113.2/24
                      dhcp4: "no"
                      auto6: "no"
                    ieee802_1x:
                      identity: myhost_capath
                      eap: tls
                      private_key: /etc/pki/tls/client.key.nocrypt
                      client_cert: /etc/pki/tls/client.pem
                      private_key_password_flags:
                        - not-required
                      ca_path: /etc/pki/tls/my_ca_certs

            - name: "TEST: I can ping the EAP server"
              command: ping -c1 203.0.113.1

            - name: trigger failure in case the role did not fail
              fail:
                msg: after test
              when:
                - __network_NM_NVR.stdout in __NM_capath_ignored_NVRs

          rescue:
            - name: Check ping result
              fail:
                msg: "ping failed, stdout:
                  {{ ansible_failed_result.stdout }}, stderr:
                  {{ ansible_failed_result.stderr }}"
              when:
                - "'cmd' in ansible_failed_result"
                - ansible_failed_result.cmd[0] == 'ping'
            - name: Check that role failed as expected
              fail:
                msg: "ca_path test did not fail but
                    {{ __network_NM_NVR.stdout }} known to be broken.
                    Investigate or update the versions."
              when:
                - __network_NM_NVR.stdout in __NM_capath_ignored_NVRs
                - ansible_failed_result.msg == 'after test'
            - name: Check if role failed unexpectedly
              fail:
                msg: "ca_path test failed but {{ __network_NM_NVR.stdout }}
                    not known to be broken. Investigate or add the version."
              when:
                - __network_NM_NVR.stdout not in __NM_capath_ignored_NVRs
                - ansible_failed_result.msg != 'after test'
```

tyll: FYI: I filed https://github.com/ansible/ansible/issues/70000 about the failure in always not being reported by ansible-playbook's returncode.

tests/playbooks/tests_802_1x.yml
```diff
@@ -88,7 +100,90 @@
...
+                      client_cert: /etc/pki/tls/client.pem
+                      private_key_password_flags:
+                        - not-required
+                      ca_path: /etc/pki/tls/my_ca_certs
+            - name: "TEST: I can ping the EAP server"
+              command: ping -c1 203.0.113.1
+            - name: trigger failure in case the role did not fail
+              fail:
+                msg: after test
+              when:
```
pcahyna, line 166: I think this when: is unneeded, because you check this condition in rescue:. And it is better to have this fail: unconditional, so that the checks in rescue: are guaranteed to always execute.

tests/tasks/test_802.1x_capath.yml
```diff
@@ -0,0 +1,94 @@
...
+      with_items:
+        - ansible_failed_result
+        - ansible_failed_task
+        - __network_NM_NVR.stdout
+        - __NM_capath_ignored_NVRs
+
+    - name: Assert role behavior
+      assert:
+        that: __network_connections_result.failed ==
+          (__network_NM_NVR.stdout in __NM_capath_ignored_NVRs)
```
pcahyna, line 84: don't compare booleans, this was fixed only very recently: https://github.com/pallets/jinja/issues/813

pcahyna: in 2.11, only half a year ago, we have to support much older Jinjas https://jinja.palletsprojects.com/en/2.11.x/changelog/#version-2-11-0

PR #233, opened 2020-05-17
Prefix all local role variables with '__network'

jaedolph:
To avoid conflicts with other roles, it is recomended to prefix all variables
that are only used internally with '__' and the name of the role ('__network_').

See discussion here: https://github.com/linux-system-roles/network/pull/157#discussion_r420965924


pcahyna:
Just a remark, the `network_` part is used to avoid conflicts with anything else, the leading underscores to mark the variables as internal.

Issue #205, opened 2020-04-22
Warn or fail if `persistent_state` is present and state is set to present or absent.


tyll:
If both `persistent_state` and `state` are set for a profile, allowing state should not be specified or `up` or `down`. Setting it to `absent` or `present` would be a mistake and indicates a problem. It happened in the past, see: https://github.com/linux-system-roles/network/pull/196/commits/fbf79ac3a0e81e5a94bc5ef3d649e741edba1158

elvgarrui:
I'm willing to complete this but I have a question: Do you mean that if persistent_state is set then state can only be up or down? I'm a bit confused by the message issue.

tyll:
> I'm willing to complete this but I have a question: Do you mean that if persistent_state is set then state can only be up or down? I'm a bit confused by the message issue.

Basicaslly, yes: if persistent_state is set then state can only be up, down or not be specified.

elvgarrui:
Should we close this issue now?

tyll:
yes, thanks

PR #237, opened 2020-05-26
Fail if state and persistent_state are incompatible

elvgarrui:
When persistent_state is present and state is set to present or absent, a
ValidationError raises. A unit test validating this has been added.

Ref #205 

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -1064,14 +1064,18 @@ class ArgValidator_DictConnection(ArgValidatorDict):
         """
         actions = []
         state = result.get("state")
-        if state in self.VALID_PERSISTENT_STATES:
-            del result["state"]
-            persistent_state_default = state
-            state = None
-        else:
-            persistent_state_default = None
+        persistent_state = result.get("persistent_state")
 
-        persistent_state = result.get("persistent_state", persistent_state_default)
+        if state in self.VALID_PERSISTENT_STATES:
+            if persistent_state:
+                raise ValidationError(
+                    name,
+                    "State cannot be '{0}' if persistent_state is specified".format(
+                        state
+                    ),
+                )
+            persistent_state = state
+            state = None
 
         # default persistent_state to present (not done via default_value in the
         # ArgValidatorStr, the value will only be set at the end of
--- a/tests/playbooks/tests_802_1x.yml
+++ b/tests/playbooks/tests_802_1x.yml
@@ -42,7 +42,7 @@
             network_connections:
               - name: "{{ interface }}"
                 persistent_state: absent
-                state: absent
+                state: down
         - name: >-
             TEST: 802.1x profile with unencrypted private key,
             domain suffix match, and system ca certs
@@ -85,7 +85,7 @@
             network_connections:
               - name: "{{ interface }}"
                 persistent_state: absent
-                state: absent
+                state: down
 
         - include_tasks: tasks/test_802.1x_capath.yml
       always:
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -2986,6 +2986,22 @@ class TestValidator(unittest.TestCase):
             ):
                 assert validator.deprecated_by in validators.keys()
 
+    def test_valid_persistent_state(self):
+        """
+        Test that when persistent_state is present and state is set to present
+        or absent, a ValidationError raises.
+        """
+        validator = network_lsr.argument_validator.ArgValidator_DictConnection()
+        input_connection = {
+            "name": "test",
+            "persistent_state": "present",
+            "state": "present",
+            "type": "ethernet",
+        }
+        self.assertValidationError(validator, input_connection)
+        input_connection.update({"state": "absent"})
+        self.assertValidationError(validator, input_connection)
+
 
 @my_test_skipIf(nmutil is None, "no support for NM (libnm via pygobject)")
 class TestNM(unittest.TestCase):


module_utils/network_lsr/argument_validator.py
```diff
@@ -896,6 +896,14 @@ def _validate_post_fields(self, value, name, result):
         state = result.get("state")
         persistent_state = result.get("persistent_state")
 
+        # if persistent state is present state cannot be set to present or absent
+        if persistent_state == "present" and state:
```
tyll, line 900: Please check if persistent state is present or absent. Also, this can be one if statement like:

`if persistent_state in ... and state in ...:`

module_utils/network_lsr/argument_validator.py
```diff
@@ -896,6 +896,14 @@ def _validate_post_fields(self, value, name, result):
...
+            if state in ("absent", "present"):
+                raise ValidationError(
+                    name,
+                    "State cannot be '{}' if persistent_state is present".format(state),
```
tyll, line 904: How about making this postive like "state can only be up or down if persistent_state is ....". Also, using `.format()` with  `{}` is not Python 2.6 compatible if I recall correctly.

tyll: Not sure, why this is not failing in the unit test...

elvgarrui: Fixed, thanks!

module_utils/network_lsr/argument_validator.py
elvgarrui: Fixed, thanks!

module_utils/network_lsr/argument_validator.py
```diff
@@ -896,6 +896,16 @@ def _validate_post_fields(self, value, name, result):
         state = result.get("state")
         persistent_state = result.get("persistent_state")
 
+        # if persistent state is present state cannot be set to present or absent
+        if persistent_state in result.keys() and state in result.keys():
```
tyll, line 900: This does not seem to be right, on first glance it seems you need to check wether persistent_state is not `None`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -896,6 +896,16 @@ def _validate_post_fields(self, value, name, result):
...
+            if state in ("absent", "present"):
+                raise ValidationError(
+                    name,
+                    "State cannot be '{0}' if persistent_state is present".format(
```
tyll, line 904:
```suggestion
                    "State cannot be '{0}' if persistent_state is specified".format(
```

elvgarrui: Fixed, thanks!

elvgarrui:
@tyll I've fixed the code and the unit test. Thanks for your reviews!!

elvgarrui:
The integration test `playbooks/test_802_1x.yml` is failing after the update. I think the problem is in line 87:88, because `persistent_state` is specified but `state` is absent. I've tried erasing the configuration of `persistent_state` (line 87), and it still failed. Same when erasing  `state` (line 88). I'm not sure anymore if the failure is generated because of a bad configuration of the test or if my commit was wrong.

tyll:
> The integration test `playbooks/test_802_1x.yml` is failing after the update. I think the problem is in line 87:88, because `persistent_state` is specified but `state` is absent. I've tried erasing the configuration of `persistent_state` (line 87), and it still failed. Same when erasing `state` (line 88). I'm not sure anymore if the failure is generated because of a bad configuration of the test or if my commit was wrong.

It seems to me that `state` should be `down` there if it is specified as `absent`. There are multiple occurrences in that file.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1064,6 +1064,14 @@ def _validate_post_state(self, value, name, result):
         """
         actions = []
         state = result.get("state")
+        persistent_state = result.get("persistent_state")
+        # if persistent state is present state cannot be set to present or absent
```
tyll, line 1068:
```suggestion
```

The comment is not needed because it is clear from the error message

module_utils/network_lsr/argument_validator.py
```diff
@@ -1064,6 +1064,14 @@ def _validate_post_state(self, value, name, result):
...
+        if persistent_state and state in ("absent", "present"):
```
tyll, line 1069:
```suggestion
        if persistent_state and state in self.VALID_PERSISTENT_STATES:
```
Seems like there is a constant that you can use, just saw it after the diff.

elvgarrui:
[citest bad]

module_utils/network_lsr/argument_validator.py
```diff
@@ -1064,6 +1064,13 @@ def _validate_post_state(self, value, name, result):
...
+        persistent_state = result.get("persistent_state")
+        if persistent_state and state in self.VALID_PERSISTENT_STATES:
+            raise ValidationError(
+                name,
+                "State cannot be '{0}' if persistent_state is specified".format(state),
+            )
+
         if state in self.VALID_PERSISTENT_STATES:
             del result["state"]
             persistent_state_default = state
```
tyll, line 1076:
```suggestion
        persistent_state = result.get("persistent_state")
        if state in self.VALID_PERSISTENT_STATES:
            if persistent_state:
                raise ValidationError(
￼                name,
￼                "State cannot be '{0}' if persistent_state is specified".format(state),
￼            )
            
            del result["state"]
```

What do you think? Everything until `# default persistent_state to present (not done via default_value in the` can probably be removed.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1064,6 +1064,13 @@ def _validate_post_state(self, value, name, result):
         """
         actions = []
         state = result.get("state")
+        persistent_state = result.get("persistent_state")
```
tyll, line 1067: Looking at the lines following this addition, I think this code should be merged since one condition is repeated here. And `persistent_state_default` does not make sense to have anymore. See my other comment for details.

module_utils/network_lsr/argument_validator.py
elvgarrui: @tyll: I also think this makes much more sense now. Anyway, I think is better to just set state to None after giving its value to persistent state (instead of using `del result["state"]`), since the `result` dict is modified in lines 1103:1105 with the rest of the properties, improving readability. (You can see it on the updated commit)

PR #241, opened 2020-06-04
Test bond with two interfaces

tyll:
@vbenes I created the PR for you to make it easier to comment


tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    # Setup simulated device test1 and test2 with IPv6 auto support
+    ip link add test1 type veth peer name test1p
+    ip link set test1p netns testsetup
+    ip netns exec testsetup ip link set test1p up
+    ip link add test2 type veth peer name test2p
+    ip link set test2p netns testsetup
+    ip netns exec testsetup ip link set test2p up
+
+    # Create the 'testbr' - providing both 10.x ipv4 and 2620:52:0 ipv6 dhcp
```
tyll, line 27: Please use RFC5737 addresses: https://tools.ietf.org/html/rfc5737

tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
...
+    ip netns exec testsetup ip link set testbr up
+    ip netns exec testsetup ip addr add 10.0.0.1/24 dev testbr
+    ip netns exec testsetup ip -6 addr add 2620:52::1/64 dev testbr
+
+    # Add test1, test2 peers into the testbr
+    ip netns exec testsetup ip link set test1p master testbr
+    ip netns exec testsetup ip link set test2p master testbr
+
+    # Run joint DHCP4/DHCP6 server with router advertisement enabled in veth namespace
+    ip netns exec testsetup dnsmasq --pid-file=/tmp/dhcp_testbr.pid \
```
tyll, line 38: Please use a directory in `/run` since the test is expected to run as root so we do not need to use `/tmp`.

tests/tests_bond_nm.yml
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
tyll, line 1: The test can also run with initiscripts since bond is supported there, too. Running `tests/ensure_provider_tests.py generate` will create the shims for this.

tests/playbooks/tests_bond.yml
```diff
@@ -0,0 +1,81 @@
...
+          vars:
+            profile: "{{ item }}"
+          loop:
+            - "{{ master_profile }}"
+            - "{{ slave1_profile }}"
+            - "{{ slave2_profile }}"
+
+      always:
+        - block:
+            - import_role:
```
tyll, line 64: The cleanup could be simplified by running the module instead importing the role:
https://github.com/linux-system-roles/network/blob/master/tests/playbooks/tests_checkpoint_cleanup.yml#L70

tests/playbooks/tests_bond.yml
```diff
@@ -0,0 +1,81 @@
...
+        - import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              # Create a bond master
+              - name: "{{ master_profile }}"
+                state: up
+                type: bond
+                interface_name: "{{ master_device }}"
+                # bond configuration settings: (optional)
```
tyll, line 35:
```suggestion
```

tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
...
+    name: dnsmasq
+    state: present
+
+
+- name: Create test interfaces
+  shell: |
+    # Manage all veth devices starting with test*
+    echo 'ENV{ID_NET_DRIVER}=="veth", ENV{INTERFACE}=="test*", ENV{NM_UNMANAGED}="0"' >/etc/udev/rules.d/88-veth.rules
+    udevadm control --reload-rules
+    udevadm control --reload-rules
```
tyll, line 14: The other tests use veth as well but don't need this

tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
tyll, line 1: Maybe call this file `create_test_interfaces_with_dhcp_server.yml` to indicate that it will setup dhcp, too.

tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
...
+    # Manage all veth devices starting with test*
+    echo 'ENV{ID_NET_DRIVER}=="veth", ENV{INTERFACE}=="test*", ENV{NM_UNMANAGED}="0"' >/etc/udev/rules.d/88-veth.rules
+    udevadm control --reload-rules
+    udevadm control --reload-rules
+
+    # Create a namespace for devices
+    ip netns add testsetup
+
+    # Setup simulated device test1 and test2 with IPv6 auto support
+    ip link add test1 type veth peer name test1p
```
tyll, line 20:
```suggestion
    ip link add {{dhcp_interface1}} type veth peer name {{dhcp_interface1}}p
```
Then you need to use `dhcp_interface1` instead of `slave1_device` in the vars and would not need to hardcode the interface name.

tests/tasks/create_test_interfaces.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Create test interfaces
+  shell: |
+    # Manage all veth devices starting with test*
+    echo 'ENV{ID_NET_DRIVER}=="veth", ENV{INTERFACE}=="test*", ENV{NM_UNMANAGED}="0"' >/etc/udev/rules.d/88-veth.rules
+    udevadm control --reload-rules
+    udevadm control --reload-rules
+
+    # Create a namespace for devices
+    ip netns add testsetup
```
tyll, line 17:
```suggestion
    ip netns add lsrdhcp
```

To avoid conflicts with other tests.

tests/tasks/remove_test_interfaces.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Remove test interfaces
+  shell: |
+    ip netns delete testsetup
+    ip link delete test1
+    ip link delete test2
+
+- name: Kill hostapd process
```
tyll, line 9:
```suggestion
- name: Kill dnsmasq process
```

tests/tasks/create_test_interfaces.yml
tyll: Another thought, since we need to use dhcp with hostapd as well, can we maybe re-use this dhcp setup there, to avoid duplication?

tests/tasks/create_test_interfaces.yml
vbenes: ok, 192.0.2.1 and 2001:DB8::1 now used

tests/tasks/create_test_interfaces.yml
vbenes: /run used

tests/tests_bond_nm.yml
vbenes: will cover this later

tests/playbooks/tests_bond.yml
vbenes: removed

tests/tasks/create_test_interfaces.yml
vbenes: this is needed for NM to see test* device as managed directly after calling ip add, nevertheless removing this was missing and double udevadm control was incorrect too
    udevadm control --reload-rules
    udevadm settle --timeout=5

tests/playbooks/tests_bond.yml
vbenes: am I able to remove namespace?

tests/tasks/create_test_interfaces.yml
vbenes: ok, using this as you suggested

tests/playbooks/tests_bond.yml
vbenes: ok, I am, but when I am creating such a setup I think there should be cleanup too so one doesn't need to copy paste 20 lines of code. 
https://docs.ansible.com/ansible/latest/modules/ip_netns_module.html

tests/tasks/create_test_interfaces.yml
vbenes: done

tests/tasks/remove_test_interfaces.yml
vbenes: done

tests/playbooks/tests_bond.yml
```diff
@@ -0,0 +1,95 @@
...
+                    state: down
+                  - name: "{{ slave1_profile }}"
+                    persistent_state: absent
+                    state: down
+                  - name: "{{ master_profile }}"
+                    persistent_state: absent
+                    state: down
+
+              ignore_errors: true
+            - include_tasks: tasks/remove_test_interfaces_with_dhcp.yml
```
pcahyna, line 93: I suggest to use `import_tasks` here, it is reported to be faster and also allows more static analysis.

tests/playbooks/tests_bond.yml
```diff
@@ -0,0 +1,95 @@
...
+                  - name: "{{ slave2_profile }}"
+                    persistent_state: absent
+                    state: down
+                  - name: "{{ slave1_profile }}"
+                    persistent_state: absent
+                    state: down
+                  - name: "{{ master_profile }}"
+                    persistent_state: absent
+                    state: down
+
```
pcahyna, line 91: I think the blank line should be removed or moved after `ignore_errors: true`. The way how it is now it may visually seem that `ignore_errors: true` belongs to the task below, not the one above.

tests/playbooks/tests_bond.yml
vbenes: added where possible

tests/playbooks/tests_bond.yml
vbenes: blank line removed

tests/playbooks/tests_bond.yml
vbenes: done

tests/tasks/create_test_interfaces_with_dhcp.yml
```diff
@@ -28,9 +28,10 @@
     ip -6 addr add 2001:DB8::1/32 dev testbr
 
     if grep 'release 6' /etc/redhat-release; then
-        # We need radvd only in rhel6
+        # We need bridge-utils and radvd only in rhel6
         if ! rpm -q --quiet radvd; then yum -y install radvd; fi
-        
+        if ! rpm -q --quiet bridge-utils; then yum -y install bridge-utils; fi
```
pcahyna, line 33: if you install stuff in the test setup, then you will not properly test whether the role installs the package itself.
Also, isn't it better to use the `package` module for this instead of shell commands, conditions, etc.?

vbenes: I need to install it just for RHEL6, not the others, how should I do that?

vbenes: I need setup to be deployed to mimic complex network setup.. few ethernets with ipv4/ipv6 addresses and dhcp.. I don't want to use eth0 for this for various reasons.

tests/playbooks/tests_bond.yml
```diff
@@ -0,0 +1,110 @@
...
+              # rhbz1508614. we need to up after all profiles are created
+              # state: up
+                type: bond
+                interface_name: "{{ master_device }}"
+                bond:
+                  mode: active-backup
+                  miimon: 110
+              # enslave an ethernet to the bond
+              - name: "{{ slave1_profile }}"
+              # rhbz1508614. we need to up after all profiles are created
```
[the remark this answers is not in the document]
tyll: Also, is it really necessary to do first all present and then all up events? There might have been a misunderstanding. This is the order the role might do it to fix this but specifying the member profiles first or last might actually solve this, too.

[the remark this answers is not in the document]
tyll: #250 is an attempt to remove this restriction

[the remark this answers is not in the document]
tyll: I merged #250, can you please adjust the test?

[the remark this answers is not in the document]
vbenes: done, but it looks now that something is triggering cleanup (always) phase and profiles got removed before profile and ipv4/6 checks.

tyll:
@vbenes please squash the commits and rebase so that we get a fresh report from the CI. Thanks!

tests/tasks/create_test_interfaces_with_dhcp.yml
tyll: This would allow to install packages depending on the distribution version:

```yaml
- name: Install dependencies
  package:
    name:
      - bridge-utils
      - radvd
 when:
   - ansible_distribution_major_version == '6'
```

tyll:
Rebased in #268

PR #250, opened 2020-06-17
Support creating and activating bond profiles in one run with initscripts

### LINKED ISSUE #242: bond creation from example does not work with initscipts
I've taken the example from https://github.com/linux-system-roles/network/blob/master/examples/bond_simple.yml

modified it to eth0, or any other device present in the system and executed it via 
ansible-playbook -vv -i localhost, bond_simple.yml

but it failed with network.service.

it looks slave profile is never created, as if I do so I can up bond via ifup

slave ifcfg
TYPE=Ethernet
DEVICE=eth0
ONBOOT=yes
MASTER=bond0
SLAVE=yes

tyll:
This fixes #242

### DIFF
--- a/examples/bond_simple.yml
+++ b/examples/bond_simple.yml
@@ -3,31 +3,34 @@
 - hosts: network-test
   vars:
     network_connections:
-
-      # Create a bond master
+      # Specify the bond profile
       - name: bond0
         state: up
         type: bond
         interface_name: bond0
+        # ip configuration (optional)
+        ip:
+          address:
+            - "192.0.2.24/24"
+            - "2001:db8::23/64"
         # bond configuration settings: (optional)
         bond:
           mode: active-backup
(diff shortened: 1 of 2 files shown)


pcahyna:
Will this allow to specify bond masters/slaves in an arbitrary order in `network_connections`? Right now a particular (and I believe undocumented) order has been needed, otherwise the role failed.

tyll:
> Will this allow to specify bond masters/slaves in an arbitrary order in `network_connections`? Right now a particular (and I believe undocumented) order has been needed, otherwise the role failed.

The profiles that are referenced in other profiles still need to be specified first.

pcahyna:
This PR itself does not touch any tests. Does #241 add tests for it?

tyll:
> This PR itself does not touch any tests. Does #241 add tests for it?

yes

PR #252, opened 2020-06-25
States: Ignore already removed profile for absent

### LINKED ISSUE #188: Support state:down persistent_state:absent for non-existent profile

tyll:
When a profile is specified as absent, ignore state requests if the
profile is already removed or not completely specified to improve
idempotence.

Also restructure the states test. This introduces a clear structure for 
the individual test steps, properly assigns tags for each test from the 
file and provides a clear error message with a description of the test
that failed in case of errors. Support for tests that expect a failure
is still missing.

This fixes #188


tests/playbooks/tests_states.yml
```diff
@@ -4,74 +4,134 @@
...
+            lsr_cleanup:
+              - tasks/cleanup_profile+device.yml
+      tags:
+        - tests::states:remove_down
 
-      always:
-        - assert:
-            that: __network_test_failed
-            fail_msg: "The role did not fail when it should have"
+    - block:
```
pcahyna, line 109: This whole blocks seems needlessly repetitive. As the profile was already removed and downed in the previous block, can't you simply omit all `lsr_setup:` (the system is already set up properly) and keep only `lsr_test:` and `lsr_assert:`?

tests/playbooks/tests_states.yml
```diff
@@ -4,74 +4,134 @@
...
-          set_fact:
-            __network_test_failed: false
+    - block:
+        - include_tasks: tasks/run_test.yml
+          vars:
+            lsr_description: I can remove an existing profile without taking it
+              down
+            lsr_setup:
+              - tasks/create_bridge_profile.yml
+              - tasks/activate_profile.yml
```
pcahyna, line 78: I think you can remove the `lsr_setup` above and `lsr_cleanup` in the previous block to chain the tests together.

tests/playbooks/tests_states.yml
```diff
@@ -4,74 +4,134 @@
...
-          set_fact:
-            __network_test_failed: true
+    - block:
+        - include_tasks: tasks/run_test.yml
+          vars:
+            lsr_description: I can take a profile down that is absent
+            lsr_setup:
+              - tasks/create_bridge_profile.yml
+              - tasks/activate_profile.yml
+              - tasks/remove_profile.yml
```
pcahyna, line 96: I think that here, too, you can remove the `lsr_setup` above and `lsr_cleanup` in the previous block to chain the tests together.

tests/playbooks/tests_states.yml
tyll: The repetition is there to make the tests independent from each other to allow to select only one of them with "--tags tests::states:remove_down_twice". I guess a possibility might be to actually join test/assert and have only the "twice" test but since the output from one role invocation is very much, it will then create a lot of logs for just one test.

pcahyna:
Can you please split e86d95f6bcc9fdc7d0b9676d3cb72ec3c53ab5c5 into functional change and test change? It would be interesting to cherry-pick the test change without the functional change in a scratch PR (or do a step-wise update of this PR) to verify that the test properly fails for the previous version.

tests/playbooks/tests_states.yml
pcahyna: I remember that we did not like Molecule exactly because it did not allow this chaining of tests where the state created by one test was used as setup for another test. If this is not the your priority anymore, we could reconsider the decision to avoid Molecule (ok, there are also other problems).
Question is, whether the ability of selecting just one test via tags is worth the complication (and the increased run time due to repetitive role invocations). I am not sure myself here. So far, we have apparently not been missing the ability to select a subset of a test playbook, but maybe your experience is different.

tyll:
> Can you please split [e86d95f](https://github.com/linux-system-roles/network/commit/e86d95f6bcc9fdc7d0b9676d3cb72ec3c53ab5c5) into functional change and test change? It would be interesting to cherry-pick the test change without the functional change in a scratch PR (or do a step-wise update of this PR) to verify that the test properly fails for the previous version.

I filed #256 that reverts the changes in the module.

pcahyna:
Thanks. Generally speaking, is there a defined procedure/workflow for testing that the tests properly capture the problem? I used to open such scratch PRs with tests only, but not consistently. An example of such workflow would be:
- commit changes to tests, push to branch, create PR
- let the CI fail
- in the meantime, commit the code changes
- when CI fails, push the code changes to the branch and watch the re-triggered CI succeed. The failure of the previous CI run will be recorded for posterity with the commit (but only until you rebase).

tests/playbooks/tests_states.yml
tyll: When I add or debug a test, I find it quite useful to be able to select that specific test since running all the tests takes up a lot of time. If it takes longer for the tests to run in the CI, it does not matter that much because this can happen without me watching it.

Also making the tests independent allows to run all the tests instead of aborting after the first failure. This is not possible currently but I try to develop the tests with some idea how to map them to pytest. Then the setup/cleanup would be part of a fixture  and the test/assert part of the actual test. Maybe there will be some opportunities to optimize this then but by writing the tests only in Ansible, optimization seems to be quite hard (when prioritizing readability of tests and logs before execution time).

tyll:
> Thanks. Generally speaking, is there a defined procedure/workflow for testing that the tests properly capture the problem? I used to open such scratch PRs with tests only, but not consistently. An example of such workflow would be:
> 
> * commit changes to tests, push to branch, create PR
> * let the CI fail
> * in the meantime, commit the code changes
> * when CI fails, push the code changes to the branch and watch the re-triggered CI succeed. The failure of the previous CI run will be recorded for posterity with the commit (but only until you rebase).

I think my preference would be to work with two checkouts in the CI and then
- Run the tests from the PR branch against the role/code from the main branch
- Run the tests from the PR branch against itself

Then you do not need any complicated manual process. This would also require the tests to be independent so the CI shows all failures instead of only the first. And the CI runner would need adjust the roles symlink or do something else to specify where to find the role.

pcahyna:
> Then you do not need any complicated manual process.

But how to report the results then? We do not want to report them as failed tests, because the failure is expected. We can't invert the result, because most of tests will pass for the old version as well. The CI will need to get much smarter at aggregating the results, executing tests twice is not enough.

> This would also require the tests to be independent so the CI shows all failures instead of only the first.

For playbooks it will be easy, for multiple tests in a playbook not so easy.
But if you reject the idea of chaining multiple test plays and you write your tests as independent blocks, they can be separated into multiple playbooks anyway.

> And the CI runner would need adjust the roles symlink or do something else to specify where to find the role

This will not be difficult and we should do it anyway.

tests/playbooks/tests_states.yml
pcahyna: > Also making the tests independent allows to run all the tests instead of aborting after the first failure. This is not possible currently

If they are independent, they could be separated easily into separate playbooks and running all of them will then be easy.

tyll:
> > Then you do not need any complicated manual process.
> 
> But how to report the results then? We do not want to report them as failed tests, because the failure is expected. We can't invert the result, because most of tests will pass for the old version as well. The CI will need to get much smarter at aggregating the results, executing tests twice is not enough.

This needs manual inspection anyhow, so it could just be reported in a comment that mentions which tests failed.

> > This would also require the tests to be independent so the CI shows all failures instead of only the first.
> 
> For playbooks it will be easy, for multiple tests in a playbook not so easy.
> But if you reject the idea of chaining multiple test plays and you write your tests as independent blocks, they can be separated into multiple playbooks anyway.

We can also record the failure in a rescue blog and then as a last step in the playbook fail the whole run and print the test results in the end.

> > And the CI runner would need adjust the roles symlink or do something else to specify where to find the role
> 
> This will not be difficult and we should do it anyway.

:+1:

module_utils/network_lsr/argument_validator.py
```diff
@@ -1088,8 +1088,10 @@ def _validate_post_state(self, value, name, result):
             actions.append(persistent_state)
 
         # If the profile should be absent at the end, it needs to be present in
-        # the meantime to allow to (de)activate it
-        if persistent_state == "absent" and state:
+        # the meantime to allow to (de)activate it. This is only possible if it
+        # is completely defined, for which `type` needs to be specified.
+        # Otherwise, downing is happening on a best-effort basis
+        if persistent_state == "absent" and state and result.get("type"):
```
pcahyna, line 1094: Does this change the role interface? What happened before the change if persistent_state == "absent" and state is given and type is not present?

tyll: It depends on whether the profile existed on the system, if it was there, it would be used, otherwise there would be a failure. Type needs to be specified for "present" to work in case the profile is not on the system, otherwise it would just check this.

pcahyna:
> > > Then you do not need any complicated manual process.
> > 
> > 
> > But how to report the results then? We do not want to report them as failed tests, because the failure is expected. We can't invert the result, because most of tests will pass for the old version as well. The CI will need to get much smarter at aggregating the results, executing tests twice is not enough.
> 
> This needs manual inspection anyhow, so it could just be reported in a comment that mentions which tests failed.

I hoped for a more automated inspection. The CI runner should ensure automatically that there is at least one test failure in the old code. The identification of whether it is the correct one will probably need manual inspection, although even this could be eventually automated with enough structure of the tests.

> > > This would also require the tests to be independent so the CI shows all failures instead of only the first.
> > 
> > 
> > For playbooks it will be easy, for multiple tests in a playbook not so easy.
> > But if you reject the idea of chaining multiple test plays and you write your tests as independent blocks, they can be separated into multiple playbooks anyway.
> 
> We can also record the failure in a rescue blog and then as a last step in the playbook fail the whole run and print the test results in the end.

That's an interesting suggestion - I considered something like this to ensure cleanups are always executed, but found it too much a complication of the test structure. But with a wrapper/driver like your run_test.yml, it will not be that complicated anymore.

module_utils/network_lsr/argument_validator.py
pcahyna: So it is not an interface change, because if the interface had not been there, the role failed, and if it had been there, the `actions.append("present")` was superfluous.

tyll: yes, with profile instead of interface.

tyll:
> > > > Then you do not need any complicated manual process.
> > > 
> > > 
> > > But how to report the results then? We do not want to report them as failed tests, because the failure is expected. We can't invert the result, because most of tests will pass for the old version as well. The CI will need to get much smarter at aggregating the results, executing tests twice is not enough.
> > 
> > 
> > This needs manual inspection anyhow, so it could just be reported in a comment that mentions which tests failed.
> 
> I hoped for a more automated inspection. The CI runner should ensure automatically that there is at least one test failure in the old code. The identification of whether it is the correct one will probably need manual inspection, although even this could be eventually automated with enough structure of the tests.

Sure, further optimizations are probably possible. For me it would already be great not having to create draft PRs to show a CI failure.

PR #253, opened 2020-06-25
library: Rename invocation return value

tyll:
Since `invocation` is an Ansible special value that is set for
triple verbosity, it is now visible for smaller verbosity. Therefore
rename it to `_invocation`, which seems to work with less verbosity.
Also adjust the content to match the regular `invocation` style.


pcahyna:
Can you please show the sample output? I am interested whether Ansible formats the value properly.

tyll:
> Can you please show the sample output? I am interested whether Ansible formats the value properly.

I am using the debug task with the var option. This pretty-prints the output.

PR #257, opened 2020-07-03
Simplify persistent state handling

tyll:
Since the order of the actions list does not matter anymore because the
role iterates over all actions in a defined order, simplify the
handling for persistent states.


tyll:
[citest bad]

tyll:
[citest bad]

PR #258, opened 2020-07-06
README: Clarify order of dependent profiles

tyll:


tyll:
This fails because the CI is having network problems.

thom311:
Till, I am not entirely sure why you pinged me for feedback on such a (seemingly) clear patch. I wonder whether I miss an issue?

Otherwise, lgtm!

tyll:
> Till, I am not entirely sure why you pinged me for feedback on such a (seemingly) clear patch. I wonder whether I miss an issue?

Thank you, I wanted to ensure that the described order is correct and that no edge case might have been missed.

PR #259, opened 2020-07-06
Forget unmanaged state in NetworkManager

tyll:
When removing a profile with initscripts, also notify NetworkManager so
it can forget that the specified device was to be ignored.


library/network_connections.py
```diff
@@ -2340,6 +2341,36 @@ def check_name(self, idx, name=None):
...
         return f
 
+    def forget_nm_connection(self, path):
+        """
+        Forget a NetworkManager connection by loading the path of the deleted
+        profile. This inverts the effect of loading a profile with
+        `NM_CONTROLLED=no` earlier, which made NetworkManager ignore the
+        device.
+
+        This does ot use the Python libnm bindings because they might not be
```
pcahyna, line 2351: *not

library/network_connections.py
```diff
@@ -2340,6 +2341,36 @@ def check_name(self, idx, name=None):
...
+        present on the system, since the module is currently operating for the
```
pcahyna, line 2352: Would it make sense to install those bindings every time that there is a NetworkManager package present on the system, even if the module is not operating for the nm provider? What package is it actually, `python-gobject-base`?

library/network_connections.py
tyll: thx, fixed

library/network_connections.py
tyll: > Would it make sense to install those bindings every time that there is a NetworkManager package present on the system, even if the module is not operating for the nm provider? 

Installing it seems to be rather intrusive. This is now the same code I proposed for https://bugzilla.redhat.com/show_bug.cgi?id=1832897 therefore I do not see much harm in the current way. Another advantage of this approach is that it keeps the provider specific code separate.

> What package is it actually, `python-gobject-base`?
Yes or `python3-gobject-base`.

pcahyna:
Is the first commit related to the second in any way?

tyll:
> Is the first commit related to the second in any way?

The downstream bug https://bugzilla.redhat.com/show_bug.cgi?id=1800627 exposed the problem using the test from the first commit. I can also move it into its own PR If you prefer.

pcahyna:
> > Is the first commit related to the second in any way?
> 
> The downstream bug https://bugzilla.redhat.com/show_bug.cgi?id=1800627 exposed the problem using the test from the first commit. 

I know, but the former version did expose the problem just as well as the new one, didn't it? So the change is not related to the fact that the test can expose the problem.

BTW. Why does a RHEL 7 CI test fail? The failure looks real and related to the PR.

tyll:
> > > Is the first commit related to the second in any way?
> > 
> > 
> > The downstream bug https://bugzilla.redhat.com/show_bug.cgi?id=1800627 exposed the problem using the test from the first commit.
> 
> I know, but the former version did expose the problem just as well as the new one, didn't it? So the change is not related to the fact that the test can expose the problem.
> 
> BTW. Why does a RHEL 7 CI test fail? The failure looks real and related to the PR. EDIT: I was able to reproduce the failure locally with `rhel-guest-image-7.7-200`.

@thom311 any idea? Does NetworkManager on RHEL/CentOS 7 maybe not support forgetting individual profiles?

pcahyna:
Now even rhel-8 is failing.

tyll:
> Now even rhel-8 is failing.

there was a bug in the test which is why centos 7 was not failing but it should fail now similar to rhel 7.

pcahyna:
Good to hear that CentOS 7 and RHEL 7 are at least behaving consistently.

tyll:
> Good to hear that CentOS 7 and RHEL 7 are at least behaving consistently.

yes, the CentOS image does not contain NetworkManager but it needs to be present when using initscripts to make NM store the state. Now it should be installed if necessary before using initscripts.

thom311:
> @thom311 any idea? Does NetworkManager on RHEL/CentOS 7 maybe not support forgetting individual profiles?

If you issue (the D-Bus equivalent of) `nmcli connection reload` or `nmcli connection load`, then files that no longer exist on disk must be unloaded/removed. The effect of that is pretty much the same as issuing `nmcli connection delete`. If that's not the case, please report a bug.

tyll:
> > @thom311 any idea? Does NetworkManager on RHEL/CentOS 7 maybe not support forgetting individual profiles?
> 
> If you issue (the D-Bus equivalent of) `nmcli connection reload` or `nmcli connection load`, then files that no longer exist on disk must be unloaded/removed. The effect of that is pretty much the same as issuing `nmcli connection delete`. If that's not the case, please report a bug.

It works like this in RHEL/CentOS 8 but not 7.

thom311:
> It works like this in RHEL/CentOS 8 but not 7.

Ah. You are right. That changed with the large rework in 1.20. Nowadays, "load" really means: revisit all related files on disk, and update them -- possibly remove them.

Indeed that changed since RHEL7. The rework cannot be backported, because it's too large. And there is no desire to try to come up with a "small" adjustment for this behaviour.

Sorry about the inconvenience, and that I didn't realize this to be the case right away.

pcahyna:
Well, better to have it fixed in RHEL 8 only than to not have it fixed at all. I will adjust the affected test somehow so the problematic part is skipped on RHEL 7.

PR #260, opened 2020-07-07
Add Pytest integration tests

elvgarrui:
tests/integration/test_ethernet.py is a script capable of using the network
module without executing it from Ansible and using Pytest. The example test
sets an ip in a test interface.

### DIFF
--- a/pylint_extra_requirements.txt
+++ b/pylint_extra_requirements.txt
@@ -1,3 +1,5 @@
 # SPDX-License-Identifier: MIT
 
 # Write extra requirements for running pylint here:
+mock
+pytest
--- a/tests/integration/test_ethernet.py
+++ b/tests/integration/test_ethernet.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*
+# SPDX-License-Identifier: BSD-3-Clause
+
+import logging
+import os
+import pytest
+import subprocess
+import sys
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+parentdir = os.path.normpath(os.path.join(os.path.dirname(__file__), "../../"))
+with mock.patch.object(
+    sys,
+    "path",
+    [parentdir, os.path.join(parentdir, "module_utils/network_lsr")] + sys.path,
+):
+    with mock.patch.dict(
+        "sys.modules",
+        {
+            "ansible": mock.Mock(),
+            "ansible.module_utils": __import__("module_utils"),
+            "ansible.module_utils.basic": mock.Mock(),
+        },
+    ):
+        import library.network_connections as nc
+
+
+class PytestRunEnvironment(nc.RunEnvironment):
+    def log(self, connections, idx, severity, msg, **kwargs):
+        if severity == nc.LogLevel.ERROR:
+            logging.error("Error: {}".format(connections[idx]))
+            raise RuntimeError(msg)
+        else:
+            logging.debug("Log: {}".format(connections[idx]))
+
+    def run_command(self, argv, encoding=None):
+        command = subprocess.Popen(
+            argv, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        return_code = command.wait()
+        out, err = command.communicate()
+        return return_code, out.decode("utf-8"), err.decode("utf-8")
+
+    def _check_mode_changed(self, *args, **kwargs):
+        pass
+
+
+def _configure_network(connections, provider):
+    cmd = nc.Cmd.create(
+        provider,
+        run_env=PytestRunEnvironment(),
+        connections_unvalidated=connections,
+        connection_validator=nc.ArgValidator_ListConnections(),
+    )
+    cmd.run()
+
+
+@pytest.fixture(scope="session")
+def provider(request):
+    return request.config.getoption("--provider")
+
+
+@pytest.fixture
+def testnic1():
+    veth_name = "testeth"
+    try:
+        subprocess.call(
+            [
+                "ip",
+                "link",
+                "add",
+                veth_name,
+                "type",
+                "veth",
+                "peer",
+                "name",
+                veth_name + "peer",
+            ],
+            close_fds=True,
+        )
+        yield veth_name
+    finally:
+        subprocess.call(["ip", "link", "delete", veth_name])
+
+
+def _get_ip_addresses(interface):
+    ip_address = subprocess.check_output(["ip", "address", "show", interface])
+    return ip_address.decode("UTF-8")
+
+
+def test_static_ip_with_ethernet(testnic1, provider):
+    ip_address = "192.0.2.127/24"
+    connections = [
+        {
+            "name": testnic1,
+            "type": "ethernet",
+            "state": "up",
+            "ip": {"address": [ip_address]},
+        }
+    ]
+    _configure_network(connections, provider)
+    assert ip_address in _get_ip_addresses(testnic1)
+    assert os.path.exists("/etc/sysconfig/network-scripts/ifcfg-" + testnic1)
--- a/tests/playbooks/integration_pytest_python3.yml
+++ b/tests/playbooks/integration_pytest_python3.yml
@@ -1,0 +1,97 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Install dependencies for integration tests
+  hosts: all
+  vars:
+    - rpmdependencies:
+        - git
+        - python3-pip
+        - rsync
+
+  tasks:
+    - name: Install EPEL for RHEL and CentOS
+      # yamllint disable-line rule:line-length
+      command: "yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm"
+      when: ansible_distribution in ["CentOS", "RedHat"]
+
+    - name: Install rpm dependencies
+      package:
+        state: present
+        name: "{{ rpmdependencies }}"
+
+    - name: Install Pytest
+      command: "pip3 install pytest"
+
+
+# Import needed in order to install initscripts dependencies on the remote
+# system.
+- import_playbook: "../tests_default_initscripts.yml"
+
+# Import needed in order to install Network Manager dependencies on the remote
+# system.
+- import_playbook: "../tests_default_nm.yml"
+
+
+- name: Run Pytest tests
+  hosts: all
+  vars:
+    - rundir: /run/system-roles-test
+  tasks:
+    - file:
+        state: directory
+        path: "{{ rundir }}"
+        recurse: true
+
+    - command: git rev-parse --show-toplevel
+      register: git_top_directory
+      delegate_to: localhost
+
+    - debug:
+        var: git_top_directory
+
+    - synchronize:
+        src: "{{ git_top_directory.stdout }}/"
+        dest: "{{ rundir }}/"
+        recursive: yes
+        delete: yes
+        rsync_opts:
+          - "--exclude=.pyc"
+          - "--exclude=__pycache__"
+      when: False
+
+    # TODO: using tar and copying the file is a workaround for the synchronize
+    # module that does not work in test-harness. Related issue:
+    # https://github.com/linux-system-roles/test-harness/issues/102
+    #
+    - name: Create Tar file
+      shell: 'tar -cvf {{ git_top_directory.stdout }}/testrepo.tar
+             --exclude "*.pyc" --exclude "__pycache__" --exclude testrepo.tar
+             -C {{ git_top_directory.stdout }} .'
+      delegate_to: localhost
+
+    - name: Copy testrepo.tar to the remote system
+      copy:
+        src: "{{ git_top_directory.stdout }}/testrepo.tar"
+        dest: "{{ rundir }}"
+
+    - name: Untar testrepo.tar
+      command: tar xf testrepo.tar
+      args:
+        chdir: "{{ rundir }}"
+
+    - block:
+        - name: Run pytest with nm
+          command: "pytest {{ rundir }}/tests/integration/ --provider=nm"
+          register: playbook_run
+      always:
+        - debug:
+            var: playbook_run.stdout_lines
+
+    - block:
+        - name: Run pytest with initscripts
+          command: >
+            pytest '{{ rundir }}/tests/integration/' --provider=initscripts
+          register: playbook_run
+      always:
+        - debug:
+            var: playbook_run.stdout_lines
--- a/tests/tests_integration_pytest.yml
+++ b/tests/tests_integration_pytest.yml
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Empty play to gather facts
+  hosts: all
+
+- import_playbook: playbooks/integration_pytest_python3.yml
+  when: (ansible_distribution in ["CentOS", "RedHat"] and
+        ansible_distribution_major_version == "8") or
+        ansible_distribution == "Fedora"
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--provider", action="store", default="nm", help="Network provider"
+    )


tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,89 @@
+from unittest import mock
```
tyll, line 1: Ideally, the script should also work with Python 2, so we can use it on RHEL 6/7, too.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,89 @@
...
+        connections_unvalidated=connections,
+        connection_validator=nc.ArgValidator_ListConnections(),
+    )
+
+    cmd.run()
+
+
+@pytest.fixture
+def set_test_connection():
+    veth_name = "testeth"
```
tyll, line 51: Please use a constant for the interface

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,89 @@
...
+                veth_name + "peer",
+            ]
+        )
+        yield veth_name
+    finally:
+        subprocess.run(["ip", "link", "delete", veth_name])
+
+
+def check_interface_ip_set(connection_name, expected_ipaddr):
+    ip_a = subprocess.run(["ip", "addr", "show", connection_name], capture_output=True)
```
tyll, line 72: FYI: In the future it might be good to use nispor to get the interface state: https://github.com/nispor/nispor - if it works on RHEL 6. Maybe we can store a pre-compiled binary somewhere to use it @cathay4t what do you think?

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
+import os
+import pytest
+import subprocess
+import sys
+
+VETH_NAME = "testeth"
+
+if sys.version_info >=(3, 3):
```
tyll, line 8: A typical approach for this is to use:

```
try:
    from unittest import mock

File: tests/integration/test_ethernet.py [hunk 5]
```diff
@@ -0,0 +1,93 @@
...
+    from unittest import mock
+else:
+    import mock
```
tyll, line 11:
```suggestion
try:
    from unittest import mock
except ImportError:
    import mock
```

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
+import os
```
tyll, line 1: Please add a licensing comment, see the other files for this.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
...
+        with mock.patch.dict(
+            "sys.modules",
+            {
+                "ansible": mock.Mock(),
+                "ansible.module_utils": __import__("module_utils"),
+            },
+        ):
+            import library.network_connections as nc
+
+    run_env = nc.RunEnvironment()
```
tyll, line 29: Please create a class like "TestRunEnvironment" that inherits from "RunEnvironment" and implements the necessary methods like the "AnsibleRunEnvironment" does it in the library.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
...
+        run_env=run_env,
+        connections_unvalidated=connections,
+        connection_validator=nc.ArgValidator_ListConnections(),
+    )
+
+    cmd.run()
+
+
+@pytest.fixture
+def set_test_connection():
```
tyll, line 55:
```suggestion
def testnic1():
```

might make it more clear since the fixture is used like a variable in the actual test method.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
...
+    try:
+        subprocess.run(
+            [
+                "ip",
+                "link",
+                "add",
+                VETH_NAME,
```
tyll, line 62: Sorry about the back and forth, since the interface name is only used inside the fixture and passed to the test method, a variable makes more sense instead of a constant. I did not realize this before.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
...
+                VETH_NAME + "peer",
+            ]
+        )
+        yield VETH_NAME
+    finally:
+        subprocess.run(["ip", "link", "delete", VETH_NAME])
+
+
+def check_interface_ip_set(connection_name, expected_ipaddr):
+    ip_a = subprocess.run(["ip", "addr", "show", connection_name], capture_output=True)
```
tyll, line 76:
```suggestion
    ip_a = subprocess.run(["ip", "address", "show", connection_name], capture_output=True)
```

I was not aware that the full command name is "address", so please use it here.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,93 @@
...
+                "name",
+                VETH_NAME + "peer",
+            ]
+        )
+        yield VETH_NAME
+    finally:
+        subprocess.run(["ip", "link", "delete", VETH_NAME])
+
+
+def check_interface_ip_set(connection_name, expected_ipaddr):
```
tyll, line 75: It seems to be nicer to make this just return the output of "ip addresss..." and call it for example "get_ip_addresses(interface)" and the assert could be "assert ip_address in get_ip_address(interface)" in the test method.

tests/integration/test_ethernet.py
ffmancera: IMO, this should go on the import section.

tests/integration/test_ethernet.py
ffmancera, line 1: This file is missing a copyright header, it should look like:

```
# -*- coding: utf-8 -*-
```

tests/integration/test_ethernet.py
elvgarrui: Fixed, thanks! =)

tests/integration/test_ethernet.py
elvgarrui: I added an if statement to consider this difference, thanks for pointing it!

tests/integration/test_ethernet.py
elvgarrui: I fixed it, thanks! I did this like the first way you mentioned. Although I might not be right, for me it sounds wrong to add classes or functions to the library as long as they are only going to be used for tests. Maybe it would be useful to make a "utils" module for tests in a future though.

tests/integration/test_ethernet.py
elvgarrui: ok!

tests/integration/test_ethernet.py
elvgarrui: ok, done!

tests/integration/test_ethernet.py
elvgarrui: I think I've reestructured it as you say, thanks! It is more clear like this.

tests/integration/test_ethernet.py
elvgarrui: I added it, thanks!

elvgarrui:
@tyll, @ffmancera: Thanks for the reviews, I went over them all! =)

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,96 @@
...
+    class TestRunEnvironment(nc.RunEnvironment):
+        def log(self, connections, idx, severity, msg, **kwargs):
+            if severity == nc.LogLevel.ERROR:
+                print("Error: {}".format(connections[idx]))
+                raise RuntimeError(msg)
+            else:
+                print("Log: {}".format(connections[idx]))
+
+        def _check_mode_changed(self, *args, **kwargs):
+            pass
```
tyll, line 40: This should be moved to the top level

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,96 @@
...
+    finally:
+        subprocess.run(["ip", "link", "delete", veth_name])
+
+
+def _get_ip_addresses(connection_name):
+    ip_a = subprocess.run(
+        ["ip", "address", "show", connection_name], capture_output=True
+    )
+    interface = ip_a.stdout.decode("UTF-8")
+    return interface
```
tyll, line 81:
```suggestion
    return ip_a.stdout.decode("UTF-8")
```

`interface` does not seem to be a proper name but it also seems the variable is not needed.

tyll:
Thank you, it looks good. There are some issues regarding Python 2 support: https://travis-ci.com/github/linux-system-roles/network/jobs/359087282#L257
I hope it is feasible to fix them.

tests/integration/test_ethernet.py
elvgarrui: True! Fixed

elvgarrui:
@tyll I was able to fix the existing CI problems but I still cannot figure out how to make the test work using pytest in Python2.7.
I'm using the command "python2.7 -m pytest" to execute it, and the following error occurs:
```
ImportError while importing test module '/home/elvira/network/tests/integration/test_ethernet.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/integration/test_ethernet.py:25: in <module>
    {"ansible": mock.Mock(), "ansible.module_utils": __import__("module_utils")},
E   ImportError: No module named module_utils
```

tests/integration/test_ethernet.py
elvgarrui: Done, thanks!

tox.ini
```diff
@@ -77,7 +77,9 @@ changedir = {toxinidir}
 deps =
     ansible
     colorama
+    mock
```
tyll, line 80: please put the extra requirements in the pylint_extra_requirements.txt file - the system roles are a little bit  special since they try to use a common tox.ini from the template repository in the system roles project.

tyll:
> @tyll I was able to fix the existing CI problems but I still cannot figure out how to make the test work using pytest in Python2.7.
> I'm using the command "python2.7 -m pytest" to execute it, and the following error occurs:
> 
> ```
> ImportError while importing test module '/home/elvira/network/tests/integration/test_ethernet.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> tests/integration/test_ethernet.py:25: in <module>
>     {"ansible": mock.Mock(), "ansible.module_utils": __import__("module_utils")},
> E   ImportError: No module named module_utils
> ```

You need to install ansible for python2.7 as well. Are you using a virtualenv/venv? It would be good to get familiar with. What distribution are you running the tests on? A good next step would probably be to start writing down the requirements that are needed to run the integration tests, ideally in an ansible Playbook.

richm:
You should use `tox -e py27` to run the unit tests on python 2.7.  Then, to run integration tests with python 2.7 and a specified version of ansible
```
. .tox/env-2.7/bin/activate # was setup by running tox -e py27
pip install 'ansible<2.8' # for ansible 2.7
cd tests
TEST_SUBJECTS=/path/to/my/image.qcow2 ANSIBLE_STDOUT_CALLBACK=debug \
ansible-playbook -vv -i /usr/share/ansible/inventory/standard-inventory-qcow2 tests_mytest.yml
```
at least, something like this works on Fedora - for ubuntu/suse/mac/windows ymmv
for older versions of ansible, you might also have to install older versions of other components used by ansible e.g. `jinja2<2.9` etc.

tyll:
> You should use `tox -e py27` to run the unit tests on python 2.7. Then, to run integration tests with python 2.7 and a specified version of ansible
> 
> ```
> . .tox/env-2.7/bin/activate # was setup by running tox -e py27
> pip install 'ansible<2.8' # for ansible 2.7
> cd tests
> TEST_SUBJECTS=/path/to/my/image.qcow2 ANSIBLE_STDOUT_CALLBACK=debug \
> ansible-playbook -vv -i /usr/share/ansible/inventory/standard-inventory-qcow2 tests_mytest.yml
> ```
> 
> at least, something like this works on Fedora - for ubuntu/suse/mac/windows ymmv
> for older versions of ansible, you might also have to install older versions of other components used by ansible e.g. `jinja2<2.9` etc.

Thanks @richm . This PR is about adding an integration test with pytest. But I guess actually using "tox -e py27" should run this as well but might also need additional tweaking.

elvgarrui:
[citest bad]

richm:
the python coding looks correct

richm:
How do you run this test?  Are you doing
```
cd tests
TEST_SUBJECT=/path/to/image.qcow2 ansible-playbook -vv -i /usr/share/ansible/inventory/standard-inventory-qcow2 tests_integration_pytest.yml
```
?

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,82 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Empty play to gather facts
+  hosts: all
```
richm, line 4: Is this really needed?

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,82 @@
...
+    - ansible_distribution_major_version == '8'
+
+- name: Run test playbook
+  hosts: all
+  vars:
+    - rundir: /run/system-roles-test
+  tasks:
+    - file:
+        state: directory
+        path: /run
```
richm, line 33: Why `/run` as opposed to a temp directory like `/tmp` or `/var/tmp`?  Or use the `tempfile` https://docs.ansible.com/ansible/latest/modules/tempfile_module.html#tempfile-module module?

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,82 @@
...
+                break
+            else
+                curdir=$(dirname "${curdir}")
+            fi
+        done
+        if [ -z "${gitdir}" ]
+        then
+            exit 1
+        fi
+        echo "$(dirname ${gitdir})"
```
richm, line 54: IIUC what this is supposed to be doing, you can use `command: git rev-parse --show-toplevel` instead of this `shell:` script

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,82 @@
...
+- import_playbook: playbooks/centos8_deps.yml
+  when:
+    - ansible_distribution == "CentOS"
+    - ansible_distribution_major_version == '8'
+
+- name: Run test playbook
+  hosts: all
+  vars:
+    - rundir: /run/system-roles-test
+  tasks:
```
richm, line 30: Instead of having separate playbooks for each platform, with the duplicated code, it would be better to have vars files for each platform defining the packages / services for each platform.  https://github.com/oasis-roles/meta_standards#platform-specific-variables - something like this:
* make a `vars/integration/` subdirectory of `tests/`
* create the files `vars/integration/CentOS-6.yml` etc. - one file for each supported platform

Each file would look something like this:
```yaml
__test_pre_packages: [epel-release]
__test_packages: [git, python3-pip, ansible]
__test_services: [rsyncd]
__test_pip_packages: [pytest]
```
of course these will be different depending on the platform - in addition, for epel support on rhel7, you may have to specify the url of the rpm directly e.g. `__test_pre_packages: ["https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"]` as I don't think the package `epel-release` is in RHEL7.
Then, in `tests/tests_integration_pytest.yml` do something like this:
```yaml
- name: Set platform/version specific variables for testing
  include_vars: "{{ __test_vars_file }}"
  loop:
    - default.yml
    - "{{ ansible_facts['os_family'] }}.yml"
    - "{{ ansible_facts['distribution'] }}.yml"
    - "{{ ansible_facts['distribution'] }}_{{ ansible_facts['distribution_major_version'] }}.yml"
    - "{{ ansible_facts['distribution'] }}_{{ ansible_facts['distribution_version'] }}.yml"
  vars:
    __test_vars_file: "{{ role_path }}/tests/vars/integration/{{ item }}"
  when: __test_vars_file is file

- name: ensure test pre-packages are installed first
  package:
    name: "{{ __test_pre_packages }}"
  when: __test_pre_packages is defined

- name: ensure test packages are installed
  package:
    name: "{{ __test_packages }}"
  when: __test_packages is defined

- name: ensure test services are enabled
  service:
    name: "{{ item }}"
    state: started
    enabled: true
  loop: "{{ __test_services | d([]) }}"

- name: ensure pip packages are installed
  pip:
    name: "{{ __test_pip_packages }}"
  when: __test_pip_packages is defined
```

elvgarrui:
> TEST_SUBJECT=/path/to/image.qcow2 ansible-playbook -vv -i /usr/share/ansible/inventory/standard-inventory-qcow2 tests_integration_pytest.yml

Yes!, I'm just using -v instad of -vv. The playbook works fine for my test image (CentOS 8), but not for the rest of them. That's what I'm looking into right now, to be able to run it in every distribution.

tests/tests_integration_pytest.yml
elvgarrui: It does look like a more clean way of stating everything, I'm going to try to change it this way, thanks!!

pylint_extra_requirements.txt
```diff
@@ -1,3 +1,5 @@
 # SPDX-License-Identifier: MIT
 
 # Write extra requirements for running pylint here:
+mock
+pytest
```
richm, line 5: Is this change related to the integration testing?

elvgarrui: Yes. The file tests/integration/test_ethernet.py uses pytest and mock to make integration tests directly using python instead of using an ansible playbook. The playbook is added just so that the test harness can execute these tests automatically. 
The pylint requirements are necessary so as to not get an import error when running `tox`.

tests/tests_integration_pytest.yml
tyll: Awesome, good to know! Thanks!

tests/tests_integration_pytest.yml
tyll: `/run` does not have any problems regarding insecure temporary files. Using tempfile has the problem that it would not be persistent so it would need to be synced all the time when re-using a machine.

tests/tests_integration_pytest.yml
tyll: This is needed when using `when: ... ansible_distribution` with `import_playbook`

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+    - file:
+        state: directory
+        path: /run
+
+    - command: git rev-parse --show-toplevel
+      register: testdir
```
tyll, line 31:
```suggestion
      register: git_top_directory
```

It seems that `git_top_directory` or similar is a better name here than `testdir` since the actual tests are in a sub directory.

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+  hosts: all
+  roles:
+    - linux-system-roles.network
+
+- import_playbook: playbooks/centos7_deps.yml
+  when:
+    - ansible_distribution == "CentOS"
+    - ansible_distribution_major_version >= '7'
+
+- import_playbook: playbooks/fedora32_deps.yml
```
tyll, line 16: Where did CentOS 8 go? I remember that it worked there, already, didn't it?

tests/playbooks/fedora32_deps.yml
```diff
@@ -0,0 +1,18 @@
...
+  hosts: all
+  vars:
+    - rpmdependencies:
+        - git
+        - python3-pip
+  tasks:
+    - name: Install EPEL
+      package:
+        name: epel-release
+        state: present
```
tyll, line 12: This will not work on Fedora.

tests/playbooks/fedora32_deps.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Install CentOS 8 deps for network system role tests
```
tyll, line 2: This does not match the filename.

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Install CentOS deps for network system role tests
+  hosts: all
+  vars:
+    - rpmdependencies:
+        - git
+        - python3-pip
+        - python2-pip
```
tyll, line 8: Why python 2 and python 3?

tests/tests_integration_pytest.yml
elvgarrui: I just thought about using the same playbook for both, so I changed this line 14 to `>= '7'`

tests/playbooks/centos7_deps.yml
elvgarrui: I was thinking of using this playbook to load the CentOS7 python 2 dependencies too. It might be too much of a mixture though.

tests/playbooks/fedora32_deps.yml
elvgarrui: True, Fixed!

tests/playbooks/fedora32_deps.yml
elvgarrui: You are right. It is also unnecessary for the given distribution version.

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,103 @@
...
+    cmd = nc.Cmd.create(
+        provider,
+        run_env=PytestRunEnvironment(),
+        connections_unvalidated=connections,
+        connection_validator=nc.ArgValidator_ListConnections(),
+    )
+    cmd.run()
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
elvgarrui: True, thanks!

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,103 @@
...
+def cmdopt(request):
```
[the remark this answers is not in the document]
elvgarrui: changed!

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,106 @@
...
+                "peer",
+                "name",
+                veth_name + "peer",
+            ],
+            close_fds=True,
+        )
+        create_iface.wait()
+        yield veth_name
+    finally:
+        delete_iface = subprocess.Popen(["ip", "link", "delete", veth_name])
```
tyll, line 85: If you do not need stdout, you can use `subprocess.call()` to simplify things:
https://docs.python.org/2.7/library/subprocess.html#using-the-subprocess-module

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+  when:
+    - ansible_distribution == "Fedora"
+    - ansible_distribution_major_version == '32'
+
+- name: Run Pytest tests
```
tyll, line 16: It might be good to move this into its own playbook and import it only for the already supported distributions, CentOS 8 and Fedora 32 to avoid the failures on the other distributions.

tests/tests_integration_pytest.yml
tyll: In general this would be possible but currently it is only imported on Fedora 32, so it would not be used on CentOS 8 as far as I can see. Once we have something working for the distributions, the next step would be to clean it up using the variables as suggested in the other comment and group these packages maybe by other features:

- if it is CentOS/RHEL: set the pre_pacakges to epel-release
- if it is python3: Use python3-pip
- all: install git, rsync
- all: install pytest via pip

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,35 @@
...
+  tasks:
+    - name: Install EPEL
+      package:
+        state: present
+        name: epel-release
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == '7'
```
tyll, line 16: This is required for CentOS 8, too.

elvgarrui: I will add it. It was working correctly on my CentOS 8 vm without it, though!

tests/integration/test_ethernet.py
elvgarrui: I see, thanks!

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,103 @@
...
+            "ansible.module_utils.basic": mock.Mock(),
+        },
+    ):
+        import library.network_connections as nc
+
+
+class PytestRunEnvironment(nc.RunEnvironment):
+    def log(self, connections, idx, severity, msg, **kwargs):
+        if severity == nc.LogLevel.ERROR:
+            print("Error: {}".format(connections[idx]))
```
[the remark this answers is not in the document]
elvgarrui: Done, thanks!

tests/integration/test_ethernet.py
```diff
@@ -0,0 +1,103 @@
...
+            raise RuntimeError(msg)
+        else:
+            print("Log: {}".format(connections[idx]))
+
+    def run_command(self, argv, encoding=None):
+        command = subprocess.Popen(argv)
```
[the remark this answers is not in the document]
elvgarrui: I see, thanks!

tests/tests_integration_pytest.yml
elvgarrui: Done, thanks!

tests/playbooks/centos7_deps.yml
tyll: Please remove python2-pip for now.

tests/playbooks/fedora32_deps.yml
```diff
@@ -0,0 +1,20 @@
...
+    - name: Install rpm dependencies
+      package:
+        state: present
+        name: "{{ rpmdependencies }}"
+    - name: Install rsync
+      package:
+        state: present
+        name: rsync
+    - name: Search for rsync route
+      command: which rsync
```
tyll, line 18: This looks like a debugging leftover

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,32 @@
...
+      command: "pip install setuptools --upgrade"
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == '7'
+
+    - name: Install Pytest for Python 2
+      command: "pip install pytest==2.7.2 mock"
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == '7'
```
tyll, line 32: Please remove this for now

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,32 @@
...
+        state: present
+        name: "{{ rpmdependencies }}"
+
+    - name: Install Pytest and Ansible
+      command: "pip3 install pytest"
+    - name: Update pip
+      command: "pip install setuptools --upgrade"
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == '7'
```
tyll, line 26: Please remove this

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,32 @@
...
+      package:
+        state: present
+        name: epel-release
+
+    - name: install rpm dependencies
+      package:
+        state: present
+        name: "{{ rpmdependencies }}"
+
+    - name: Install Pytest and Ansible
```
tyll, line 20:
```suggestion
    - name: Install Pytest
```

tests/playbooks/centos7_deps.yml
```diff
@@ -0,0 +1,32 @@
...
+  vars:
+    - rpmdependencies:
+        - git
+        - python3-pip
+        - python2-pip
+  tasks:
+    - name: Install EPEL
+      package:
+        state: present
+        name: epel-release
```
tyll, line 13: I think this is the only difference between Fedora 32 and CentOS 8, so you can add a when: condition here and move all the installation steps into the tests/playbooks/integration_pytest_python3.yml  playbook as far as I can see.

tyll: Also please add the installation step for EPEL with RHEL (see the EPEL wiki for the URL/command).

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,66 @@
...
+
+- import_playbook: centos7_deps.yml
+  when:
+    - ansible_distribution == "CentOS"
+    - ansible_distribution_major_version >= '7'
+
+- import_playbook: fedora32_deps.yml
+  when:
+    - ansible_distribution == "Fedora"
+    - ansible_distribution_major_version == '32'
```
tyll, line 14: Please replace this with the actual installation steps

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Empty play to gather facts
+  hosts: all
+
+- import_playbook: playbooks/integration_pytest_python3.yml
+  when: (ansible_distribution in ["CentOS","RedHat"] and
```
tyll, line 7:
```suggestion
  when: (ansible_distribution in ["CentOS", "RedHat"] and
```

tests/tests_integration_pytest.yml
```diff
@@ -0,0 +1,9 @@
...
+        ansible_distribution_major_version == "8") or
+        (ansible_distribution == "Fedora")
```
tyll, line 9:
```suggestion
        ansible_distribution == "Fedora"
```

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+        src: "{{ git_top_directory.stdout }}/testrepo.tar"
+        dest: "{{ rundir }}"
+
+    - name: Untar testrepo.tar
+      command: tar xf testrepo.tar
+      args:
+        chdir: "{{ rundir }}"
+
+
+# Import needed in order to install Network Manager dependencies in the remote
```
tyll, line 75:
```suggestion
# Import needed in order to install Network Manager dependencies on the remote
```

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+  tasks:
+    - block:
+        - name: Run pytest with nm
+          command: "pytest {{ rundir }}/tests/integration/ --provider=nm"
+          register: playbook_run
+      always:
+        - debug:
+            var: playbook_run.stdout_lines
+
+# Import needed in order to install initscripts dependencies into the remote
```
tyll, line 90:
```suggestion
# Import needed in order to install initscripts dependencies on the remote
```

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+# system.
+- import_playbook: "../tests_default_initscripts.yml"
```
tyll, line 92: It would probably be nicer to move the imports for installing dependencies to the top right after installing the other dependencies, then all the task for running pytest can be in one playbook. Then the `rundir` variable does not need to be defined several times.

elvgarrui:
@tyll I completed all the reviews, thanks!!

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+      command: which rsync
+
+    - name: Install Pytest
+      command: "pip3 install pytest"
+
+
+# Import needed in order to install initscripts dependencies on the remote
+# system.
+- import_playbook: "../tests_default_initscripts.yml"
+- hosts: all
```
tyll, line 32: The hosts: line does not make sense, here.

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+
+# Import needed in order to install Network Manager dependencies on the remote
+# system.
+- import_playbook: "../tests_default_nm.yml"
+- hosts: all
```
tyll, line 37: The hosts: line does not make sense, here.

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+      command: "yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm"
+      when: ansible_distribution in ["CentOS", "RedHat"]
+
+    - name: Install rpm dependencies
+      package:
+        state: present
+        name: "{{ rpmdependencies }}"
+
+    - name: Search for rsync path
+      command: which rsync
```
tyll, line 23:
```suggestion
```

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+        - name: Run pytest with nm
+          command: "pytest {{ rundir }}/tests/integration/ --provider=nm"
+          register: playbook_run
+      always:
+        - debug:
+            var: playbook_run.stdout_lines
+
+    - name: Set network provider to 'initscripts'
+      set_fact:
+        network_provider: initscripts
```
tyll, line 96: This seems to be wrong/a leftover

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,104 @@
...
+
+
+- name: Run Pytest tests
+  hosts: all
+  vars:
+    - rundir: /run/system-roles-test/
+  tasks:
+    - file:
+        state: directory
+        path: /run
```
tyll, line 47:
```suggestion
        path: "{{ rundir }}"
        recurse: true
```

If this works, it would allow to avoid the hardcoded `/run` directory here and only use the variable.

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -0,0 +1,100 @@
...
+    - name: Search for rsync path
+      command: which rsync
+
+    - name: Install Pytest
+      command: "pip3 install pytest"
+
+- name: Run Pytest tests
+  hosts: all
+  vars:
+    - rundir: /run/system-roles-test/
```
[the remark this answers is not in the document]
tyll: Which task failed? The copy task for the tarball?

tests/playbooks/integration_pytest_python3.yml
elvgarrui: Done, thanks!

tests/playbooks/integration_pytest_python3.yml
elvgarrui: Done, thanks!

tests/playbooks/integration_pytest_python3.yml
elvgarrui: It works!! Thanks

elvgarrui:
[citest bad]

tests/playbooks/integration_pytest_python3.yml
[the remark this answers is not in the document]
elvgarrui: The fail was on the copy. However, it is not happening now. I will remove the slash then.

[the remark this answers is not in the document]
tyll: I see, this is a special behavior of the copy module: https://docs.ansible.com/ansible/latest/modules/copy_module.html#parameter-dest - it works now without the slash because the playbook creates the directory beforehand (using the modified copy task).

tyll (approved):
Good work, thank you!

PR #261, opened 2020-08-07
sync with latest template including shellcheck

richm:
fix shellcheck issues
fix yamllint issues with truthy values

### DIFF
--- a/.travis/config.sh
+++ b/.travis/config.sh
@@ -53,7 +53,9 @@
 #
 #       - RUN_SHELLCHECK_DISABLED
 #       - RUN_SHELLCHECK_EXTRA_ARGS
-type -f lsr_check_python_version > /dev/null 2>&1 || . ${SCRIPTDIR}/utils.sh
+type -f lsr_check_python_version > /dev/null 2>&1 || . "${SCRIPTDIR}/utils.sh"
+# https://github.com/koalaman/shellcheck/wiki/SC2034
+# shellcheck disable=SC2034
 if lsr_check_python_version python -lt 3.0
 then
     PYTHON2_EXCLUDES="tests/ensure_provider_tests.py,scripts/print_all_options.py"
@@ -61,4 +63,6 @@ then
     RUN_PYLINT_EXCLUDE='^(\..*|ensure_provider_tests\.py|print_all_options\.py)$'
     RUN_FLAKE8_EXTRA_ARGS="--exclude ${FLAKE8_DEFAULT_EXCLUDES},${PYTHON2_EXCLUDES}"
 fi
+# https://github.com/koalaman/shellcheck/wiki/SC2034
+# shellcheck disable=SC2034
 LSR_PUBLISH_COVERAGE=normal
--- a/.travis/config.sh
+++ b/.travis/config.sh
@@ -1,4 +1,5 @@
-#!/bin/bash
+# https://github.com/koalaman/shellcheck/wiki/SC2148
+# shellcheck shell=bash
 # SPDX-License-Identifier: MIT
 #
 # Use this file to specify custom configuration for a project. Generally, this
@@ -40,6 +41,16 @@
 #
 #       - RUN_FLAKE8_DISABLED
 #       - RUN_FLAKE8_EXTRA_ARGS
+type -f lsr_check_python_version > /dev/null 2>&1 || . "${SCRIPTDIR}/utils.sh"
+# https://github.com/koalaman/shellcheck/wiki/SC2034
+# shellcheck disable=SC2034
+if lsr_check_python_version python -lt 3.0
+then
+    PYTHON2_EXCLUDES="tests/ensure_provider_tests.py,scripts/print_all_options.py"
+    FLAKE8_DEFAULT_EXCLUDES=".svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg"
+    RUN_PYLINT_EXCLUDE='^(\..*|ensure_provider_tests\.py|print_all_options\.py)$'
+    RUN_FLAKE8_EXTRA_ARGS="--exclude ${FLAKE8_DEFAULT_EXCLUDES},${PYTHON2_EXCLUDES}"
+fi
 #
 #   * .travis/runsyspycmd.sh:
 #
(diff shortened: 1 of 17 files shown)


.travis/custom.sh
```diff
@@ -3,13 +3,14 @@
 
 set -e
 
-ME=$(basename $0)
-SCRIPTDIR=$(readlink -f $(dirname $0))
+#uncomment if you use $ME - otherwise set in utils.sh
```
nhosoi, line 6: This pr sets ME in utils.sh [0]. So we could just get rid of setting ME in each shell script? (not just from the scripts ME is not used, but is being used such as runcoveralls.sh, runflake8.sh, etc.?)
[0] - https://github.com/linux-system-roles/network/pull/261/files#diff-0dc30956d7844f3c69f23e3fd299cee0R197

richm: @nhosoi yes, we could probably get rid of `ME` in each shell script.  However, I didn't want to do this in this PR, since it will require some manual inspection to see if any of the custom scripts use `ME`, or did not source utils.sh.  We can do that in another PR.

.travis/config.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
tyll, line 1: This changes looks wrong. config.sh is not an executable AFAIU.

.yamllint_defaults.yml
```diff
@@ -10,5 +10,7 @@ rules:
   brackets:
     max-spaces-inside: 1
     level: error
-  truthy: disable
+  truthy:
+    allowed-values: ["yes", "no", "true", "false"]
```
tyll, line 14: IMHO this huge commit buries this important policy change. I remember that there was an agreement to avoid such huge commits. Also, I still prefer to have either yes/no or true/false as allowed values instead of both pairs.

.travis/config.sh
richm: > This changes looks wrong. config.sh is not an executable AFAIU.

shellcheck flags this - and the way shellcheck suppressions work it would be difficult to suppress this one.  I suppose I could add it to the global ignore list, but then we would have to ensure that a real executable script has the `#!/bin/bash`

.yamllint_defaults.yml
richm: > IMHO this huge commit buries this important policy change. I remember that there was an agreement to avoid such huge commits. Also, I still prefer to have either yes/no or true/false as allowed values instead of both pairs.

ok - well - I'm not sure how to cherry pick changes from the template when syncing changes to all of the system roles repos - suggestions welcome

richm: > I remember that there was an agreement to avoid such huge commits. 

Then please, please review the template changes, and let's please get the template changes into a form where *all* of the repo maintainers can agree that, once the change is committed to template, that same exact change can be synced to all of the repos - otherwise, this process just won't scale

.travis/config.sh
```diff
@@ -55,4 +48,17 @@ fi
...
+#       - RUN_SHELLCHECK_DISABLED
+#       - RUN_SHELLCHECK_EXTRA_ARGS
+type -f lsr_check_python_version > /dev/null 2>&1 || . ${SCRIPTDIR}/utils.sh
+if lsr_check_python_version python -lt 3.0
+then
+    PYTHON2_EXCLUDES="tests/ensure_provider_tests.py,scripts/print_all_options.py"
+    FLAKE8_DEFAULT_EXCLUDES=".svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg"
+    RUN_PYLINT_EXCLUDE='^(\..*|ensure_provider_tests\.py|print_all_options\.py)$'
+    RUN_FLAKE8_EXTRA_ARGS="--exclude ${FLAKE8_DEFAULT_EXCLUDES},${PYTHON2_EXCLUDES}"
+fi
```
tyll: This config code belongs to flake8/pylint but was moved to the shellcheck section in this PR.

.yamllint_defaults.yml
tyll: I added my comments about the values to https://github.com/linux-system-roles/template/pull/34#issuecomment-645402359.

Still, this is a squashed commit from the template instead of the individual commits originally used committed to the template and the commit message here does not even state that it will add/change anything related to yamllint/truthy.

> > IMHO this huge commit buries this important policy change. I remember that there was an agreement to avoid such huge commits. Also, I still prefer to have either yes/no or true/false as allowed values instead of both pairs.
> 
> ok - well - I'm not sure how to cherry pick changes from the template when syncing changes to all of the system roles repos - suggestions welcome

You can add the template repo as an additional remote to the working copies of the roles, then fetch and cherry pick commits from there. I did it the other way around when merging commits back from the network role to the template repo.

A probably cleaner solution would be to move all the template tools/code into its own python package that could be installed inside tox and do proper releases with release notes, there, then once something is done it can be released on pypi and would be picked up by the roles. Then the PRs in the roles would only contain the specific changes for the roles.

.travis/config.sh
tyll: What is the difficulty here? Does adding the "disable ..." comment instead of the shebang not help, here?

.travis/config.sh
richm: hmm - I was not aware that you were organizing the config.sh into "sections" . . .

.travis/config.sh
richm: it should be ok in this case - but in general, if the shellcheck directive is the first line in the file, or just after the shebang, it applies to the entire file - in this case, that is the desired behavior

richm: fixed in https://github.com/linux-system-roles/network/pull/261/commits/e2c5cd73388c6d75f3376f44aa5f2c799199c335

.yamllint_defaults.yml
richm: fixed in https://github.com/linux-system-roles/network/pull/261/commits/e2c5cd73388c6d75f3376f44aa5f2c799199c335 - added
```
rules:
    truthy: disable
```
to .yamllint.yml until such time as we can agree on what boolean format to enforce in system roles

.travis/config.sh
richm: fixed in https://github.com/linux-system-roles/network/pull/261/commits/e2c5cd73388c6d75f3376f44aa5f2c799199c335

.yamllint_defaults.yml
richm: > You can add the template repo as an additional remote to the working copies of the roles, then fetch and cherry pick commits from there.

Which is completely different than the current method of using sync-template.sh . . . which would require a significant rewrite of that script to implement such a procedure, or a completely new script.

Given this, and the other comments, I would prefer to move to a library based approach, possibly using a custom tox plugin as we have discussed in the past - something like https://github.com/greg-hellings/tox-ansible which https://github.com/oasis-roles uses - that would also increase our collaboration with the oasis group which we have been wanting to do

richm: So what I would like to do is to merge this PR as-is (or with some other minor modifications based on further review comments), then at some point in the near future implement a plugin based CI model.

richm:
ok - rebase and sign-off done
ok to merge?

tyll:
> > Could you please squash the "address review comments" commit in the respective other commits.
> 
> ok
> 
> > Also it would be nice to fix the issues first and enable shellcheck afterwards.
> 
> In this PR or in separate PRs?

In this PR is fine.

> Also, do you mean - have a commit that fixes all of the shellcheck issues first, in all of the files, including the files not modified by the network role that are synced from the template? Then a commit that enables the shellcheck test? I think that's the only way to do it if you want to both 1) have separate commits for fixing the shellcheck issues and enabling shellcheck 2) have each commit leave the repo in a state whereby you can run `tox` and all of the tests pass.
> 
> If so, my preference would be to do 2) but not 1). That is, have a single commit that both enables shellcheck and fixes all of the shellcheck issues.

If you think it is too much effort to separate adding shellcheck and fixing the issues, one commit can be fine, too. It seems that adding shellcheck consists of only three distinct changes (adding it to tox.ini, adding a new wrapper and updating the config.sh), so maybe it is feasible.

cathay4t:
@richm There is a conflict in `tox.ini`, can you rebase and I can continue the review work?

Thank you!

Issue #236, opened 2020-05-22
Cannot configure DEFROUTE in ifcfg files


on3iropolos:
If possible I would like to be able to configure the DEFROUTE within the configuration files opposed to manually naming every route. In my use case, I need to change the default interface to no longer have default route.

PR #263, opened 2020-08-11
Add defroute config option

xrow:
Will close https://github.com/linux-system-roles/network/issues/236

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -316,6 +316,11 @@ class IfcfgUtil:
 
         ifcfg["DEVICE"] = connection["interface_name"]
 
+        if connection["defroute"] == "yes":
+            ifcfg["DEFROUTE"] = "yes"
+        elif connection["defroute"] == "no":
+            ifcfg["DEFROUTE"] = "no"
+ 
         if connection["type"] == "ethernet":
             ifcfg["TYPE"] = "Ethernet"
             ifcfg["HWADDR"] = connection["mac"]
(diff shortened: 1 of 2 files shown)


tyll:
Thank you for your patch. There are a few remaining tasks:

- This does not seem to touch the NetworkManager provider code - is this not needed, there? @thom311 what's your opinion?
- Please document the new option in the README
- This breaks the unit tests, so they need to be adjusted as well
- There seems to be another, unrelated problem with the CI (see https://github.com/linux-system-roles/template/pull/40) just in case you are seeing a failure about this.

thom311:
- this option would belong under the `ip` group, where it is related to `gateway4`.

- also, in a similar way an option for IPv6 would be necessary.

- the option must also be implemented for NetworkManager

- the option is for the most part the inverse of NetworkManager's `ipv4.never-default`. `DEFROUTE=no` directly corresponds to `ipv4.never-default=yes`. If you configure `ipv4.never-default=no`/`DEFROUTE=yes` and `ipv4.gateway`/`GATEWAY` set, then the behavior is also clear. However, it's less clear if you don't specify a `gateway4`. I think, for initscripts `DEFROUTE=yes` only makes sense if you also set a `GATEWAY`. In NetworkManager, if you set `ipv4.never-default=no` and leave `ipv4.gateway` unset, it may still obtain the gateway from DHCP. So, this case differs between initscripts and NetworkManager, and possibly the role should reject setting DEFROUTE without GATEWAY as invalid configuration.

- what this option really does, is configuring a standard default route (`0.0.0.0/0`) on the interface. Note that you can achieve the same thing by adding a manual route to `0.0.0.0/0`. I think `ipv4.never-default` and `ipv4.gateway` (or their initscripts pendant) have no other significance. Note that NetworkManager before version 1.22.0, it wasn't supported to add `/0` routes in `ipv4.routes`, on those systems only the standard default route could be added (by `ipv4.never-default=no`).

jaedolph:
Since this PR had gone stale, I have created a new one that addresses this issue #309. Have also tried to address the review comments regarding NetworkManager.

cathay4t:
Closing as dup of #309

PR #264, opened 2020-08-11
Fix Travis CI issue with Docker server API version mismatch

richm:
Travis CI is using a version of Docker server that supports API version 1.38.  However, the Python PYPI package `docker` version 4.3 was recently released which uses Docker server API version 1.39, and there is apparently no way to tell molecule to configure the `docker` Python API to use a different API.  The solution is to install the latest 4.2.x version of the `docker` Python PYPI package.

### DIFF
--- a/.travis/runtox
+++ b/.travis/runtox
@@ -16,6 +16,24 @@ ME=$(basename $0)
 SCRIPTDIR=$(readlink -f $(dirname $0))
 BANNERSIZE=90
 
+# LSR_MOLECULE_DOCKER_VERSION is the version string to pass
+# to pip install - this is usually a comparison operator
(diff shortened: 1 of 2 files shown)


richm:
forgot to cherrypick the tox.ini change :P

richm:
> I am fine with merging this. Ideally, the commit message would use proper capitalization and punctuation and not repeat the topic and focus on the why (see https://github.com/linux-system-roles/network/blob/main/contributing.md#write-a-good-commit-message). Thanks.

@tyll Done.  Please review.

richm:
@tyll - please review

richm:
@tyll - please review

tyll:
[citest bad]

PR #268, opened 2020-08-20
tests: add simple bond tests for both nm and initscripts

tyll:
* add a helper to create two veth devices with IPv4/IPv6 infra
 * add a bond in active-backup mode with miimon value
 * add two slaves connections for two veth devices
 * check automated IPv4/IPv6 addresses are assigned to master
 * check all connections are present
 * add a helper to delete previously created veth device setup


tests/tasks/remove_test_interfaces_with_dhcp.yml
```diff
@@ -0,0 +1,25 @@
...
+    udevadm control --reload-rules
+    udevadm settle --timeout=5
+
+
+- name: Stop dnsmasq/radvd services
+  shell: |
+    pkill -F /run/dhcp_testbr.pid
+    rm -rf /run/dhcp_testbr.pid
+    rm -rf /run/dhcp_testbr.lease
+    if grep 'release 6' /etc/redhat-release; then
```
richm, line 20: Is this test run only on rhel and centos?

tests/tests_bond_nm.yml
```diff
@@ -0,0 +1,23 @@
...
+
+# workaround for: https://github.com/ansible/ansible/issues/27973
+# There is no way in Ansible to abort a playbook hosts with specific OS
+# releases Therefore we include the playbook with the tests only if the hosts
+# would support it.
+# The test requires or should run with NetworkManager, therefore it cannot run
+# on RHEL/CentOS 6
+- import_playbook: playbooks/tests_bond.yml
+  when:
+    - ansible_distribution_major_version != '6'
```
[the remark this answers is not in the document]
tyll: No, it also runs on Fedora and there is no version "6", therefore != `6` means Fedora or RHEL/CentOS != `6` in this context.

tests/tasks/remove_test_interfaces_with_dhcp.yml
tyll: No, but on Fedora the file does not exist, so the `else` block is used, there.

tests/tests_bond_nm.yml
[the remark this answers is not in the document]
richm: Ok.  So it implicitly skips this test on Fedora, rather than explicitly specifying that the test should be skipped on Fedora.
Is it possible that this role will someday support other platforms than EL and Fedora?

tests/tasks/remove_test_interfaces_with_dhcp.yml
richm: The file does exist on Fedora:
```
>cat /etc/fedora-release 
Fedora release 32 (Thirty Two)
>cat /etc/redhat-release 
Fedora release 32 (Thirty Two)
>ls -alrtF /etc/redhat-release
lrwxrwxrwx. 1 root root 14 Jun  5 14:00 /etc/redhat-release -> fedora-release
```
So t implicitly skips this test on Fedora, rather than explicitly specifying that the test should be skipped on Fedora.

tyll: ah, ok, so the file exists but does not contain `release 6` on Fedora as long as Fedora 60 is not released. The test is not skipped on Fedora, it will just use a different setup.

tests/tests_bond_nm.yml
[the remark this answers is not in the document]
tyll: It only skips it on CentOS/RHEL 6. I hope that by the time we support other plattforms that we do not need to support CentOS 6 anymore, so then this condition can just be removed.

tests/tests_bond_nm.yml
```diff
@@ -0,0 +1,23 @@
...
+      set_fact:
+        network_provider: nm
+      tags:
+        - always
+
+
+# workaround for: https://github.com/ansible/ansible/issues/27973
+# There is no way in Ansible to abort a playbook hosts with specific OS
+# releases Therefore we include the playbook with the tests only if the hosts
+# would support it.
```
pcahyna, line 18: FYI, Ansible 2.7 is not maintained anymore (https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-release-cycle) and since 2.8 there is a way to abort a host in the play. Of course, this code is found in many test playbooks so you may choose to keep it consistent and change it globally at some point.

PR #273, opened 2020-08-26
Add CHANGELOG

tyll:


CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
+# Changelog
+## [1.2.0] - 2020-08-26
+### Changes
+- Rename ethtool features to use underscores instead of dashes to support
```
richm, line 4: Doesn't this make it a breaking API change?  https://semver.org/
"increment the:
    MAJOR version when you make incompatible API changes,
"
so shouldn't the version be 2.0.0?

tyll: It still accepts the old notation.

pcahyna: That's great to hear, but shouldn't it be said in the changelog entry? This way it looks scary, like a compatibility break. What about "Rename ethtool features to use underscores instead of dashes to support Jinja2 dot notation, keeping the old names for compatibility"?

tyll: I would not use "keep" but changing it to:

```... Jinja2 dot notation. Accept old notation for compatibility with existing playbooks.```

Would you like to file a PR?

pcahyna: I agree with this wording. @cathay4t can you please update PR #282 ?

PR #274, opened 2020-08-26
Update formatting according to new black version

tyll:

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1032,7 +1032,8 @@ class NMUtil:
 
             if connection["ieee802_1x"]["ca_path"]:
                 s_8021x.set_property(
-                    NM.SETTING_802_1X_CA_PATH, connection["ieee802_1x"]["ca_path"],
+                    NM.SETTING_802_1X_CA_PATH,
+                    connection["ieee802_1x"]["ca_path"],
                 )
 
             s_8021x.set_property(
@@ -1912,7 +1913,7 @@ class Cmd(object):
         """ Hook for after all changes where made successfuly """
 
     def rollback_transaction(self, idx, action, error):
-        """ Hook if configuring a profile results in an error
+        """Hook if configuring a profile results in an error
 
         :param idx: Index of the connection that triggered the error
         :param action: Action that triggered the error
@@ -2019,7 +2020,7 @@ class Cmd_nm(Cmd):
                 self._checkpoint = None
 
     def _check_ethtool_setting_support(self, idx, connection):
-        """ Check if SettingEthtool support is needed and available
+        """Check if SettingEthtool support is needed and available
 
         If any feature is specified, the SettingEthtool setting needs to be
         available. Also NM needs to know about each specified setting. Do not
@@ -2236,7 +2237,7 @@ class Cmd_nm(Cmd):
                 self.log_error(idx, "up connection failed while waiting: %s" % (e))
 
     def _try_reapply(self, idx, con):
-        """ Try to reapply a connection
+        """Try to reapply a connection
 
         If there is exactly one active connection with the same UUID activated
         on exactly one device, ask the device to reapply the connection.
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -612,7 +612,8 @@ class ArgValidator_DictEthtoolFeatures(ArgValidatorDict):
                 ),
                 ArgValidatorBool("esp_tx_csum_hw_offload", default_value=None),
                 ArgValidatorDeprecated(
-                    "esp-tx-csum-hw-offload", deprecated_by="esp_tx_csum_hw_offload",
+                    "esp-tx-csum-hw-offload",
+                    deprecated_by="esp_tx_csum_hw_offload",
                 ),
                 ArgValidatorBool("fcoe_mtu", default_value=None),
                 ArgValidatorDeprecated("fcoe-mtu", deprecated_by="fcoe_mtu"),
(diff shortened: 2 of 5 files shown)


pcahyna:
I thought that the benefit of Black was to provide consistent formatting, but if it is a moving target, it sort of nullifies this benefit.

pcahyna:
what about freezing the version of Black that is used in the CI? We need to do it for ansible-lint anyway.

tyll:
> I thought that the benefit of Black was to provide consistent formatting, but if it is a moving target, it sort of nullifies this benefit.

It is mostly consistent but since it is still under development, there might be some changes sometimes. It still provides the benefit that there are no discussions about the formatting.

> what about freezing the version of Black that is used in the CI? We need to do it for ansible-lint anyway.

I believe this will make it more annoying for the developers if different projects use different versions. Also, it rarely happens that it changes. Also, for ansible-lint it is just a temporary workaround.

pcahyna:
> > what about freezing the version of Black that is used in the CI? We need to do it for ansible-lint anyway.
> 
> I believe this will make it more annoying for the developers if different projects use different versions. Also, it rarely happens that it changes. Also, for ansible-lint it is just a temporary workaround.

I believe it is a misunderstanding - IIUC from the discussion with @richm and @i386x we wanted to lock the version of ansible-lint permanently, so that it does not happen again that it suddenly starts complaining about something that used to be allowed and failing the CI runs. (The problem which triggered the discussion was not the bug that causes it to throw an exception, but the added check of role name which started failing on many roles.) Again, consistency and reproducibility of test results.

richm:
> > > what about freezing the version of Black that is used in the CI? We need to do it for ansible-lint anyway.
> > 
> > 
> > I believe this will make it more annoying for the developers if different projects use different versions. Also, it rarely happens that it changes. Also, for ansible-lint it is just a temporary workaround.
> 
> I believe it is a misunderstanding - IIUC from the discussion with @richm and @i386x we wanted to lock the version of ansible-lint permanently, so that it does not happen again that it suddenly starts complaining about something that used to be allowed and failing the CI runs. (The problem which triggered the discussion was not the bug that causes it to throw an exception, but the added check of role name which started failing on many roles.) Again, consistency and reproducibility of test results.

We would like to lock the version of everything so we don't run into problems like this in the future (see molecule v2 -> v3 upgrade previously which broke molecule testing).  The problem is then - how will we know if it is safe to upgrade?  We would need CI for the CI - run a special version of CI that has no locked versions of anything - if that passes, we know it is safe to upgrade to whatever version was pulled when the unlocked versions were used.

The ansible-lint issue is a special case - why is it suddenly complaining about rolenames like `linux-system-roles.network`?  Are we going to have to rename `linux-system-roles` to something without `-`?  We know that we will have to do that when we build the Collections package, but do we have to do that for legacy role support?  I would think not, so then the next question is - is it ok to use the latest version of ansible-lint, but just add a global suppression for that rolename item?

As far as this particular instance of black - I think it is ok for the network role to update the formatting to conform to black, and not to lock the version of black.  We don't have a lot of roles which will be changing in the near future, so I don't expect them to get dinged by this black update issue.  If we do, we can evaluate if it will be expensive to reformat the python files - it is usually as easy as running black without the `--check --diff` arguments we use in tox.  Worst case, we can disable black for a particular repo using .travis/config.sh

tyll:
> We would like to lock the version of everything so we don't run into problems like this in the future (see molecule v2 -> v3 upgrade previously which broke molecule testing). The problem is then - how will we know if it is safe to upgrade? We would need CI for the CI - run a special version of CI that has no locked versions of anything - if that passes, we know it is safe to upgrade to whatever version was pulled when the unlocked versions were used.

To me it seems that it is less effort to fix/lock this occasionally than to actively monitor and test all releases. @cathay4t @ffmancera what is your preference?

pcahyna:
> The ansible-lint issue is a special case - why is it suddenly complaining about rolenames like linux-system-roles.network?

I don't think it is so special - issues with linters that will start complaining about something that used to be allowed will likely come up again and again.

> We don't have a lot of roles which will be changing in the near future, so I don't expect them to get dinged by this black update issue.

Doesn't the change in Black mean that every role that has any Python file, uses Black and there is Python code that is formatted differently using the new Black version will now fail in Travis CI? Maybe I misunderstood the impact.

richm:
> > The ansible-lint issue is a special case - why is it suddenly complaining about rolenames like linux-system-roles.network?
> 
> I don't think it is so special - issues with linters that will start complaining about something that used to be allowed will likely come up again and again.
> 
> > We don't have a lot of roles which will be changing in the near future, so I don't expect them to get dinged by this black update issue.
> 
> Doesn't the change in Black mean that every role that has any Python file, uses Black and there is Python code that is formatted differently using the new Black version will now fail in Travis CI?

Yes, any role with python code will potentially fail Travis CI now.  But 1) except for network and storage, I don't think any of the roles that have python code will be changing any time soon 2) it may be that black doesn't complain because the role has formatted the code already to satisfy both the old and the new black formatting 3) storage has disabled black - .travis/config.sh has RUN_BLACK_DISABLED=true 4) it is pretty easy to have black automatically format the code

So we can probably handle these on a case-by-case basis as the need arises

> Maybe I misunderstood the impact.

tyll:
> > We don't have a lot of roles which will be changing in the near future, so I don't expect them to get dinged by this black update issue.
> 
> Doesn't the change in Black mean that every role that has any Python file, uses Black and there is Python code that is formatted differently using the new Black version will now fail in Travis CI? Maybe I misunderstood the impact.

This depends on the change. If roles do not use the construct black is changing, they won't be affected.

pcahyna:
> > > We don't have a lot of roles which will be changing in the near future, so I don't expect them to get dinged by this black update issue.
> > 
> > 
> > Doesn't the change in Black mean that every role that has any Python file, uses Black and there is Python code that is formatted differently using the new Black version will now fail in Travis CI? Maybe I misunderstood the impact.
> 
> This depends on the change. If roles do not use the construct black is changing, they won't be affected.

Sure. That's what I meant by "there is Python code that is formatted differently using the new Black version".

tyll:
[citest bad]

PR #279, opened 2020-09-01
Lock ansible-lint on version 4.2.0

i386x:
`ansible-lint` 4.2.0 is latest known version of `ansible-lint` that does not break CI testing with `tox` & Travis

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -108,6 +108,7 @@ commands =
 [molecule_common]
 changedir = {toxinidir}
 deps =
+    ansible-lint==4.2.0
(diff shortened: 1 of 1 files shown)


tyll:
FYI: It is already pinned in: https://github.com/linux-system-roles/network/blob/main/molecule_extra_requirements.txt

i386x:
> FYI: It is already pinned in: https://github.com/linux-system-roles/network/blob/main/molecule_extra_requirements.txt

Ah, didn't notice this, thank you. So this PR is not needed then.

PR #282, opened 2020-09-08
CHANGELOG: emphasize the backward compatibility

cathay4t:


tyll (approved):

pcahyna (approved):
lgtm - but something is wrong with some tests. The citool-staging failues can be infrastructure problems, but for the others, something else must be wrong. Related might be #286.

tyll:
@cathay4t seems that this needs to be rebased.

cathay4t:
> @cathay4t seems that this needs to be rebased.

Done.

tyll:
ansible 2.8 tests were disabled in the CI, no need to wait for them anymore.

pcahyna:
> ansible 2.8 tests were disabled in the CI, no need to wait for them anymore.

@tyll This is not entirely true. The lack of results from ansible-2.8-centos7 runs in this PR was due to a combination of two factors:
1. the number of workers with this configuration was scaled down to one staging + one production
1. it happens fairly often that for some unknown reason the testing VM freezes (not just on the network PRs), which effectively disables the worker.

I've now scaled the number of workers with the production ansible-2.8 configuration up, so the throughput of testing should be better. Until the second problem gets resolved, this will only delay the inevitable, though.

PR #295, opened 2020-10-05
Remove workaround for host selection in ensure_provider_tests.py

liangwen12year:

### DIFF
--- a/tests/ensure_provider_tests.py
+++ b/tests/ensure_provider_tests.py
@@ -47,10 +47,6 @@ RUN_PLAYBOOK_WITH_NM = """# SPDX-License-Identifier: BSD-3-Clause
         - always
 {get_nm_version}
 
-# workaround for: https://github.com/ansible/ansible/issues/27973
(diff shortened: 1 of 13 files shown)


tyll:
Thank you. Please check the CI results.

liangwen12year:
> Thank you. Please check the CI results.
checked and updated the commit.

tyll:
The commit message mentions "Remove workaround for host selection in ensure_provider_tests.py" but this does not remove the workaround but the comment that about it being a workaround but it is not considered a workaround anymore.

liangwen12year:
> The commit message mentions "Remove workaround for host selection in ensure_provider_tests.py" but this does not remove the workaround but the comment that about it being a workaround but it is not considered a workaround anymore.

updated the commit message.

ffmancera (changes requested):
I would like to suggest `Remove comment about host selection workaround` as a commit message, what do you think?

ffmancera (approved):
LGTM, thanks!

PR #297, opened 2020-10-14
lock ansible-lint version at 4.3.5; suppress role name lint warning

richm:
lock the ansible-lint version to use the latest version 4.3.5 which
is compatible with the latest ansible 2.10
This means we need to suppress the error about the role name


.ansible-lint
```diff
@@ -1,3 +1,4 @@
 ---
 skip_list:
 - '602'
+- '106'
```
ffmancera, line 4: I assume it is not possible to fix the error about the role name, right?

richm: right

ffmancera (approved):
LGTM, thanks!

PR #298, opened 2020-10-24
nm provider: Refactor the down action of network connection

cathay4t:
When deactivating a profile in libNM, we should:

 * Check `NM.ActionConnection` existence
 * Check `NM.ActionConnection.props.state` not DEACTIVATED
 * Use signal `state-changed` of `NM.ActionConnection`.
 * Only invoke `NM.Client.deactivate_connection_async()` if not
   in DEACTIVATING state.
 * Ignore `NM.ManagerError.CONNECTIONNOTACTIVE` error.

This patch also introduced a new class `NetworkManagerProvider`
in `module_utils/network_nm_lsr`:

 * Independent from Ansible but need to use absolute import due to limitation of ansible 2.8.
 * Provide sync function wrapping async calls of libNM.
 * Use stable logging method of python.
 * Only load this module when provider is nm.

This patch also changed how logging is handling in
`Cmd_nm.run_action_down()` as initial step on isolate ansible log
mechanism from provider module.

By moving provider codes to `module_utils` folder, we can eventually
simplify the bloated `library/network_connections.py`.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -11,6 +11,7 @@ import socket
 import subprocess
 import time
 import traceback
+import logging
 
 # pylint: disable=import-error, no-name-in-module
 from ansible.module_utils.basic import AnsibleModule
@@ -65,6 +66,17 @@ class LogLevel:
     WARN = "warn"
     INFO = "info"
     DEBUG = "debug"
+
+    _LOGGING_LEVEL_MAP = {
+        logging.DEBUG: DEBUG,
+        logging.INFO: INFO,
+        logging.WARN: WARN,
+        logging.ERROR: ERROR,
+    }
+
+    @staticmethod
+    def from_logging_level(logging_level):
+        return LogLevel._LOGGING_LEVEL_MAP.get(logging_level, LogLevel.ERROR)
 
     @staticmethod
     def fmt(level):
@@ -1387,61 +1399,6 @@ class NMUtil:
         if failure_reason:
             raise MyError("connection not activated: %s" % (failure_reason))
 
-    def active_connection_deactivate(self, ac, timeout=10, wait_time=None):
-        def deactivate_cb(client, result, cb_args):
-            success = False
-            try:
-                success = client.deactivate_connection_finish(result)
-            except Exception as e:
-                if Util.error_is_cancelled(e):
-                    return
-                cb_args["error"] = str(e)
-            cb_args["success"] = success
-            Util.GMainLoop().quit()
-
-        cancellable = Util.create_cancellable()
-        cb_args = {}
-        self.nmclient.deactivate_connection_async(
-            ac, cancellable, deactivate_cb, cb_args
-        )
-        if not Util.GMainLoop_run(timeout):
-            cancellable.cancel()
-            raise MyError("failure to deactivate connection: %s" % (timeout))
-        if not cb_args.get("success", False):
-            raise MyError(
-                "failure to deactivate connection: %s"
-                % (cb_args.get("error", "unknown error"))
-            )
-
-        self.active_connection_deactivate_wait(ac, wait_time)
-        return True
-
-    def active_connection_deactivate_wait(self, ac, wait_time):
-
-        if not wait_time:
-            return
-
-        NM = Util.NM()
-
-        def check_deactivated(ac):
-            return ac.get_state() >= NM.ActiveConnectionState.DEACTIVATED
-
-        if not check_deactivated(ac):
-
-            def check_deactivated_cb():
-                if check_deactivated(ac):
-                    Util.GMainLoop().quit()
-
-            ac_id = ac.connect(
-                "notify::state", lambda source, pspec: check_deactivated_cb()
-            )
-
-            try:
-                if not Util.GMainLoop_run(wait_time):
-                    raise MyError("connection not fully deactivated after timeout")
-            finally:
-                ac.handler_disconnect(ac_id)
-
     def reapply(self, device, connection=None):
         version_id = 0
         flags = 0
@@ -1627,6 +1584,21 @@ class RunEnvironmentAnsible(RunEnvironment):
 
 
 ###############################################################################
+
+
+class NmLogHandler(logging.Handler):
+    def __init__(self, log_func, idx):
+        self._log = log_func
+        self._idx = idx
+        super(NmLogHandler, self).__init__()
+
+    def filter(self, record):
+        return True
+
+    def emit(self, record):
+        self._log(
+            self._idx, LogLevel.from_logging_level(record.levelno), record.getMessage()
+        )
 
 
 class Cmd(object):
@@ -1954,6 +1926,14 @@ class Cmd_nm(Cmd):
         self._nmutil = None
         self.validate_one_type = ArgValidator_ListConnections.VALIDATE_ONE_MODE_NM
         self._checkpoint = None
+        # pylint: disable=import-error, no-name-in-module
+        from ansible.module_utils.network_lsr.nm import (  # noqa: E501
+            NetworkManagerProvider,
+        )
+
+        # pylint: enable=import-error, no-name-in-module
+
+        self._nm_provider = NetworkManagerProvider()
 
     @property
     def nmutil(self):
@@ -2265,51 +2245,17 @@ class Cmd_nm(Cmd):
 
     def run_action_down(self, idx):
         connection = self.connections[idx]
-
-        cons = self.nmutil.connection_list(name=connection["name"])
-        changed = False
-        if cons:
-            seen = set()
-            while True:
-                ac = Util.first(
-                    self.nmutil.active_connection_list(
-                        connections=cons, black_list=seen
-                    )
-                )
-                if ac is None:
-                    break
-                seen.add(ac)
-                self.log_info(
-                    idx, "down connection %s: %s" % (connection["name"], ac.get_path())
-                )
-                changed = True
-                self.connections_data_set_changed(idx)
-                if self.check_mode == CheckMode.REAL_RUN:
-                    try:
-                        self.nmutil.active_connection_deactivate(ac)
-                    except MyError as e:
-                        self.log_error(idx, "down connection failed: %s" % (e))
-
-                    wait_time = connection["wait"]
-                    if wait_time is None:
-                        wait_time = 10
-
-                    try:
-                        self.nmutil.active_connection_deactivate_wait(ac, wait_time)
-                    except MyError as e:
-                        self.log_error(
-                            idx, "down connection failed while waiting: %s" % (e)
-                        )
-
-                cons = self.nmutil.connection_list(name=connection["name"])
-        if not changed:
-            message = "down connection %s failed: connection not found" % (
-                connection["name"]
-            )
-            if connection[PERSISTENT_STATE] == ABSENT_STATE:
-                self.log_info(idx, message)
-            else:
-                self.log_error(idx, message)
+        logger = logging.getLogger()
+        log_handler = NmLogHandler(self.log, idx)
+        logger.addHandler(log_handler)
+        timeout = connection["wait"]
+        if self._nm_provider.deactivate_connection(
+            connection["name"],
+            10 if timeout is None else timeout,
+            self.check_mode != CheckMode.REAL_RUN,
+        ):
+            self.connections_data_set_changed(idx)
+        logger.removeHandler(log_handler)
 
 
 ###############################################################################
--- a/module_utils/network_lsr/nm/active_connection.py
+++ b/module_utils/network_lsr/nm/active_connection.py
@@ -1,0 +1,125 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Handle NM.ActiveConnection
+
+import logging
+
+# Relative import is not support by ansible 2.8 yet
+# pylint: disable=import-error, no-name-in-module
+from ansible.module_utils.network_lsr.nm.client import GLib  # noqa:E501
+from ansible.module_utils.network_lsr.nm.client import NM  # noqa:E501
+from ansible.module_utils.network_lsr.nm.client import get_mainloop  # noqa:E501
+from ansible.module_utils.network_lsr.nm.client import get_client  # noqa:E501
+from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+
+# pylint: enable=import-error, no-name-in-module
+
+
+NM_AC_STATE_CHANGED_SIGNAL = "state-changed"
+
+
+def deactivate_active_connection(nm_ac, timeout, check_mode):
+    if not nm_ac or nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
+        logging.info("Connection is not active, no need to deactivate")
+        return False
+    if not check_mode:
+        main_loop = get_mainloop(timeout)
+        logging.debug(
+            "Deactivating {id} with timeout {timeout}".format(
+                id=nm_ac.get_id(), timeout=timeout
+            )
+        )
+        user_data = main_loop
+        handler_id = nm_ac.connect(
+            NM_AC_STATE_CHANGED_SIGNAL, _nm_ac_state_change_callback, user_data
+        )
+        logging.debug(
+            "Registered {signal} on NM.ActiveConnection {id}".format(
+                signal=NM_AC_STATE_CHANGED_SIGNAL, id=nm_ac.get_id()
+            )
+        )
+        if nm_ac.props.state != NM.ActiveConnectionState.DEACTIVATING:
+            nm_client = get_client()
+            user_data = (main_loop, nm_ac, nm_ac.get_id(), handler_id)
+            nm_client.deactivate_connection_async(
+                nm_ac,
+                main_loop.cancellable,
+                _nm_ac_deactivate_call_back,
+                user_data,
+            )
+            logging.debug("Deactivating NM.ActiveConnection {0}".format(nm_ac.get_id()))
+        main_loop.run()
+    return True
+
+
+def _nm_ac_state_change_callback(nm_ac, state, reason, user_data):
+    main_loop = user_data
+    if main_loop.is_cancelled:
+        return
+    logging.debug(
+        "Got NM.ActiveConnection state change: {id}: {state} {reason}".format(
+            id=nm_ac.get_id(), state=state, reason=reason
+        )
+    )
+    if nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
+        logging.debug("NM.ActiveConnection {0} is deactivated".format(nm_ac.get_id()))
+        main_loop.quit()
+
+
+def _nm_ac_deactivate_call_back(nm_client, result, user_data):
+    main_loop, nm_ac, nm_ac_id, handler_id = user_data
+    logging.debug("NM.ActiveConnection deactivating callback")
+    if main_loop.is_cancelled:
+        if nm_ac:
+            nm_ac.handler_disconnect(handler_id)
+        return
+
+    try:
+        success = nm_client.deactivate_connection_finish(result)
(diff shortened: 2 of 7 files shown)


cathay4t:
The CI failures are caused by:

 * Fedora 32 on ansible 2.8: ansible/ansible#71528
 * Fedora 32 cannot start wpa_supplicant service: ansible/ansible#69302
 * RHEL cannot remove a connection: https://github.com/linux-system-roles/network/pull/302

contributing.md
```diff
@@ -31,6 +31,9 @@ The repository is structured as follows:
 
 - `./module_utils/network_lsr/` - Contains other files that are useful for the network
   role (e.g. the YAML argument validator)
+ 
+- `./module_utils/network_nm_lsr/` - Contains other files that are useful for the network
```
tyll, line 35: It seems to me to be a lot less complex to use `./module_utils/network_lsr/nm_provider/` or some other subdirectory of `network_lsr`. Then you don't need to duplicate the extra work to get the path added to multiple places. What do you think?

library/network_connections.py
```diff
@@ -1629,6 +1586,21 @@ def fail_json(
 ###############################################################################
 
 
+class NmLsrLogHanlder(logging.Handler):
```
tyll, line 1589: There is a typo. Also, why `Lsr` here? This is all about the linux system roles, so "Lsr" is implied by the context.

module_utils/network_nm_lsr/client.py
```diff
@@ -0,0 +1,84 @@
...
+# Relative import is not support by ansible 2.8 yet
+# pylint: disable=import-error, no-name-in-module
+from ansible.module_utils.network_nm_lsr.error import LsrNetworkNmError  # noqa:E501
+
+import gi
+
+gi.require_version("NM", "1.0")
+
+# It is required to state the NM version before importing it
+# But this break the flak8 rule: https://www.flake8rules.com/rules/E402.html
```
tyll, line 14:
```suggestion
# But this break the flake8 rule: https://www.flake8rules.com/rules/E402.html
```

module_utils/network_nm_lsr/provider.py
```diff
@@ -0,0 +1,29 @@
...
+        Return True if changed.
+        """
+        nm_client = get_client()
+        changed = False
+        for nm_ac in nm_client.get_active_connections():
+            nm_profile = nm_ac.get_connection()
+            if nm_profile and nm_profile.get_id() == connection_name:
+                changed |= deactivate_active_connection(nm_ac, timeout, check_mode)
+        if not changed:
+            logging.info("No acticative connection for {0}".format(connection_name))
```
tyll, line 27: Do you mean active here?

```suggestion
            logging.info("No active connection for {0}".format(connection_name))
```

contributing.md
cathay4t: Ansible 2.8 does not allow folder in `network_lsr`.

thom311:
Hi Gris,

You say,

> When deactivating a profile in libNM, we should:
>
>    * Check NM.ActionConnection existence
>    * Check NM.ActionConnection.props.state not DEACTIVATED
>    * Use signal state-changed of NM.ActionConnection.
>    * Only invoke NM.Client.deactivate_connection_async() if not
>   in DEACTIVATING state.
>    * Ignore NM.ManagerError.CONNECTIONNOTACTIVE error.

and you also said (on IRC) "The bad design is the user of libNM don't know exactly the correct way to deactivate a connection."

Deactivating a connection in NetworkManager D-Bus API (or libnm) doesn't strike me as that bad designed. An "activation" is represented by a D-Bus object like `/org/freedesktop/NetworkManager/ActiveConnection/1` (in libnm there is a corresponding `NMActiveConnection` instance in the cache). This object has a state, and the state progression is simple and one-way: from NEW -> ACTIVATING -> ACTIVATED -> DEACTIVATING -> DEACTIVATED. The active connection instance doesn't go back, the state only goes forward, that makes it in my opinion simple to use.

the points you make about what should be done are probably right, but they are direct conclusions from an API that I would expect to behave this way. The previous code didn't do it much different, with some minor variations (where it's hard to argue which one is more correct).

Also, apparently this merge requests fixes https://bugzilla.redhat.com/show_bug.cgi?id=1817242. And in https://bugzilla.redhat.com/show_bug.cgi?id=1817242#c9 you identified some points what the current code does wrong. But does it?  Indeed, `active_connection_deactivate()` should probably ignore certain errors. Fixing that is simple. The other two points seem questionable to me.


For example, about checking the cancellable: All async operations in glib (and libnm) that accept a `GCancellable` will fail with `Util.error_is_cancelled(e)` if (and only if) the associated cancellable got cancelled. Tracking the cancellable separately and checking it explicitly seems unnecessary:

```
+    if main_loop.is_cancelled:
+        if nm_ac:
+            nm_ac.handler_disconnect(handler_id)
+        return
+
+    try:
+        success = nm_client.deactivate_connection_finish(result)
+    except GLib.Error as e:
```
This seems to complicate things, and would also be racy if the cancellable could be cancelled from another thread (which isn't the case here).


It's not clear how "state-change" and "notify:state" differ. I don't think they do. If there is a difference, we should fix that, it would be a bug. In any case, the fix for that would be to subscribe to the other signal (or as workaround for a bug both).




> By moving provider codes to module_utils folder, we can eventually
> simplify the bloated library/network_connections.py.

Bloated is a rather heated word. Often that means that the code does something that should not be it's responsibility. Better not having a feature than to make code larger and more complicated. This refactoring hardly changes the steps that the role does here (it doesn't do "less", nor is it clear what the bloat was to begin with). If you mean by "bloated" that the file is too large, ok, that is an opinion. Now you split the code into multiple files and introduce new layers that were not there (nor needed) before. Also, the code to deactivate a connection was before 99 LOC, and, it was all at one place. I guess, you are of the opinion that the code got simpler with this patch (less bloated?). I don't see that. It is harder to follow with the new layers.

For example, does the use of the "logger" module really simplify things? Well, I guess you need that now, because the new layers also need to the logging. But how does it "unbloat" the library or simplify it?

Also, the code seems to go out its way to abstract glib (with new layers like `NmMainLoop`), at places where it doesn't seem good.

I am not particularly against this change, but let's keep the arguments accurate. Your stated [goal](https://bugzilla.redhat.com/show_bug.cgi?id=1817242#c9) is to "copy/import mature code from nmstate to network-role". This is backed by arguments (the three points) that are in my opinion not accurate and/or easily fixable.

Using words like "mature code" (contrary to what was before, I guess), is a questionable claim that is not backed by facts.


---

Fox example, 
```
+        for nm_ac in nm_client.get_active_connections():
+            nm_profile = nm_ac.get_connection()
+            if nm_profile and nm_profile.get_id() == connection_name:
+                changed |= deactivate_active_connection(nm_ac, timeout, check_mode)
```
This changes the logic from before. Before, the code in `run_action_down()` keep iterating over active connections (that were for the connection in question), and tried to bring all the active connections down. Now, the code iterates over all active connections (once), and when it finds one that matches by name, it calls deactivate.  Previously, the loop would keep deactivating suitable active-connections until no more candidates where active. Granted, that's in most cases not necessary (it does so with the goal of being robust against unexpected things). But it's not clear why the new approach is better or preferable. The previous code approach was chosen deliberately. The reasons for doing it that way were not very strong and it can be changed. But then there should be a justification why the algorithm changes, beyond claims of mature code.


It is one thing to add a new "provider=nmstate". But why does the existing "nm" provider need to be refactored and these new layers introduced? I guess, the reason is that you want the code to be similar to nmstate. Then the commit message should state that as the main purpose of this patch.

contributing.md
richm: > Ansible 2.8 does not allow folder in `network_lsr`.

How so?  I'll note that when I install ansible 2.8 with pip, the module_utils directories have sub-directories e.g. `module_utils/storage/emc/` - I don't know if this is "allowed" or if you can do `import ansible.module_utils.storage.emc.emc_vnx` and it "works", but it is there in the ansible 2.8 package.

cathay4t:
The nmstate provider will not support RHEL 7 and RHEL 8.0/8.1, so we still need this nm provider for years.

The good design is user don't need to know libnm is based on DBUS and all other complex internal stuff.
But I will save my strength on debating this. I have accepted the truth that NM will keep doing whatever they did.

The `mature code` I mean here is the code/workflow well tested in nmstate.

The reason for this PR:
 * Clean up the mess on 2700+ lines of `library/network_connections.py`.
 * Mimic the workflow used by nmstate.
 * Fix the CI failure https://bugzilla.redhat.com/show_bug.cgi?id=1817242

contributing.md
cathay4t: I tried on RHEL 8 using ansible 2.8. The `explode` command shows the sub-folder is not send to remote host.

cathay4t: Yeah. with `ansible.module_utils.storage.emc.emc_vnx`, the sub-folder works. Let me ammend this PR.

cathay4t:
New patch uploaded.

Changes:
 * Fixed the typos.
 * Moved the code back to folder `module_utils/network_lsr/nm`.

contributing.md
cathay4t: Thanks for the hint. New patch uploaded.

library/network_connections.py
```diff
@@ -2265,51 +2245,17 @@ def _try_reapply(self, idx, con):
...
-            message = "down connection %s failed: connection not found" % (
-                connection["name"]
-            )
-            if connection[PERSISTENT_STATE] == ABSENT_STATE:
-                self.log_info(idx, message)
-            else:
-                self.log_error(idx, message)
+        logger = logging.getLogger()
+        log_handler = NmLogHandler(self.log, idx)
+        logger.addHandler(log_handler)
```
thom311, line 2250: maybe there should be a `with self.logger(idx)` which returns a context manager for the logger.

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -0,0 +1,127 @@
...
+        main_loop.quit()
+
+
+def _nm_ac_deactivate_call_back(nm_client, result, user_data):
+    main_loop, nm_ac, handler_id = user_data
+    logging.debug("NM.ActiveConnection deactivating callback")
+    if main_loop.is_cancelled:
+        if nm_ac:
+            nm_ac.handler_disconnect(handler_id)
+        return
```
thom311, line 74: it seems unnecessary to check whether the `main_loop` is cancelled. All glib functions will return a specific error if (and only if) the caller cancelled the provided cancellable.

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -0,0 +1,127 @@
...
+            logging.info(
+                "Connection is not active on {0}, no need to deactivate".format(
+                    nm_ac.get_id()
+                )
+            )
+            if nm_ac:
+                nm_ac.handler_disconnect(handler_id)
+            main_loop.quit()
+        else:
+            _deactivate_fail(
```
thom311, line 90: there is a race here. The client calls `deactivate_connection_async()` (after checking that the active connection is still <= ACTIVATED), but still, at that very moment (on the server side) the ActiveConnection could go down and disappear entirely. Then it wouldn't merely fail with `NM.ManagerError.CONNECTIONNOTACTIVE`, but with some entirely different message like that the D-Bus path does not refer to an existing object.

(of course, it's a race and maybe hard to reproduce)

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -0,0 +1,127 @@
...
+            main_loop,
+            handler_id,
+            nm_ac,
+            "Failed to deactivate connection {id}, error={error}".format(
+                id=nm_ac.get_id(), error=e
+            ),
+        )
+
+    if success:
+        logging.debug("NM.ActiveConnection {0} deactivated".format(nm_ac.get_id()))
```
thom311, line 109: when the callback returns, in fact NetworkManager makes an effort to ensure that the active connection already changed state to DEACTIVATING state (or later). But it seems unnecessary to rely on that, and the previous code didn't do that. Previously, the `deactivate_connection_async()` only initiated the deactivation, and then the code would check whether the DEACTIVATED state was reached (with a timeout). Maybe it would be more robust, if a success from `deactivate_connection_finish()` does not yet automatically assume that the DEACTIVATED state was reached.

module_utils/network_lsr/nm/client.py
```diff
@@ -0,0 +1,84 @@
...
+        self._mainloop = None
+
+    def quit(self):
+        logging.debug("NM mainloop quiting")
+        self._mainloop.quit()
+        self._clean_up()
+
+    def fail(self, exception):
+        self.quit()
+        raise exception
```
thom311, line 80: I am surprised this works. `fail` is commonly called while running the mainloop, so how does the python exception get propagated through the C code (in glib, which runs the mainloop)? According to https://stackoverflow.com/questions/54666827/pygobject-glib-mainloop-and-exceptions, it would seem that this does not work. But obviously, it does somehow.

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+    )
+    if nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
+        logging.debug("NM.ActiveConnection {0} is deactivated".format(nm_ac.get_id()))
+        main_loop.quit()
+
+
+def _nm_ac_deactivate_call_back(nm_client, result, user_data):
+    main_loop, nm_ac, handler_id = user_data
+    logging.debug("NM.ActiveConnection deactivating callback")
```
thom311, line 70: since you call `main_loop.quit()` from `_nm_ac_state_change_callback()`, it might be possible that this callback was invoked only the next time when the mainloop gets iterated.

Also, next you call `main_loop.is_canclled`, which will raise an exception if the mainloop was already `_clean_up()`.

All this is because async calls in glib (and libnm) will *always* invoke their result callback on the associated `GMainContext`. Of course, if they are cancelled, they will indicate that with an error `Util.error_is_cancelled(e)`, but it will be invoked. So, the solution is to not touch anything else before checking whether the callback was cancelled (unless you are sure, that there are no pending callback to be queued on the associated GMainContext (the mainloop) -- but that is often not that easy to ensure, and the current code does not ensure that).

It seems as a general pattern with glib's async request to always invoke `nm_client.deactivate_connection_finish(result)` as very first thing in a callback, and check whether the resulting error indicates cancellation (in which case not to do anything).

library/network_connections.py
```diff
@@ -1954,6 +1926,14 @@ def __init__(self, **kwargs):
         self._nmutil = None
         self.validate_one_type = ArgValidator_ListConnections.VALIDATE_ONE_MODE_NM
         self._checkpoint = None
+        # pylint: disable=import-error, no-name-in-module
+        from ansible.module_utils.network_lsr.nm import (  # noqa: E501
```
richm, line 1930: @nhosoi does the role2collection script handle this case?

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def deactivate_active_connection(nm_ac, timeout, check_mode):
+    if not nm_ac or nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
+        logging.info("Connection is not active, no need to deactivate")
+        return False
+    if not check_mode:
+        main_loop = get_mainloop(timeout)
+        logging.debug(
+            "Deactiving {id} with timeout {timeout}".format(
```
richm, line 28: Deactivating

library/network_connections.py
nhosoi: Once this [PR](https://github.com/linux-system-roles/auto-maintenance/pull/26) is merged, yes.

module_utils/network_lsr/nm/active_connection.py
cathay4t: An error is not accepted just because the action is canceled.

module_utils/network_lsr/nm/active_connection.py
cathay4t: If you are referring the race of `nm_ac.get_id()`, I have fixed it by saving the id in the user data.

For other race issue on this `some entirely different message`, never seen in nmstate 1000+ test runs. I don't know the detail and why it could failed, so I will live with it.

module_utils/network_lsr/nm/active_connection.py
cathay4t: Yes. That's correct. I have removed the `main_loop.quit()` when `success` returned by `deactivate_connection_finish()` and wait state change signal callback.

module_utils/network_lsr/nm/client.py
cathay4t: The `raise` is ran after mainloop quit. So the exception is not handlled in side mainloop.

module_utils/network_lsr/nm/active_connection.py
cathay4t: Changed to:

```python
    @property
    def is_cancelled(self):
        if self._cancellable:
            return self._cancellable.is_cancelled()
        return True
```

module_utils/network_lsr/nm/client.py
thom311: that's not what `g_main_loop_quit()` does. `GMainLoop` is basically a boolean flag (`g_main_loop_is_running()`) and two functions:

  - `g_main_loop_run()` sets is-running flag to true, and does `while (is_running) iterate-main-context` ([[1]](https://gitlab.gnome.org/GNOME/glib/-/blob/b8927cc6ad866570b100aa71e5aeabd349f439f3/glib/gmain.c#L4316)).
  - `g_main_loop_quit()` sets the is-flag to false. Nothing else ([[2]](https://gitlab.gnome.org/GNOME/glib/-/blob/b8927cc6ad866570b100aa71e5aeabd349f439f3/glib/gmain.c#L4343)).

`g_main_loop_quit()`  cannot jump to the call-stack of where `g_main_loop_run()` is invoked. It clears the `is_running` flag. Consequently, the calls after `g_main_loop_quit()` call still are invoked in the same frame of the call stack.

thom311: > So the exception is not handlled in side mainloop.

Hm, guess it depends on what you mean with "in side mainloop". I would say that code executes "inside the mainloop" if the current call-stack has `g_main_context_iterate()` in the parent frames. That means, the current code is invoked as part of a callback function to a `GSource` which the `GMainContext` is currently dispatching. With that definition, the `raise` certainly happens inside the mainloop.

there is also a warning in the docs for `g_main_loop_quit()`:
```
 * Note that sources that have already been dispatched when 
 * g_main_loop_quit() is called will still be executed.
```

that means, after a call to quit(), not only does the current call-stack come to it's normal end (there is no jump/exception), even more sources might be dispatched (if they were already ready) before `g_main_loop_run()` returns.

cathay4t:
New patch uploaded.

Changes:
 * The success of `deactivate_connection_finish` will not quit the mainloop, we will continue the wait of `NM.ActiveConnection` state change.
 * Fix the race problem of accessing `nm_ac.get_id()` when nm_ac might be None.

cathay4t:
@thom311 Could you take a review again on new patch? Thank you!

module_utils/network_lsr/nm/active_connection.py
cathay4t: Fixed. Thanks.

module_utils/network_lsr/nm/client.py
cathay4t: If you have suggestions on how to raise exception found in mainllop, please state a demo code.
If you have questions regarding why this code works, I would suggest reach out experts(not me) somewhere else.

For me, the code works as expected. And I cannot switch to `NM.Client.get_main_context()` as RHEL 7 still hold very old NetworkManager.

thom311:
lgtm

ffmancera (approved):

PR #302, opened 2020-10-26
nm provider: Refactor the volatilize action of network connection

cathay4t:
Refactor the volatilize action of nm provider:
 * Move code to `module_utils/network_lsr/nm`
 * The `module_utils/network_lsr/nm` only delete profile by given UUID
   instead of guess. The `library/network_connections.py` is responsible
   on choosing UUID.

### DIFF
--- a/module_utils/network_lsr/nm/connection.py
+++ b/module_utils/network_lsr/nm/connection.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Handle NM.RemoteConnection
+
+import logging
+
+# Relative import is not support by ansible 2.8 yet
+# pylint: disable=import-error, no-name-in-module
+from ansible.module_utils.network_lsr.nm import client  # noqa:E501
+from ansible.module_utils.network_lsr.nm import error  # noqa:E501
+
+# pylint: enable=import-error, no-name-in-module
+
+
+def delete_remote_connection(nm_profile, timeout, check_mode):
+    if not nm_profile:
+        logging.info("NULL NM.RemoteConnection, no need to delete")
+        return False
+
+    if not check_mode:
+        main_loop = client.get_mainloop(timeout)
+        user_data = main_loop
+        nm_profile.delete_async(
+            main_loop.cancellable,
+            _nm_profile_delete_call_back,
+            user_data,
+        )
+        logging.debug(
+            "Deleting profile {id}/{uuid} with timeout {timeout}".format(
+                id=nm_profile.get_id(), uuid=nm_profile.get_uuid(), timeout=timeout
+            )
+        )
+        main_loop.run()
+    return True
+
+
+def _nm_profile_delete_call_back(nm_profile, result, user_data):
+    main_loop = user_data
+    if main_loop.is_cancelled:
+        return
+
+    try:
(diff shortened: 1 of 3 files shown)


cathay4t:
@ffmancera Please take review when you have time.

cathay4t:
@richm Any idea why the Fedora CI hang forever?

richm:
> @richm Any idea why the Fedora CI hang forever?

fedora-31 has been removed from CI - please updated the required tests for network CI - we now test only fedora-32 and fedora-33

fedora-32 is only tested with ansible 2.9 and later - please remove fedora-32/ansible-2.8 as an expected required test from the network CI

tyll:
> > @richm Any idea why the Fedora CI hang forever?
> 
> fedora-31 has been removed from CI - please updated the required tests for network CI - we now test only fedora-32 and fedora-33
> 
> fedora-32 is only tested with ansible 2.9 and later - please remove fedora-32/ansible-2.8 as an expected required test from the network CI

Thanks, I adjusted the branch protection rules.

cathay4t:
The CI failued on Fedora 33 is fixed in PR https://github.com/linux-system-roles/network/pull/327

module_utils/network_lsr/nm/connection.py
```diff
@@ -0,0 +1,72 @@
...
+    if not nm_profile:
+        logging.info("NULL NM.RemoteConnection, no need to delete")
+        return False
+
+    # Whether profile has to be deactivated first
+    nm_client = get_client()
+    for nm_ac in nm_client.get_active_connections():
+        cur_nm_profile = nm_ac.get_connection()
+        if cur_nm_profile and cur_nm_profile.get_uuid() == nm_profile.get_uuid():
+            deactivate_active_connection(nm_ac, timeout, check_mode)
```
thom311, line 29: this branch does something quite different from before. Before, the code made an effort to make the profile only volatile (if the API supports that). The effect is that the profile only gets deleted after the profile is no longer active.

This branch, explicitly deactivates the profile (if active) and removes it right away.

Sure, the way how we delete a profile could also be differently. But it was done that way for reasons. The commit message makes no mention why this changes and why the new way is preferable.

I also wonder why the profile gets first deactivated, only to delete it afterwards. That does not seem necessary. There is a code comment `"Whether profile has to be deactivated first"`, but it would be good if the code comment explains *why* this is done.

tests/playbooks/tests_states.yml
```diff
@@ -78,8 +78,16 @@
               - tasks/activate_profile.yml
             lsr_test:
               - tasks/remove_profile.yml
+            lsr_assert_when:
+              # Device should be absent because of NetworkManager
```
tyll, line 82: This contradicts the user story of this test/the expected behavior: "I can remove an existing profile without taking it down" (see Thomas' comment about making the profile volatile).

module_utils/network_lsr/nm/connection.py
cathay4t: OK. I missed this use case. I will restored back to old behavior.

cathay4t:
@thom311 New patch uploaded. Please review when you have time.
The new patch will try volatilize connection profile when update2 is supported, otherwise delete the profile.

PR #303, opened 2020-10-26
inclusive language: use "port" instead of "slave"

ffmancera:
In order to promote a diverse and inclusive environment we are replacing
all the reference to "slave" for "port".


module_utils/network_lsr/argument_validator.py
```diff
@@ -1449,13 +1453,14 @@ def _validate_post(self, value, name, result):
                     "802.1x settings only allowed for ethernet or wireless interfaces.",
                 )
 
-        for k in self.VALID_FIELDS:
-            if k in result:
+        for name in self.VALID_FIELDS:
```
tyll, line 1456: I like that you improved the variables :-)

tyll:
Please add a unit test to ensure that the deprecated option still works and results in the non-deprecated option and also for other special cases for example when someone specifies both the deprecated and new option.

ffmancera:
@cathay4t This PR depends on #304

PR #48, opened 2018-05-16
Coveralls + directory move

pcahyna:
This amends the commits in #46 to move `roles/` under `tests/` and rearranges them for fewer moves.

PR #307, opened 2020-10-28
Sync with template.

nhosoi:
Merge: e9f2c3a 4f3c96a
    Merge pull request #48 from nhosoi/collections
    runcollection.sh - stop passing an argument list

This change allows to eliminate the test argument list as well as the
branch name to retrieve the lsr_role2collection.py script for the
collection test.


ffmancera (changes requested):
Hi! Thank you for this. The network-role requires all the commits to be signed using `git commit -s`. In addition, could you provide a better commit title? Something like `sync with travis/runcollection.sh template` or similar would help.

ffmancera (approved):
LGTM, thanks!

Issue #301, opened 2020-10-25
CI failures on Fedora 32


cathay4t:
Taking notes for CI failure due to known bugs:

 * Service is in unknown state: https://github.com/ansible/ansible/issues/71528
 * Check which services are running: https://github.com/ansible/ansible/issues/69302

pcahyna:
@cathay4t Regarding the first. Was this present in original Fedora 32 and fixed in an update? Or vice-versa, was it working in original Fedora 32 and broken in an update? Or is it broken both in original F432 release and the current version? I am asking because I would like to know whether the ongoing change to use most recent updated composes instead of the original release fixes the issue, introduces the issue, or makes no difference.

Regarding the second. That was fixed a months ago. Did it reappear?

richm:
Were these problems only with ansible 2.8?  We recently dropped support in CI for fedora 32 with ansible 2.8

cathay4t:
Closing as CI has Fedora 32 with ansible 2.8 any more.

PR #309, opened 2020-10-29
Add 'defroute' option

jaedolph:
Addresses #236


jaedolph:
Waiting on fixes for #301 and #310 before adding integration tests (as static ipv6 addresses can not be correctly added by initscripts to test if default ipv6 routes are present/absent)

jaedolph:
Had some issues with RHEL8/Fedora not bringing the veth interface up during my new integration test. Seems to work if I put the veth peer in a network namespace, give it an ip and activate it though. Hope this isn't too hacky.

README.md
```diff
@@ -333,6 +333,11 @@ The IP configuration supports the following options:
     given in the `dns` option and a list of domains to search given in the
     `dns_search` option.
 
+* `defroute`
```
cathay4t, line 336: I don't like `abbreviation` in public API/Schema. Can we use `auto-gateway`?

We are not just use whatever interface/name used by initscripts or network manager. In stead, we should define clear and simple network describing term.

For this option, I would like to see:
 * `auto-gateway: False` means discard gateway retrieved by DHCPv4 or IPv6-RA.
 * `auto-gateway: False` is ignored when interface been configured with static gateway.

cathay4t (changes requested):

README.md
jaedolph: What you have described is not really what this setting does. It doesn't have much to do with DHCP and just deals with how routes are defined for the interface. If set to true you will get a `default via 192.168.1.1 dev eth0` or similar but if set to false this route will not get created. 

If you are looking for a more descriptive name maybe `add-default-route` would be better?

jaedolph:
[citest bad]

README.md
cathay4t: Where does this `default via 192.168.1.1 dev eth0` learn from? Not DHCPv4 or IPv6-RA?

jaedolph: It either comes from DHCP or a static gateway set by the user. The behaviour of setting or not setting a default route can be configured in a network that does not have DHCP running.

cathay4t: why `defroute` is need when no DHCP/IPv6-RA is running?

richm:
[citest]

README.md
jaedolph: Here is an example of when this option can be used where no DHCP is used in both networks (I have seen this situation at a customer site):

Network A (connected to eth0): 10.1.2.0/24 main network used to access the internet. Gateway is at 10.1.2.1
Network B (connected to eth1): 192.168.1.0/24 internal network used for management of network devices. Also has a router at 192.168.1.1 used to access another part of the network.

In this case we may want to specify the static gateway on both interfaces (10.1.2.1 on eth0 and 192.168.1.1 on eth1) although we want the default to only be the internet connected network (10.1.2.0/24) so we would set this as DEFROUTE=YES and the 192.168.1.0/24 network as DEFROUTE=NO. I realise you could just not specify a static gateway for the 192.168.1.0 network but I have seen a lot of admins that prefer to configure it this way using the DEFROUTE setting.

Provisioning systems like Red Hat Satellite and Foreman also use this method by default for configuring primary and secondary network interfaces on hosts.

richm:
[citest]

jaedolph:
[citest bad]

README.md
cathay4t: Is my below understanding correct?
 * User would like to have __two__ interface both holding default gateways in __config__.
 * User only want one of the interface holding default gateways in __kernel__.

I failed to find any benefit could user get for this setup.

richm:
please rebase to latest `main` branch

ffmancera:
The fixes have been merged, let's rebase this with `main` branch. @jaedolph Could you do it please? If not, I can. Thank you!

ffmancera (changes requested):
I do agree with the changes done, I just have 2 comments:

* I do not like the name `defroute` is not descriptive or user friendly. I think all terms in an API must be descriptive to make it easier for users.
* The commit title is too long and the description is missing details of what is being done. Please, could you update this? Thanks!

jaedolph:
Hi @ffmancera, i will have a look at updating this tonight

jaedolph:
[citest bad]

jaedolph:
[citest bad]

jaedolph:
> I do agree with the changes done, I just have 2 comments:
> 
> 
> 
> * I do not like the name `defroute` is not descriptive or user friendly. I think all terms in an API must be descriptive to make it easier for users.
> 
> * The commit title is too long and the description is missing details of what is being done. Please, could you update this? Thanks!

I have changed the variable name to `add_default_route`. Is this descriptive enough?

Updated the PR comment to better explain the new feature.

The branch has been rebased to the latest main branch and all the tests are passing except "rhel8-y". I don't know why this one is failing...

README.md
tyll: @jaedolph I am also getting confused. `gateway4`/`gateway6` define the gateway for the default route, so setting them to a value and then disabling the default route in a different setting does not seem to make sense. `auto_gateway` seems to be a good choice here. Do we need this to have for bothn IPv4/IPv6?

library/network_connections.py
```diff
@@ -497,6 +497,15 @@ def ifcfg_create(
             if ip["gateway6"] is not None:
                 ifcfg["IPV6_DEFAULTGW"] = ip["gateway6"]
 
+            if ip["add_default_route"] is not None:
+                # note that IPV6_DEFROUTE has no effect if initscripts are used
```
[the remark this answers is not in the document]
jaedolph: Have removed this now.

README.md
```diff
@@ -335,6 +335,18 @@ The IP configuration supports the following options:
...
+- `add_default_route`
+
+    If enabled, a default route will be configured using the default gateway. If disabled,
+    the default route will be removed.
+
+    If this variable is not specified, the role will use the default behavior of the
+    `network_provider` selected.
+
+    Setting this option to `no` is equivalent to:
+    - `DEFROUTE = yes` in initscripts, or
```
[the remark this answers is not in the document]
jaedolph: Oops, that is a typo. Have fixed now.

README.md
jaedolph: @cathay4t @tyll 
Yes I agree that in this situation it would be better to configure the non-default interface without a gateway. However, the default behavior for systems like Red Hat Satellite and Foreman is to provision static interfaces with a gateway and specify the non-default interface as `DEFROUTE=no`. I believe users who are using this setup would benefit from the ability to use this role to maintain/enforce the same config as their provisioning system has created.

I feel like `auto_gateway` or `auto_default_route` are a bit misleading as it will not determine the gateway automatically if the interface is statically configured. Although if I had to choose one of these options my preference would be `auto_default_route`. 

FYI here is the description of this feature from the NetworkManager documentation:
"**never-default** If TRUE, this connection will never be the default connection for  this IP type, meaning it will never be assigned the default route by  NetworkManager."

README.md
```diff
@@ -335,6 +335,18 @@ The IP configuration supports the following options:
 
     Manual addressing can be specified via a list of addresses under the `address` option.
 
+- `add_default_route`
```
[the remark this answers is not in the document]
jaedolph: Please see my comment here https://github.com/linux-system-roles/network/pull/309#discussion_r612886527

README.md
cathay4t: @jaedolph I dislike two conflicting options in a API:
 * static gateways defined.
 * DEFROUTE=no will ignore above static gateway.

User should get a failure in network-role when they do this even that was working well in their old system using network-scripts(ifup) or NetworkManager.
We don't repeat ourselves on bad design in new project.

For DHCP/IPv6-RA, if user want to ignore the gateway or routes or DNS, we can introduce `auto-gateway`, `auto-routes` and `auto-dns` which default them to true when not defined.

The `network-role` or `linux-system-role/network` is not child project of NetworkManager, we are defining a API simple enough to cover all use cases regardless the backends we are using.

tyll: Is there someone from the Satellite or Foreman teams who can explain why they are doing it this way?

jaedolph: @cathay4t I don't agree with blocking users from creating a valid config using this method just because it is not the most efficient way to do it. I don't believe throwing an error gives any benefit to the user and will just add extra complexity in the argument validation. As for the name, I don't like `auto_gateway` but I can see how it would be good to group with `auto_routes` and `auto_dns` if we are going to support that. Will update the PR with the new name.

@tyll Can't speak for the Satellite or Foreman teams, but I believe they do it that way for simplicity. A static gateway for a network can be configured globally within a `subnet` resource and any interface created using this subnet gets an address,netmask and gateway configured on the host. Satellite then works out which interfaces on the host are marked as `primary` and sets `DEFROUTE=YES` on those and `DEFROUTE=NO` on all the other ones.

cathay4t: @jaedolph Could you elaborate on this use case of Satellite or Foreman?
 * Why they would like to define static gateway but don't want it to be set in kernel?
 * Why they cannot just remove the static gateway from yaml before sending network-role ansible for `DEFROUTE=NO` hosts?

jaedolph: I did not write the code that implements this in Sat/Foreman but there is no reason they couldn't add another 2 if statements to implement it the way you describe.

Do you want me to only make this option work for DHCP and update the README? I don't agree with this but if that is your only concern I can change it.

cathay4t: My concern is we are providing API base on specific NetworkManager/ifup behavior, it will hunt us when we move to another backed in the future where static gateway with no_default=True is not allowed.

For DHCP part, from what I heard from oVirt, they are expecting like 3 options:
 * Ignoring gateway(0.0.0.0 or ::0).
 * Ignoring extra routes(including gateway) learn from DHCPv4 or IPv6-RA.
 * Ignoring DNS setting from DHCP or IPv6-RA.

Currently, nmstate is using these words:
 * `auto_gateway`
 * `auto_routes`
 * `auto_dns`

You may use other names.

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

README.md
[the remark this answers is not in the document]
ffmancera: I agree with @jaedolph here but if you @cathay4t and @tyll do not agree, I prefer `auto_default_route` then. IMO, this is the only issue left.

ffmancera (approved):
LGTM, thank you!

jaedolph:
@ffmancera Thank you for the review. I have fixed the commit message and rebased on the main branch.

@cathay4t Should we create an issue for implementing the `auto_routes` and `auto_dns` features? I think it is a bit of "scope creep" to add them to this PR.

jaedolph:
[citest bad]

cathay4t (approved):

cathay4t:
> @cathay4t Should we create an issue for implementing the `auto_routes` and `auto_dns` features? I think it is a bit of "scope creep" to add them to this PR.

https://github.com/linux-system-roles/network/issues/393 created for tracking `auto_routes` and `auto_dns`.

cathay4t:
@jaedolph Could you rebase and resolve the conflict? Thank you!

jaedolph:
[citest bad]

jaedolph:
[citest pending]

jaedolph:
[citest bad]

richm:
Can you do a rebase and get rid of the merge commit?

jaedolph:
@cathay4t @richm I have re based now without the merge commit. I just rebased using GitHub before but forgot to fix it up.

richm (approved):
lgtm - as far as python/ansible goes

cathay4t:
Merged. Thanks for the help.

PR #312, opened 2020-10-30
meta/main.yml: CI - Add support for all Fedora images

jharuda:


richm (approved):
lgtm - I guess we have to merge this in order to see if the role actually works on Fedora 33 - if not, I guess we can make the test non-gating until we fix whatever issues, if any - worst case, the role simply does not support Fedora 33, and we have to revert the commit in a subsequent PR

tyll:
> lgtm - I guess we have to merge this in order to see if the role actually works on Fedora 33 - if not, I guess we can make the test non-gating until we fix whatever issues, if any - worst case, the role simply does not support Fedora 33, and we have to revert the commit in a subsequent PR

Why does it need to be merged for this? What is checking the main branch but not the PR?

richm:
> Why does it need to be merged for this? What is checking the main branch but not the PR?

never mind - I think I was looking at an earlier implementation that was using the meta/main.yml from the repo and not the one from the PR

richm:
@jharuda @cathay4t for Fedora 33 - we are getting this error: https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-network-pull-linux-system-roles_network-312-42fd5e7-fedora-33-20201031-174531/artifacts/ansible.log
```
TASK [Hash cacert] *************************************************************
task path: /tmp/tmpdh7a58dg/tests/playbooks/tasks/test_802.1x_capath.yml:31
fatal: [/cache/Fedora-Cloud-Base-33-20201031.0.x86_64.qcow2]: FAILED! => {"changed": false, "cmd": "openssl x509 -hash -noout -in /etc/pki/tls/my_ca_certs/cacert.pem", "msg": "[Errno 2] No such file or directory: b'openssl'", "rc": 2}
```
I guess `openssl` is not installed by default anymore??  So that will need to be added to the test as a package dependency.

cathay4t:
@richm Thanks. I have fixed the openssl test dependency in PR https://github.com/linux-system-roles/network/pull/315 Let's wait a little bit.

pcahyna:
Hi, we are favoring an alternate approach of simply declaring Fedora: all. This way we will not have to bump the supported versions every time there is a new Fedora release, and I believe that the intent is to include the newest version among the supported versions anyway, so having to add it manually should be in most cases superfluous. What do you think, @cathay4t @tyll ?

pcahyna:
(Yes, it can mean some test failures if there are things that need to be adjusted for the new version, as shown in the discussion above.)

jharuda:
I changed it to use all Fedora images.

richm (approved):
lgtm

cathay4t (approved):
LGTM.

PR #315, opened 2020-11-12
test 802.1x: Install openssl as test dependency

cathay4t:
The Fedora 33 cloud image does not have openssl installed by default,
install openssl before using its cli tool.


tests/tasks/test_802.1x_capath.yml
```diff
@@ -28,6 +28,10 @@
     src: "cacert.pem"
     dest: "/etc/pki/tls/my_ca_certs/cacert.pem"
     mode: 0644
+- name: Install openssl (test depednecy)
```
tyll, line 31: there is a typo in dependency

PR #316, opened 2020-11-12
test env: Bring test veth peer up

cathay4t:
Bring the test veth peer up and set as managed by NetworkManager when
possible.

### DIFF
--- a/tests/tasks/manage_test_interface.yml
+++ b/tests/tasks/manage_test_interface.yml
@@ -17,9 +17,19 @@
 
 # veth
 - name: Create veth interface {{ interface }}
-  command: ip link add {{ interface }} type veth peer name peer{{ interface }}
+  command: "{{ item }}"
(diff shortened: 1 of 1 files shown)


tests/tasks/manage_test_interface.yml
```diff
@@ -17,9 +17,16 @@
 
 # veth
 - name: Create veth interface {{ interface }}
-  command: ip link add {{ interface }} type veth peer name peer{{ interface }}
```
tyll, line 20: Please amend the commit message to explain why this is needed instead of only explaining what is done.

cathay4t: Commit message updated.

cathay4t:
It has been two weeks, the change is test only and trivial. CI passed except Fedora CI is hanging.

Merging.

PR #234, opened 2020-05-21
Add support for ethtool coalesce settings

ichigo663:
Also add related tests.

Resolves: #112

PR #322, opened 2020-11-27
Add support for ethtool coalesce settings

ichigo663:
Also add related tests.

Resolves: #112


ichigo663:
Supersedes #234

tests/playbooks/tests_ethtool_coalesce.yml
```diff
@@ -0,0 +1,115 @@
...
+              - name: "{{ interface }}"
+                type: ethernet
+                autoconnect: no
+                ip:
+                  dhcp4: no
+                  auto6: no
+        - name: Get profile's coalescing options
+          command: >
+              grep -q ^ETHTOOL_OPTS
+              /etc/sysconfig/network-scripts/ifcfg-{{ interface }}
```
cathay4t, line 44: Fedora 33 is using key-file by default now. You test should not use hard coded path.
In stead of checking what is written in profile, can you find a way to check whether certain coalesce option is applyed to kernel.

cathay4t (changes requested):

tests/playbooks/tests_ethtool_coalesce.yml
ichigo663: I'm afraid not, coalesce settings need kernel support and very few drivers have it. One option could be to use netdevsim but coalesce settings will be supported from Linux 5.11 onwards.

That's why I'm checking the profile instead.

Maybe telling NM to use ifcfg for this test is an option, what do you think?

cathay4t: How about use `nmcli -f ethtool  c  show  eth1`?

ichigo663: So simple that I did not think about it, I changed the PR to use `nmcli -g ethtool`.
PR updated, thanks!

cathay4t (approved):
LGTM. Thanks.

PR #324, opened 2020-11-30
RFE: Support dummy interfaces

liangwen12year:


cathay4t (changes requested):
Integration test case required.

liangwen12year:
> Integration test case required.
Got it!

tests/playbooks/tests_dummy.yml
```diff
@@ -0,0 +1,30 @@
...
+              - tasks/assert_device_absent.yml
+            lsr_test:
+              - tasks/create_dummy_profile.yml
+            lsr_assert:
+              - tasks/assert_profile_present.yml
+              - tasks/assert_device_present.yml
+            lsr_cleanup:
+              - tasks/cleanup_profile+device.yml
+      tags:
+        - "tests::cleanup"
```
tyll, line 30: cleanup seems to be wrong here

liangwen12year: got it. I found the root cause of the failure. 

`nmcli con add type dummy ifname dummy0 ipv4.addresses 192.0.2.42/30` will never create a profile in `/etc/sysconfig/network-scripts/`

Maybe we can change the tests file to exclude `- tasks/assert_profile_present.yml`

cathay4t (approved):

PR #327, opened 2020-12-04
test: Fix profile assertion on Fedora 33

cathay4t:
The NetworkManager in Fedora 33 does not use ifcfg-rh plugin by default,
the CI will fail on Fedora 33 with:

```
TASK [assert that profile 'bond0' is present] **********************************
task path: /tmp/tmpaz9m374e/tests/playbooks/tasks/assert_profile_present.yml:4
fatal: [/cache/fedora-33.qcow2]: FAILED! => {
    "assertion": "profile_stat.stat.exists",
    "changed": false,
    "evaluated_to": false,
    "msg": "profile /etc/sysconfig/network-scripts/ifcfg-bond0 does not exist"
}
```

Previously, we are checking the existence of
`/etc/sysconfig/network-scripts/` to determine whether ifcfg-rh plugin
is enabled. This is incorrect on Fedora 33.

The fix is checking the FILENAME[1] used for storing the NetworkManager
connection, the profile is considered as exists when it exists and does
not contains `/run`.

Since we cannot tell which provider we are using, we just check both
initscripts files and NetworkManager connections.

[1]: nmcli -f NAME,FILENAME connection show

### DIFF
--- a/tests/integration/test_ethernet.py
+++ b/tests/integration/test_ethernet.py
@@ -122,4 +122,7 @@ def test_static_ip_with_ethernet(testnic1, provider, network_lsr_nm_mock):
     ]
     _configure_network(connections, provider)
     assert ip_address in _get_ip_addresses(testnic1)
-    assert os.path.exists("/etc/sysconfig/network-scripts/ifcfg-" + testnic1)
+    if provider == "initscripts":
+        assert os.path.exists("/etc/sysconfig/network-scripts/ifcfg-" + testnic1)
+    else:
+        subprocess.check_call(["nmcli", "connection", "show", testnic1])
--- a/tests/tasks/assert_profile_absent.yml
+++ b/tests/tasks/assert_profile_absent.yml
@@ -3,5 +3,5 @@
 - include: get_profile_stat.yml
 - name: "assert that profile '{{ profile }}' is absent"
   assert:
-    that: not profile_stat.stat.exists
-    msg: "profile {{ profile_path }} does exist"
+    that: not lsr_net_profile_exists
+    msg: "profile {{ profile }} does exist"
--- a/tests/tasks/assert_profile_present.yml
+++ b/tests/tasks/assert_profile_present.yml
@@ -3,5 +3,5 @@
 - include: get_profile_stat.yml
 - name: "assert that profile '{{ profile }}' is present"
   assert:
-    that: profile_stat.stat.exists
-    msg: "profile {{ profile_path }} does not exist"
+    that: lsr_net_profile_exists
+    msg: "profile {{ profile }} does not exist"
--- a/tests/tasks/get_profile_stat.yml
+++ b/tests/tasks/get_profile_stat.yml
@@ -1,26 +1,24 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
-- name: "Get stat for network-scripts"
-  stat:
-    get_attributes: false
-    get_checksum: false
-    get_mime: false
-    path: "/etc/sysconfig/network-scripts"
-  register: network_scripts_stat
-- name: Set profile path (network-scripts)
-  set_fact:
-    profile_path: /etc/sysconfig/network-scripts/ifcfg-{{ profile }}
-  when:
-    - network_scripts_stat.stat.exists
-- name: Set profile path (NetworkManager system-connections)
-  set_fact:
-    profile_path: /etc/NetworkManager/system-connections/{{ profile }}
-  when:
-    - not network_scripts_stat.stat.exists
+- set_fact: lsr_net_profile_exists=false
+
 - name: stat profile file
   stat:
     get_attributes: false
     get_checksum: false
     get_mime: false
-    path: "{{ profile_path }}"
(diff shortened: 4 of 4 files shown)


liangwen12year (commented):
Just out of curiousity, do we also need to consider separately for CentOS ?
And I tested on Fedora 33 qcow2 image, https://kojipkgs.fedoraproject.org/compose/cloud/Fedora-Cloud-33-20201205.0/compose/Cloud/x86_64/images/, the tests failed on `tests_states_nm.yml` and `tests_bond_nm.yml`, the network interface profile was created in `/etc/sysconfig/network-scripts/`.

cathay4t:
> Just out of curiousity, do we also need to consider separately for CentOS ?
> And I tested on Fedora 33 qcow2 image, https://kojipkgs.fedoraproject.org/compose/cloud/Fedora-Cloud-33-20201205.0/compose/Cloud/x86_64/images/, the tests failed on `tests_states_nm.yml` and `tests_bond_nm.yml`, the network interface profile was created in `/etc/sysconfig/network-scripts/`.

CentOS/RHEL 8 will enable ifcfg-rh plugin of NetworkManager which store connections as `/etc/sysconfig/network-scripts/ifcfg-{{ profile_name }}`.

For Fedora 33+ and RHEL 9+, NetworkManager will not enable ifcfg-rh plugin by default, so it will store connections in `/etc/NetworkManager/system-connections/{{ profile_name }}.nmconnection`

thom311:
when adding a profile via D-Bus API, in general the user has no direct control about which file (or even which settings backend will be used).

For example, if NM is configured to prefer "keyfile" plugin, then it may place the file to `/etc/NetworkManager/system-connections`. If the profile is IN_MEMORY_ONLY, it will instead place it to `/run/NetworkManager/system-connections`.

Also, the chosen filename is not known a-priori. Special characters are replaced and (depending on version), NetworkManager may add a ".nmconnection" suffix. See [here](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/3490a09a7d966d11b971d08a2bb802f76f6d5e58/libnm-core/nm-keyfile/nm-keyfile.c#L4294) and [here](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/3490a09a7d966d11b971d08a2bb802f76f6d5e58/src/settings/plugins/keyfile/nms-keyfile-writer.c#L288).

When a profile is added, you can ask the filename via D-Bus (`nmcli -f NAME,FILENAME connection`). But you cannot know it in advance.

---

of course, for testing, none of this matters. Just assume to know what the expected filename is. Possibly use wildcards like `/etc/NetworkManager/system-connections/{{ profile_name }}.*` and assert that it matches exactly one file. Or ask NetworkManager which file it uses, and assert that it too uses the expected name.

tyll:
Also, some profile types (such as dummies) are not supported by ifcfg-rh and some NetworkManager versions don't use a `.nmconnection` suffix.

cathay4t:
Thanks for the info. I will change the test code to use `nmcli c` instead of checking profile file path.

liangwen12year:
> > Just out of curiousity, do we also need to consider separately for CentOS ?
> > And I tested on Fedora 33 qcow2 image, https://kojipkgs.fedoraproject.org/compose/cloud/Fedora-Cloud-33-20201205.0/compose/Cloud/x86_64/images/, the tests failed on `tests_states_nm.yml` and `tests_bond_nm.yml`, the network interface profile was created in `/etc/sysconfig/network-scripts/`.
> 
> CentOS/RHEL 8 will enable ifcfg-rh plugin of NetworkManager which store connections as `/etc/sysconfig/network-scripts/ifcfg-{{ profile_name }}`.
> 
> For Fedora 33+ and RHEL 9+, NetworkManager will not enable ifcfg-rh plugin by default, so it will store connections in `/etc/NetworkManager/system-connections/{{ profile_name }}.nmconnection`

The new commit failed again on `tests_states_nm.yml` and `tests_bond_nm.yml` when testing with fedora 33 qcow2 image. The truth is that those interface profiles were still created in  `/etc/sysconfig/network-scripts/` with the name `ifcfg-{{ profile }}`

cathay4t:
@liangwen12year Can you take a look on this new patch?
It have all  the CI passed(including Fedora 33). The failure on RHEL/CentOS 6 is caused by yum repo.

tyll:
> It have all the CI passed(including Fedora 33). The failure on RHEL/CentOS 6 is caused by yum repo.

Regarding RHEL/CentOS6, I guess this is because EPEL for CentOS/RHEL 6 is EOL.

liangwen12year:
> @liangwen12year Can you take a look on this new patch?
> It have all the CI passed(including Fedora 33). The failure on RHEL/CentOS 6 is caused by yum repo.

Got it !

liangwen12year:
> @liangwen12year Can you take a look on this new patch?
> It have all the CI passed(including Fedora 33). The failure on RHEL/CentOS 6 is caused by yum repo.

Tested, it worked correctly, and it is also compatible with (Support Dummy Interface)[https://github.com/linux-system-roles/network/pull/324]. LGTM. Thanks a lot!

PR #329, opened 2020-12-10
use github actions instead of travis

richm:
adds a github actions tox workflow for CI testing, since
travis is discontinuing support for open source projects


richm:
note that the new jobs/statuses won't be used until this PR is merged

.github/workflows/tox.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: checkout PR
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.pyver }}
+      - name: Install platform dependencies
+        run: |
+          set -euxo pipefail
+          .travis/preinstall
```
tyll, line 21: The preinstall script should probably be moved into .github/workflows

.github/workflows/tox.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Install pip, tox, python dependencies
+        run: |
+          set -euxo pipefail
+          python -m pip install --upgrade pip
+          pip install tox
```
tyll, line 26: Could this be moved in the preinstall script?

.github/workflows/tox.yml
richm: 1) The purpose of this PR is to move as quickly as possible from travis to github actions to cover the gap in CI coverage left when travis pulled the plug on open source projects.  So I don't want to make any other changes in this PR.
2) This will be provided by tox-lsr, which will hopefully be following once github actions are working

.github/workflows/tox.yml
richm: See above.

cathay4t (approved):
Just test only changes, let's merge it and see how it goes.

PR #330, opened 2020-12-11
Collections - workaround for the module_utils path finding issue in ansible 2.9

nhosoi:
Please ignore. Thanks.

### DIFF
--- a/module_utils/network_lsr/nm/active_connection.py
+++ b/module_utils/network_lsr/nm/active_connection.py
@@ -6,11 +6,8 @@ import logging
 
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.client import GLib  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import NM  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import get_mainloop  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import get_client  # noqa:E501
-from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+from ansible.module_utils.network_lsr.nm import client  # noqa:E501
+from ansible.module_utils.network_lsr.nm import error  # noqa:E501
 
 # pylint: enable=import-error, no-name-in-module
 
@@ -19,11 +16,11 @@ NM_AC_STATE_CHANGED_SIGNAL = "state-changed"
 
 
 def deactivate_active_connection(nm_ac, timeout, check_mode):
-    if not nm_ac or nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
+    if not nm_ac or nm_ac.props.state == client.NM.ActiveConnectionState.DEACTIVATED:
         logging.info("Connection is not active, no need to deactivate")
         return False
     if not check_mode:
-        main_loop = get_mainloop(timeout)
+        main_loop = client.get_mainloop(timeout)
         logging.debug(
             "Deactivating {id} with timeout {timeout}".format(
                 id=nm_ac.get_id(), timeout=timeout
@@ -34,12 +31,12 @@ def deactivate_active_connection(nm_ac, timeout, check_mode):
             NM_AC_STATE_CHANGED_SIGNAL, _nm_ac_state_change_callback, user_data
         )
         logging.debug(
-            "Registered {signal} on NM.ActiveConnection {id}".format(
+            "Registered {signal} on client.NM.ActiveConnection {id}".format(
                 signal=NM_AC_STATE_CHANGED_SIGNAL, id=nm_ac.get_id()
             )
         )
-        if nm_ac.props.state != NM.ActiveConnectionState.DEACTIVATING:
-            nm_client = get_client()
+        if nm_ac.props.state != client.NM.ActiveConnectionState.DEACTIVATING:
+            nm_client = client.get_client()
             user_data = (main_loop, nm_ac, nm_ac.get_id(), handler_id)
             nm_client.deactivate_connection_async(
                 nm_ac,
@@ -47,7 +44,9 @@ def deactivate_active_connection(nm_ac, timeout, check_mode):
                 _nm_ac_deactivate_call_back,
                 user_data,
             )
-            logging.debug("Deactivating NM.ActiveConnection {0}".format(nm_ac.get_id()))
+            logging.debug(
+                "Deactivating client.NM.ActiveConnection {0}".format(nm_ac.get_id())
+            )
         main_loop.run()
     return True
 
@@ -57,18 +56,20 @@ def _nm_ac_state_change_callback(nm_ac, state, reason, user_data):
     if main_loop.is_cancelled:
         return
     logging.debug(
-        "Got NM.ActiveConnection state change: {id}: {state} {reason}".format(
+        "Got client.NM.ActiveConnection state change: {id}: {state} {reason}".format(
             id=nm_ac.get_id(), state=state, reason=reason
         )
     )
-    if nm_ac.props.state == NM.ActiveConnectionState.DEACTIVATED:
-        logging.debug("NM.ActiveConnection {0} is deactivated".format(nm_ac.get_id()))
+    if nm_ac.props.state == client.NM.ActiveConnectionState.DEACTIVATED:
+        logging.debug(
+            "client.NM.ActiveConnection {0} is deactivated".format(nm_ac.get_id())
+        )
         main_loop.quit()
 
 
 def _nm_ac_deactivate_call_back(nm_client, result, user_data):
     main_loop, nm_ac, nm_ac_id, handler_id = user_data
-    logging.debug("NM.ActiveConnection deactivating callback")
+    logging.debug("client.NM.ActiveConnection deactivating callback")
     if main_loop.is_cancelled:
         if nm_ac:
             nm_ac.handler_disconnect(handler_id)
@@ -76,8 +77,10 @@ def _nm_ac_deactivate_call_back(nm_client, result, user_data):
 
     try:
         success = nm_client.deactivate_connection_finish(result)
-    except GLib.Error as e:
-        if e.matches(NM.ManagerError.quark(), NM.ManagerError.CONNECTIONNOTACTIVE):
+    except client.GLib.Error as e:
+        if e.matches(
+            client.NM.ManagerError.quark(), client.NM.ManagerError.CONNECTIONNOTACTIVE
+        ):
             logging.info(
                 "Connection is not active on {0}, no need to deactivate".format(
                     nm_ac_id
@@ -122,4 +125,4 @@ def _deactivate_fail(main_loop, handler_id, nm_ac, msg):
     if nm_ac:
         nm_ac.handler_disconnect(handler_id)
     logging.error(msg)
-    main_loop.fail(LsrNetworkNmError(msg))
+    main_loop.fail(error.LsrNetworkNmError(msg))
--- a/module_utils/network_lsr/nm/client.py
+++ b/module_utils/network_lsr/nm/client.py
@@ -4,24 +4,34 @@ import logging
 
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+from ansible.module_utils.network_lsr.nm import error  # noqa:E501
 
 import gi
 
-gi.require_version("NM", "1.0")
+try:
+    gi.require_version("NM", "1.0")
 
-# It is required to state the NM version before importing it
-# But this break the flake8 rule: https://www.flake8rules.com/rules/E402.html
-# Use NOQA: E402 to suppress it.
-from gi.repository import NM  # NOQA: E402
-from gi.repository import GLib  # NOQA: E402
-from gi.repository import Gio  # NOQA: E402
+    # It is required to state the NM version before importing it
+    # But this break the flake8 rule: https://www.flake8rules.com/rules/E402.html
+    # Use NOQA: E402 to suppress it.
+    from gi.repository import NM  # NOQA: E402
+    from gi.repository import GLib  # NOQA: E402
+    from gi.repository import Gio  # NOQA: E402
 
-# pylint: enable=import-error, no-name-in-module
+    # pylint: enable=import-error, no-name-in-module
 
-NM
-GLib
-Gio
+    NM
+    GLib
+    Gio
+except ValueError:
+    # This is to workaround a bug in ansible 2.9 which causes
+    # this code to be executed on the control node, where NM
+    # is not guaranteed to exist. On the other hand, it is
+    # ensured on the managed nodes as NM package is installed
+    # in the network role. Therefore, this exception handling
+    # does not affect the network installation and configuration
+    # on the managed nodes.
+    pass
 
 
 def get_client():
@@ -48,7 +58,7 @@ class _NmMainLoop(object):
 
     def _timeout_call_back(self, _user_data):
         logging.error("Timeout")
-        self.fail(LsrNetworkNmError("Timeout"))
+        self.fail(error.LsrNetworkNmError("Timeout"))
 
     @property
     def cancellable(self):
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1958,13 +1958,11 @@ class Cmd_nm(Cmd):
         self.validate_one_type = ArgValidator_ListConnections.VALIDATE_ONE_MODE_NM
         self._checkpoint = None
         # pylint: disable=import-error, no-name-in-module
-        from ansible.module_utils.network_lsr.nm import (  # noqa: E501
-            NetworkManagerProvider,
-        )
+        from ansible.module_utils.network_lsr.nm import provider  # noqa:E501
 
         # pylint: enable=import-error, no-name-in-module
 
-        self._nm_provider = NetworkManagerProvider()
+        self._nm_provider = provider.NetworkManagerProvider()
 
     @property
     def nmutil(self):
--- a/module_utils/network_lsr/nm/__init__.py
+++ b/module_utils/network_lsr/nm/__init__.py
@@ -1,9 +1,7 @@
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.provider import (  # noqa:E501
-    NetworkManagerProvider,
-)
+from ansible.module_utils.network_lsr.nm import provider  # noqa:E501
 
 # pylint: enable=import-error, no-name-in-module
 
-NetworkManagerProvider
+provider.NetworkManagerProvider
--- a/module_utils/network_lsr/nm/provider.py
+++ b/module_utils/network_lsr/nm/provider.py
@@ -4,10 +4,8 @@ import logging
 
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.active_connection import (  # noqa:E501
-    deactivate_active_connection,
-)
-from ansible.module_utils.network_lsr.nm.client import get_client  # noqa:E501
+from ansible.module_utils.network_lsr.nm import active_connection  # noqa:E501
+from ansible.module_utils.network_lsr.nm import client  # noqa:E501
 
 # pylint: enable=import-error, no-name-in-module
 
@@ -17,12 +15,14 @@ class NetworkManagerProvider:
         """
         Return True if changed.
         """
-        nm_client = get_client()
+        nm_client = client.get_client()
         changed = False
         for nm_ac in nm_client.get_active_connections():
             nm_profile = nm_ac.get_connection()
             if nm_profile and nm_profile.get_id() == connection_name:
-                changed |= deactivate_active_connection(nm_ac, timeout, check_mode)
+                changed |= active_connection.deactivate_active_connection(
+                    nm_ac, timeout, check_mode
+                )
         if not changed:
             logging.info("No active connection for {0}".format(connection_name))
 


nhosoi:
@pcahyna, @richm,
The change made in this pr issues the error `ValueError: NetworkManager package is missing?`, instead of cryptic `__spec__ is None` if I run the CI tests without the dependent packages. Do you think this effort is in the right track? (Probably, we should list all the necessary packages for the role in the message, though. Otherwise, the user would hit the missing package error one by one when it's really accessed... :)
```
        "exception": "Traceback (most recent call last):
  File \"/tmp/tmpfisbgac3/ansible_collections/fedora/system_roles/plugins/module_utils/network_lsr/nm/__init__.py\", line 4, in <module>
    from ansible_collections.fedora.system_roles.plugins.module_utils.network_lsr.nm.provider import (  # noqa:E501
  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load
  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked
  File \"<frozen importlib._bootstrap>\", line 655, in _load_unlocked
  File \"<frozen importlib._bootstrap>\", line 618, in _load_backward_compatible
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 107, in load_module
    mod = self._find_module(fullname, None, load=True)[1]
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 247, in _find_module
    exec(code_object, newmod.__dict__)
  File \"/tmp/tmpfisbgac3/ansible_collections/fedora/system_roles/plugins/module_utils/network_lsr/nm/provider.py\", line 7, in <module>
    from ansible_collections.fedora.system_roles.plugins.module_utils.network_lsr.nm.active_connection import (  # noqa:E501
  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load
  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked
  File \"<frozen importlib._bootstrap>\", line 655, in _load_unlocked
  File \"<frozen importlib._bootstrap>\", line 618, in _load_backward_compatible
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 107, in load_module
    mod = self._find_module(fullname, None, load=True)[1]
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 247, in _find_module
    exec(code_object, newmod.__dict__)
  File \"/tmp/tmpfisbgac3/ansible_collections/fedora/system_roles/plugins/module_utils/network_lsr/nm/active_connection.py\", line 9, in <module>
    from ansible_collections.fedora.system_roles.plugins.module_utils.network_lsr.nm.client import GLib  # noqa:E501
  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load
  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked
  File \"<frozen importlib._bootstrap>\", line 655, in _load_unlocked
  File \"<frozen importlib._bootstrap>\", line 618, in _load_backward_compatible
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 107, in load_module
    mod = self._find_module(fullname, None, load=True)[1]
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 247, in _find_module
    exec(code_object, newmod.__dict__)
  File \"/tmp/tmpfisbgac3/ansible_collections/fedora/system_roles/plugins/module_utils/network_lsr/nm/client.py\", line 11, in <module>
    gi.require_version(\"NM\", \"1.0\")
  File \"/usr/lib64/python3.8/site-packages/gi/__init__.py\", line 129, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace NM not available

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File \"/usr/lib/python3.8/site-packages/ansible/executor/task_executor.py\", line 147, in run
    res = self._execute()
  File \"/usr/lib/python3.8/site-packages/ansible/executor/task_executor.py\", line 665, in _execute
    result = self._handler.run(task_vars=variables)
  File \"/usr/lib/python3.8/site-packages/ansible/plugins/action/normal.py\", line 46, in run
    result = merge_hash(result, self._execute_module(task_vars=task_vars, wrap_async=wrap_async))
  File \"/usr/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 825, in _execute_module
    (module_style, shebang, module_data, module_path) = self._configure_module(module_name=module_name, module_args=module_args, task_vars=task_vars)
  File \"/usr/lib/python3.8/site-packages/ansible/plugins/action/__init__.py\", line 206, in _configure_module
    (module_data, module_style, module_shebang) = modify_module(module_name, module_path, module_args, self._templar,
  File \"/usr/lib/python3.8/site-packages/ansible/executor/module_common.py\", line 1250, in modify_module
    (b_module_data, module_style, shebang) = _find_module_utils(module_name, b_module_data, module_path, module_args, task_vars, templar, module_compression,
  File \"/usr/lib/python3.8/site-packages/ansible/executor/module_common.py\", line 1089, in _find_module_utils
    recursive_finder(module_name, remote_module_fqn, b_module_data, py_module_names,
  File \"/usr/lib/python3.8/site-packages/ansible/executor/module_common.py\", line 725, in recursive_finder
    module_info = CollectionModuleInfo(py_module_name[-idx],
  File \"/usr/lib/python3.8/site-packages/ansible/executor/module_common.py\", line 645, in __init__
    self.get_source()
  File \"/usr/lib/python3.8/site-packages/ansible/executor/module_common.py\", line 660, in get_source
    pkg = import_module(self._package_name)
  File \"/usr/lib64/python3.8/importlib/__init__.py\", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File \"<frozen importlib._bootstrap>\", line 1014, in _gcd_import
  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load
  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked
  File \"<frozen importlib._bootstrap>\", line 655, in _load_unlocked
  File \"<frozen importlib._bootstrap>\", line 618, in _load_backward_compatible
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 107, in load_module
    mod = self._find_module(fullname, None, load=True)[1]
  File \"/usr/lib/python3.8/site-packages/ansible/utils/collection_loader.py\", line 247, in _find_module
    exec(code_object, newmod.__dict__)
  File \"/tmp/tmpfisbgac3/ansible_collections/fedora/system_roles/plugins/module_utils/network_lsr/nm/__init__.py\", line 10, in <module>
    raise ValueError(\"NetworkManager package is missing?\")
ValueError: NetworkManager package is missing?
",


```

pcahyna:
@nhosoi is it possible to detect that the code is running on the control node, and in this case silently ignore the error, instead of gracefully failing?

nhosoi:
> @nhosoi is it possible to detect that the code is running on the control node, and in this case silently ignore the error, instead of gracefully failing?

Hi @pcahyna, reading this section of "Ansible module architecture", the library/modules and module_utils codes are zipped, sent to the managed node, unzipped and executed. So far, I cannot find how to detect the current node is the control node or the managed node...
https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#ansiballz-framework

BTW, if it's ok to ignore the particular error, can we just ignore it by doing something like this? (sorry, I haven't tested the code yet, though...)
```
+try:
+    from ansible.module_utils.network_lsr.nm.provider import (  # noqa:E501
+        NetworkManagerProvider,
+    )
+except ImportError:
+    raise ImportError("import NetworkManagerProvider failed by ImportError")
+except ValueError as err:
+    if "__spec__" in err:
+        pass
+    else:
+        raise
```

nhosoi:
@pcahyna, I've gotten an input from @richm that we may not need to make further changes on this issue by ourselves. If that's the case, is it ok to close this experimental pr?

nhosoi:
[citest]

nhosoi:
[citest]

nhosoi:
@richm, @sivel, @pcahyna, could you please take a look at the diffs in this pr?
I think I followed the instructions by @sivel, but I might have made any mistakes...
```
Testing the suggestion from @sivel:
replacing
  from ansible.module_utils.network_lsr.nm.provider import NetworkManagerProvider,
with
  from ansible.module_utils.network_lsr.nm import provider
and then use provider.NetworkManagerProvider
```
The CI tests in the collections format still fails unless I pre-install the dependent packages for the network role.
[0] - http://pastebin.test.redhat.com/926539

The issue is the network module_utils uses `gi` to load NM [1] , which is executed on the control node to ensure the module_utils paths. Unfortunately, NetworkManager is not installed on the CI test control node by default, so the attempt fails.  And the similar failure would be observed at the customer site unless the control host/node has the package pre-installed.  
[1] - https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/nm/client.py#L11

Please note that the failed place in the stacktraces in [0] is the line 660, which is one line above of the proposed fix in [2], which makes me worried the network role may not be spared by the fix only in [2].
```
 657     def get_source(self):
 658         # FIXME (nitz): need this in py2 for some reason TBD, but we shouldn't (get_data delegates
 659         # to wrong loader without it)
 660         pkg = import_module(self._package_name)
 661         data = pkgutil.get_data(to_native(self._package_name), to_native(self._mod_name + '.py'))
 662         return data
```
[2] - https://github.com/ansible/ansible/issues/68361#issuecomment-601740874

Another note: I could verify Certificate CI tests passed [3] with the suggested changes without any extra packages pre-installed. So, I'd think this Network case would be quite special.
[3] - https://github.com/linux-system-roles/certificate/pull/71

sivel:
Generally speaking, any modules that are not part of the python stdlib, should be wrapped in `try/except`.  `ansible-test sanity --test import` would fail on any imports that were not properly guarded.

We have a number of examples throughout the codebase, but generally you would need to make sure that missing imports never result in a traceback.

```
try:
    import some_missing_module
    HAS_SOME_MODULE = True
except ImportError:
    HAS_SOME_MODULE = False
```

Then in the modules `main()` you would do something like:

```

def main():
    module = AnsibleModule(...)
    if not HAS_SOME_MODULE:
        module.fail_json('missing some_module')
```

If the import code is in a `module_utils`, you still follow the same pattern, you would just update your `if not` statement.

See:

* https://github.com/ansible-collections/community.docker/blob/main/plugins/module_utils/common.py
* https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/expect.py

On another note, we realize that the current implementation in 2.9 that attempts to import `module_utils` in the module building phase isn't correct.  And 2.10 doesn't do that, but the code from 2.9 to 2.10 was completely rewritten, and we haven't been able to address that portion of code in 2.9 yet.

The changes I outline above should help address the issues, as well as bring the code up to par with the expectations enforced by `ansible-test` which is used by most collections for CI.

nhosoi:
[citest]

nhosoi:
@richm, @pcahyna, @sivel, by capturing the exception from gi.require_version [0], the error [1] I reported in https://github.com/linux-system-roles/network/pull/330#issuecomment-747079304 was worked around. The missing NetworkManager package is actually ignored by ansible 2.9 and the CI tests in the collection formats ran successfully. I was hoping to see an error message `NetworkManager package is not found. Please install it on the control node.`, but it was not logged even in the verbose mode (-vvvv). (So, probably, I'd better get rid of `Please install it on the control node.` part since it's useless, which will not be logged if the failure is in the collection code...)

If the part [0] looks ok, I'm going to squash the commits and lift WIP for wider reviews. Please note that I'm going to ask Network and Certificate team to work on the task for capturing the import exceptions suggested in [2]. 

One more concern is this tox failure https://github.com/linux-system-roles/network/pull/330/checks?check_run_id=1572437340. I've rebased with main, but the error is still observed...

[0] - https://github.com/linux-system-roles/network/pull/330/files#diff-8676715864da9f0f345f37f9a1d253e28e51696897c9fbdd1eede3daf21ba7b5L11-R29
[1] - http://pastebin.test.redhat.com/926539
[2] - https://github.com/linux-system-roles/network/pull/330#issuecomment-747084609

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -6,11 +6,8 @@
 
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.client import GLib  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import NM  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import get_mainloop  # noqa:E501
-from ansible.module_utils.network_lsr.nm.client import get_client  # noqa:E501
-from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+from ansible.module_utils.network_lsr.nm import client  # noqa:E501
+from ansible.module_utils.network_lsr.nm import error  # noqa:E501
```
richm, line 10: You might not need these `  # noqa:E501` if the line length is short enough now

module_utils/network_lsr/nm/client.py
```diff
@@ -4,24 +4,29 @@
 
 # Relative import is not support by ansible 2.8 yet
 # pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+# from ansible.module_utils.network_lsr.nm.error import LsrNetworkNmError  # noqa:E501
+from ansible.module_utils.network_lsr.nm import error  # noqa:E501
```
richm, line 8: can you remove the `# noqa:E501` here?

richm:
> @richm, @pcahyna, @sivel, by capturing the exception from gi.require_version [0], the error [1] I reported in [#330 (comment)](https://github.com/linux-system-roles/network/pull/330#issuecomment-747079304) was worked around. The missing NetworkManager package is actually ignored by ansible 2.9 and the CI tests in the collection formats ran successfully. I was hoping to see an error message `NetworkManager package is not found. Please install it on the control node.`, but it was not logged even in the verbose mode (-vvvv). (So, probably, I'd better get rid of `Please install it on the control node.` part since it's useless, which will not be logged if the failure is in the collection code...)

Right.  I'm not actually sure how you print/log something from within a module (which runs on the managed host) in such a way that it will be reported in the ansible-playbook output (running on the control host) - perhaps @sivel knows (e.g. neither print(), nor logging.error, nor self.module.log, will work - note that self.module.log will log to the journal on the managed host, so for debugging you could use `command: journalctl`, but we do not expect users to do or see that).  It is expected to pass the error message to `module.fail_json()` - but in this case that would involve throwing an exception and bubbling it up through a few layers to the `module.main` function.  That will likely involve a bit more programming.

> 
> If the part [0] looks ok, I'm going to squash the commits and lift WIP for wider reviews. Please note that I'm going to ask Network and Certificate team to work on the task for capturing the import exceptions suggested in [2].

Ok.

> 
> One more concern is this tox failure https://github.com/linux-system-roles/network/pull/330/checks?check_run_id=1572437340. I've rebased with main, but the error is still observed...

`print(f"ERROR: Playbook does not match nominal value: {testfile}")`
This is not python26 syntax - either we change the test so that this script is not run in the py26 environment (@cathay4t can we do this?), or rewrite the script `ensure_provider_tests.py` to run on py26/py27/py3

> 
> [0] - https://github.com/linux-system-roles/network/pull/330/files#diff-8676715864da9f0f345f37f9a1d253e28e51696897c9fbdd1eede3daf21ba7b5L11-R29
> [1] - http://pastebin.test.redhat.com/926539
> [2] - [#330 (comment)](https://github.com/linux-system-roles/network/pull/330#issuecomment-747084609)

sivel:
> Right. I'm not actually sure how you print/log something from within a module (which runs on the managed host) in such a way that it will be reported in the ansible-playbook output (running on the control host) - perhaps @sivel knows

Depending on what you want:

* `module.fail_json()`
* `module.warn()`
* `module.deprecate()`

`warn` and `deprecate` smuggle the data back in the JSON response from the module to the controller, and they they get `[WARNING]` or `[DEPRECATION WARNING]` outputs.

In 2.10+ we have new code in `ansible.module_utils.common.warnings` to allow code without a `module` context to warn.  But you don't have that available of course for 2.9 compat.

The `file` module has a pattern of raising exceptions and registering a `sys.excepthook` instead of directly using `module.fail_json`, but that may be too much effort: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/file.py

module_utils/network_lsr/nm/active_connection.py
nhosoi: I think I added them since they were needed in the collection converted files...

module_utils/network_lsr/nm/client.py
nhosoi: Same here...

richm:
> Depending on what you want:
>    module.fail_json()

I think we should use this method, as it is the documented/recommended method for reporting such errors, and these are legitimate errors, not warnings or some other sort of notice.

nhosoi:
> > Depending on what you want:
> > module.fail_json()
> 
> I think we should use this method, as it is the documented/recommended method for reporting such errors, and these are legitimate errors, not warnings or some other sort of notice.

Not sure this is the correct way to use module.fail_json, I'm testing this commit now...
https://github.com/linux-system-roles/network/pull/330/commits/42a19c93a719dd719b32153a278b1295ee5749d7

sivel:
> Not sure this is the correct way to use module.fail_json, I'm testing this commit now...
> [42a19c9](https://github.com/linux-system-roles/network/commit/42a19c93a719dd719b32153a278b1295ee5749d7)

Typically you would want to import `HAS_NM` into the module that depends on it, and then from within `main()` in that module, call `module.fail_json`.

richm:
I don't think we need to do any of the try/except stuff for system roles.  All of our modules are private to the role - that is, we do not support users using our modules directly - the `library/network_connections.py` module is only supported for use in the network role.  So the role can ensure that the right packages are in place so that the import will succeed.
Currently there is no way to prevent a user from using the module directly, so @sivel a topic for another time is - how can we make private modules and roles in a collection, where a "private module" is a module that can only be used from other roles/modules in its collection, and a "private role" is a role that can only be used by other roles in its collection?

sivel:
> how can we make private modules and roles in a collection

There is no such feature

nhosoi:
I squashed the commits with more detailed comments.
@richm, @pcahyna, @liangwen12year, @cathay4t, could you please review?
Thanks!

module_utils/network_lsr/nm/client.py
```diff
@@ -4,24 +4,28 @@
...
 
-NM
-GLib
-Gio
+    NM
+    GLib
+    Gio
+    HAS_NM = True
+except (ImportError, ValueError) as e:
+    HAS_NM = False
```
richm, line 28: This isn't needed to resolve the import problem is it?  I thought this was only needed to set the `HAS_NM` flag, which we don't need, as discussed in https://github.com/linux-system-roles/network/pull/330#issuecomment-747737327

module_utils/network_lsr/nm/client.py
```diff
@@ -83,4 +87,5 @@ def fail(self, exception):
 
 
 def get_mainloop(timeout):
-    return _NmMainLoop(timeout)
+    if HAS_NM:
+        return _NmMainLoop(timeout)
```
richm, line 91: Same with this - we don't need to check `HAS_NM`

module_utils/network_lsr/nm/client.py
nhosoi: This is needed. Without this catch, gi.require_version("NM", "1.0") raises an exception `ValueError: Namespace NM not available` and the module `network_connections` in the task `Configure networking connection profiles` fails. Please see http://pastebin.test.redhat.com/926539.

richm: > This is needed. Without this catch, gi.require_version("NM", "1.0") raises an exception `ValueError: Namespace NM not available` and the module `network_connections` in the task `Configure networking connection profiles` fails. Please see http://pastebin.test.redhat.com/926539.

Is it because this module also runs on EL6 where there is no NM?

nhosoi: > > This is needed. Without this catch, gi.require_version("NM", "1.0") raises an exception `ValueError: Namespace NM not available` and the module `network_connections` in the task `Configure networking connection profiles` fails. Please see http://pastebin.test.redhat.com/926539.
> 
> Is it because this module also runs on EL6 where there is no NM?

No, this problem occurs when the control node has no NM installed. In our CI test env, the control node fedora 32 has no NM. Thank you for merging this reverting pr [1]. The merge and the deployment undo the NM pre-installation and make this PR testable.

Following the advice from @sivel, "to import the package, not the class" (as described by @rjeffman in https://github.com/linux-system-roles/certificate/pull/71#issuecomment-748051347) fixes the module_utils import failure even with ansible 2.9. So, with this PR 330, we won't see the `__spec__ is None` error any more. But this `gi.require_version("NM", "1.0")` code [0] apparently fails if NM is not installed. Since the code [0] explicitly accesses the package, I think it's reasonable to capture the exception. (Please note that it is safe not to do so on the managed nodes since the package is installed by the network role. This is pure ansible 2.9 collections issue.)

For testing, if we redeploy test-harness having the revert [1] and temporarily remove the try-except catch in this PR, we should be able to see the failure from `gi.require_version("NM", "1.0")` in, e.g., `centos-8/ansible-2.9 (staging)`. Shall we try that?

[0] https://github.com/linux-system-roles/network/pull/330/files#diff-8676715864da9f0f345f37f9a1d253e28e51696897c9fbdd1eede3daf21ba7b5R12

[1] - https://github.com/linux-system-roles/test-harness/pull/136

> I'm trying to understand the conditions (platforms, platform versions) where you get this error.

The condition is the control node, whatever version unless NM is installed...

nhosoi:
[citest]

module_utils/network_lsr/nm/client.py
pcahyna: @nhosoi is there a risk that the function will be actually executed on the control node? If not, then indeed we don't need to check `HAS_NM`, right?

nhosoi: @richm, @pcahyna, ok. Let me remove the `if HAS_NM:` check and see how it goes...

nhosoi: Yep, you are right, @richm, @pcahyna. We don't need the check. Let me squash the test commit into the main one.
```
Testing /tmp/tmpvdv8w27n/tests/network/tests_802_1x_nm.yml...SUCCESS
```

nhosoi: Thanks, @richm, @pcahyna. You are right. By just ignoring the ValueError cause by the missing NetworkManager package, the CI tests in the collections format passes. This is the commit message explaining it.
```
2) In the later module_utils path finding path, gi.require_version("NM", "1.0")
   in module_utils/network_lsr/nm/client.py fails with "ValueError: Namespace
   NM not available" on the control node. By ignoring the exception, the failure
   is worked around. Please note that the missing package issue never occurs
   on the managed nodes since in case of "nm", the NetworkManager package is
   installed in the network role.
```

module_utils/network_lsr/nm/client.py
```diff
@@ -4,24 +4,27 @@
...
+    # pylint: enable=import-error, no-name-in-module
 
-NM
-GLib
-Gio
+    NM
+    GLib
+    Gio
+except ValueError:
+    pass
```
richm, line 27: please add a comment here explaining that the `except ValueError; pass` is to workaround a bug in ansible 2.9 which causes this code to be executed on the control node (if that is the explanation)

richm (approved):
lgtm

richm:
@cathay4t please review

cathay4t (approved):
LGTM. Thanks.

cathay4t:
@nhosoi Thanks for the patch. Please rebase and we can merge it once CI passes.

nhosoi:
> @nhosoi Thanks for the patch. Please rebase and we can merge it once CI passes.

Done! Thanks a lot, @cathay4t.

nhosoi:
[citest bad]

liangwen12year (approved):
LGTM, approved! But looks like the CI was also impacted by my PR, will investiaged it !

nhosoi:
[citest bad]

nhosoi:
> LGTM, approved! But looks like the CI was also impacted by my PR, will investigated it ! @nhosoi , I fixed my PR, and CI will not be impacted anymore, wonder if you can recommit or trigger the CI again ?

Thank you for reviewing this PR, @liangwen12year!! I rebased again. I think this pr passed all the required platform-ansible version combinations. If it is good to go, could you please merge? Thanks, in advance!

liangwen12year:
merged

PR #334, opened 2020-12-17
Fix python line length problem

richm:
The line was too long, and needed to be wrapped in a way that was
compatible with python black formatting.


richm:
I would like to merge this despite the tox ci failures - I am working through all of the problems in CI - this one is simple to address

liangwen12year (approved):
LGTM, approved!

PR #336, opened 2020-12-23
add ansible-lint to github actions

richm:
Since we have disabled linting in molecule, use tox for running
ansible-lint.

PR #340, opened 2021-01-07
Support dns-options in network role

liangwen12year:


liangwen12year:
@cathay4t , do you know why CI failed ? is the failure caused by my PR?

examples/eth_dns_support.yml
```diff
@@ -0,0 +1,46 @@
...
+
+          dns:
+            - 192.0.2.2
+            - 198.51.100.5
+          dns_search:
+            - example.com
+            - subdomain.example.com
+          dns_options:
+            - rotate
+            - timeout:1
```
cathay4t, line 22: How about:

```yaml
dns:
  servers:
    - 1.1.1.1
    - 8.8.8.8
  searches:
    - example.com
    - example.org
   options:
    - free_string_a
```

cathay4t (changes requested):
Please include document changes in `README.md`.

cathay4t:
And test case is also required.

richm:
> @cathay4t , do you know why CI failed ? is the failure caused by my PR?

Looks like the failure was caused by the PR.  Can you run `tox -e py37` locally?

liangwen12year:
> > @cathay4t , do you know why CI failed ? is the failure caused by my PR?
> 
> Looks like the failure was caused by the PR. Can you run `tox -e py37` locally?

Indeed, "tox- e py37" failed locally.

cathay4t (changes requested):
1. The test does not actually verify whether expected DNS entry been written to OS( or at least to nmcli connections).
2. Without using `NM.SettingIPConfig.props.dns_priority`, you cannot guarantee user will get their DNS server in the order they asked.
~~3. All the DNS config should been grouped in to `dns`. And contains `servers`, `searches`, `options`.~~

examples/eth_dns_support.yml
tyll: `dns:` and `dns_search:` already exist, only `dns_options:` was added, so this would be a bigger change. Another thing: Using plural `searches` sounds odd to me.

liangwen12year: @tyll , @cathay4t , Indeed, I already finished implementation for grouping dns into what Gris suggested, but I found a lot of test failures in  './tests/unit/test_network_connections.py'.

cathay4t: As `dns:` and `dns_search` already exist, I am OK to place `dns_options` along with them.

cathay4t:
> > @cathay4t , do you know why CI failed ? is the failure caused by my PR?
> 
> Looks like the failure was caused by the PR. Can you run `tox -e py37` locally?

@richm The failure of py36 is caused by CentOS 6 outdated repo link issue (#335 ). Could you somehow fix the tox-lsr for that?

richm:
> > > @cathay4t , do you know why CI failed ? is the failure caused by my PR?
> > 
> > 
> > Looks like the failure was caused by the PR. Can you run `tox -e py37` locally?
> 
> @richm The failure of py36 is caused by CentOS 6 outdated repo link issue (#335 ). Could you somehow fix the tox-lsr for that?

I did - can you review and approve https://github.com/linux-system-roles/network/pull/336 and https://github.com/linux-system-roles/network/pull/341 ?

examples/eth_dns_support.yml
```diff
@@ -0,0 +1,46 @@
...
+---
+- hosts: all
+  vars:
+    network_connections:
+      - name: eth0
+        type: ethernet
+        ip:
+          route_metric4: 100
+          dhcp4: no
+          #dhcp4_send_hostname: no
```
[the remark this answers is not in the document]
liangwen12year: removed

cathay4t:
@liangwen12year Please rebase so that we could see whether all CI is passing.

liangwen12year:
> @liangwen12year Please rebase so that we could see whether all CI is passing.

rebased

richm:
will need to update tox-lsr for the new version of lsr_role2collection

module_utils/network_lsr/argument_validator.py
```diff
@@ -499,6 +499,11 @@ def __init__(self):
                     nested=ArgValidatorStr("dns_search[?]"),
                     default_value=list,
                 ),
+                ArgValidatorList(
+                    "dns_options",
+                    nested=ArgValidatorStr("dns_options[?]"),
+                    default_value=list,
```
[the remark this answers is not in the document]
liangwen12year: @thom311 @cathay4t , regex args validator added, you can review it when you have time

module_utils/network_lsr/argument_validator.py
```diff
@@ -465,6 +475,22 @@ def _validate_post(self, value, name, result):
 
 
 class ArgValidator_DictIP(ArgValidatorDict):
+    VALID_MODES = [
+        "debug",
```
tyll, line 479: Please sort this list

module_utils/network_lsr/argument_validator.py
```diff
@@ -465,6 +475,22 @@ def _validate_post(self, value, name, result):
...
+        "ndots:\d+",
```
tyll, line 480: Please add unit tests for that show that valid values are accepted and invalid values are not accepted.

module_utils/network_lsr/argument_validator.py
liangwen12year: sorted

module_utils/network_lsr/argument_validator.py
liangwen12year: Hi Till, you mean the python unit test file - `test_network_connections.py`?

tests/unit/test_network_connections.py
```diff
@@ -3029,6 +3059,138 @@ def test_valid_persistent_state(self):
         input_connection.update({"state": "absent"})
         self.assertValidationError(validator, input_connection)
 
+    def test_dns_options_argvalidator(self):
```
liangwen12year, line 3062: @tyll , added the test cases for correct value and wrong value

module_utils/network_lsr/argument_validator.py
```diff
@@ -320,6 +322,14 @@ def _validate_impl(self, value, name):
         result = []
         for (idx, v) in enumerate(value):
             try:
+                if self.enum_values is not None and not any(
+                    re.match(str(x), str(v)) for x in self.enum_values
```
thom311, line 326: the conversion to `str()` doesn't seem right. If the values are not strings to being with, it's not clear how the pattern matching should work (it should not work at all).

module_utils/network_lsr/argument_validator.py
```diff
@@ -305,9 +306,10 @@ def _validate_impl(self, value, name):
 
 
 class ArgValidatorList(ArgValidator):
-    def __init__(self, name, nested, default_value=None):
+    def __init__(self, name, nested, enum_values=None, default_value=None):
         ArgValidator.__init__(self, name, required=False, default_value=default_value)
         self.nested = nested
+        self.enum_values = enum_values
```
thom311, line 312: `enum_values` does not seem a good name for a list of regular expressions.

module_utils/network_lsr/argument_validator.py
```diff
@@ -465,6 +475,22 @@ def _validate_post(self, value, name, result):
...
+        r"ndots:([1-9]\d*|0)$",
+        r"no-check-names$",
+        r"no-reload$",
+        r"no-tld-query$",
+        r"rotate$",
+        r"single-request$",
+        r"single-request-reopen$",
+        r"timeout:([1-9]\d*|0)$",
+        r"trust-ad$",
+        r"use-vc$",
```
thom311, line 491: Maybe add `^` to the beginning of the patterns

module_utils/network_lsr/argument_validator.py
```diff
@@ -465,6 +475,22 @@ def _validate_post(self, value, name, result):
 
 
 class ArgValidator_DictIP(ArgValidatorDict):
+    VALID_MODES = [
```
thom311, line 478: this variable doesn't seem to have a good name.

tests/unit/test_network_connections.py
```diff
@@ -3029,6 +3059,138 @@ def test_valid_persistent_state(self):
...
+        self.assertValidationError(validator, false_testcase_4)
+        self.assertValidationError(validator, false_testcase_5)
+        self.assertValidationError(validator, false_testcase_6)
+        self.assertValidationError(validator, false_testcase_7)
+        self.assertValidationError(validator, false_testcase_8)
+        self.assertValidationError(validator, false_testcase_9)
+        self.assertValidationError(validator, false_testcase_10)
+        self.assertValidationError(validator, false_testcase_11)
+        self.assertValidationError(validator, false_testcase_12)
+        self.assertValidationError(validator, false_testcase_13)
```
thom311, line 3120: why do these variables `false_testcase_13` exist? It seems to make the code harder to read (and is more verbose).

module_utils/network_lsr/argument_validator.py
tyll: yes

tests/unit/test_network_connections.py
liangwen12year: we have 13 different dns options:
```    VALID_OPTIONS = [
        r"^attempts:([1-9]\d*|0)$",
        r"^debug$",
        r"^edns0$",
        r"^ndots:([1-9]\d*|0)$",
        r"^no-check-names$",
        r"^no-reload$",
        r"^no-tld-query$",
        r"^rotate$",
        r"^single-request$",
        r"^single-request-reopen$",
        r"^timeout:([1-9]\d*|0)$",
        r"^trust-ad$",
        r"^use-vc$",
    ]
```

module_utils/network_lsr/argument_validator.py
liangwen12year: the `enum_values` used everywhere in that file, I just want to keep it consistent

module_utils/network_lsr/argument_validator.py
liangwen12year: variable name VALID_MODES is widely  used in the file, I just want to keep it consistent. But maybe you are right. Updated!

module_utils/network_lsr/argument_validator.py
liangwen12year: updated

module_utils/network_lsr/argument_validator.py
liangwen12year: updated

liangwen12year: `str()` removed

module_utils/network_lsr/argument_validator.py
thom311: The name is used for things that are conceptionally similar and for which the name makes sense. In particular, it's a list of names (string) from a fixed set of possible names ("enums"). Here it's a list of regular expressions.

Also, python has nice functional aspects. So, maybe instead the new parameter should be `validate_element_fcn`, so you could do:

```python
         for (idx, v) in enumerate(value):
             try:
+                if self.validate_element_fcn is not None and not self.validate_element_fcn(v):
+                    raise ValidationError(
+                        name,
+                        "'%s' is not valid"
+                        % (value),
+                    )
                 vv = self.nested._validate(v, name + "[" + str(idx) + "]")
             except ValidationError as e:
                 raise ValidationError(e.name, e.error_message)
             result.append(vv)
```

And document `validate_element_fcn()` may either return False or raise ValidationError on error.

And you use it like:

```python
+                ArgValidatorList(
+                    "dns_options",
+                    nested=ArgValidatorStr("dns_options[?]"),
+                    validator_element_fcn= lambda v: any(re.match(patter, v) for pattern in ArgValidator_DictIP.VALID_OPTIONS),
+                    default_value=list,
+                ),
```

thom311: actually, `ArgValidatorList` already has something like `validate_element_fcn`: that's `self.nested`. So if you want to validate the elements, set the appropriate nested function. Meaning, the right solution is not
```python
                ArgValidatorList(
                    "dns_options",
                    nested=ArgValidatorStr("dns_options[?]"),
                    enum_values=ArgValidator_DictIP.VALID_OPTIONS,
                    default_value=list,
                ),
```
but
```python
                ArgValidatorList(
                    "dns_options",
                    nested=ArgValidatorIPDnsOptions("dns_options[?]"),
                    default_value=list,
                ),

```
or,
```python
                ArgValidatorList(
                    "dns_options",
                    nested=ArgValidatorStr("dns_options[?]", regex_list=[...]),
                    default_value=list,
                ),

```

liangwen12year: I am not very convinced why we should not  use list of regex inside enums

thom311: I don't understand what you mean.

`ArgValidatorList` is a validator of a some "list thingy". But the elements of the list are not necessarily strings. Hence, `ArgValidatorList` having a property with a list of regexes to validate strings seems wrong (because regex only make sense to validate strings, and the elements of `ArgValidatorList` may not be strings). I guess, you could add a subclass `ArgValidatorStringList`, but you don't need that. Instead, `ArgValidatorList` already has a validator for the list elements: `self.nested`. Set the right `self.nested` to validate the list elements with a regex.

Also, not sure what you mean with `list of regex inside enums`.

liangwen12year: for those lists that not containg strings, they also may not  need `enum_values` to validate their value, and right now it make more sense to include function `str()` here for those lists containing non-string type when `enum_values` is not `None` . Given the fact that python is short circuiting, I do not see any problem here for lists containing non-string type. 

the following is a list of regex inside enum_values:
```
    VALID_OPTIONS = [
        r"^attempts:([1-9]\d*|0)$",
        r"^debug$",
        r"^edns0$",
        r"^ndots:([1-9]\d*|0)$",
        r"^no-check-names$",
        r"^no-reload$",
        r"^no-tld-query$",
        r"^rotate$",
        r"^single-request$",
        r"^single-request-reopen$",
        r"^timeout:([1-9]\d*|0)$",
        r"^trust-ad$",
        r"^use-vc$",
    ]
```

thom311: I see it this way:

With all `ArgValidator` classes you have complete control about validating the element simply by overwriting `_validate_impl()` or `_validate_post()`. That is the most flexible one, and it requires you to subclass `ArgValidator`.

But we have common behaviors, like a string. Hence, there is reusable `ArgValidatorStr` implementation to validate a string. So, if you wanted to validate `ethernet.duplex` you could do
```
class ArgValidator_EthernetDuplex(ArgValidatorStr):
     def __init__(self, name):
           ArgValidatorStr.__init__(self, "duplex", default_value=None)
           value = ArgValidatorStr._validate_impl(self, value, name)
           if value not in ["half", "full"]:
                  raise ValidateError("duplex value must be either half or full")
           return value
```
But validating against a fixed set of strings (enums!!) is a common thing. Hence, `ArgValidatorStr` helps you with such common, simple cases and you can instead do:
```
   ArgValidatorStr("duplex", default_value=None, enum_values=['half', 'full'])
```

Sidenote: the name "enum_values" here has the meaning that these strings are from a fixed set of valid strings. I don't understand why you keep insisting on naming a list of regular expressions to validate the string as "enum_values".

`enum_values` parameter exists for convenience on top of the common `_validate_impl()` code and `ArgValidatorStr._validate_impl()` implements that to be re-used.

But adding a parameter `regex` to `ArgValidatorList()` seems a bit odd to me. Why such a convenience paratmer, which hardly makes sense for the majority of list validators? Why

```
                ArgValidatorList(
                    "dns_options",
                    nested=ArgValidatorStr("dns_options[?]"),
                    regex=ArgValidator_DictIP.REGEX_DNS_OPTIONS,
                    default_value=list,
                )
```
instead of
```
                ArgValidatorList(
                    "dns_options",
                    nested=ArgValidatorStr("dns_options[?]", regex = ArgValidator_DictIP.REGEX_DNS_OPTIONS),
                    default_value=list,
                )
```
It makes sense to me that `ArgValidatorStr` has a convenience checker `regex` instead of `ArgValidatorList`.

tests/unit/test_network_connections.py
thom311: I don't understand what this future "false_testcase_13.update()" would do. Are you talking about https://docs.python.org/3/library/stdtypes.html#dict.update ? And what do you mean with "louse coupling"?

module_utils/network_lsr/argument_validator.py
thom311: Yes, I like this!

There is still the `enum_values` argument in `ArgValidatorList`.


I suggested the name `regex`, because I thought maybe this could be either just one string (one "regex") or a list of regexes (like in the patch, where the list is combined with "OR"). So in the future we could do:

```python
        if self.regex is not None:
            r = self.regex
            if  isinstance(value, Util.STRING_TYPE):
                r = [r]
            if not any(re.match(x, value) for x in r):
                raise ValidationError(
                    name,
                    "is '%s' but must be one of '%s'"
                    % (value, "' '".join(sorted(self.regex))),
                )
```
and you could either do:
```python
     ArgValidatorStr(
           "dns_options[?]", regex=ArgValidator_DictIP.REGEX_DNS_OPTIONS
     )
```
or
```python
     ArgValidatorStr(
           "dns_options[?]", regex=r'^(use-vc|debug|...)$'
     )
```
with one regex only.


But that is not necessary now (because it's not used). But that is why I suggested "regex" as name and not "regex_list".

tyll: > `enum_values` does not seem a good name for a list of regular expressions.

`choices` might be a better name for enum_values to make it's purpose clear. This could be done in a refactor PR if there is a consensus.

tests/unit/test_network_connections.py
liangwen12year: for "loose coupling", I mean separate `self.assertValidationError(validator, { "dns_options": ["use1-vc2-use-vc"], })` into 

```python
false_testcase_13.update({
            "dns_options": ["usevc"],
        })
self.assertValidationError(validator, false_testcase_13)

```

we have the flexible control  of testing cases , rather than squashing them together

liangwen12year: "false_testcase_13.update()" is updating the input test cases. we can update the false_test_case into true_test_case, so that we can observe the difference between different cases more clearly.

module_utils/network_lsr/argument_validator.py
liangwen12year: Till, I suddenly remerbered, the previous order before sorting is the same order as we defined in NM

module_utils/network_lsr/argument_validator.py
```diff
@@ -148,6 +151,12 @@ def _validate_impl(self, value, name):
                 "is '%s' but must be one of '%s'"
                 % (value, "' '".join(sorted(self.enum_values))),
             )
+        if self.regex is not None and not any(re.match(x, value) for x in self.regex):
+            raise ValidationError(
+                name,
+                "is '%s' but must be one of '%s'"
```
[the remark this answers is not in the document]
liangwen12year: changed to `is '%s' which does not match the regex %s'`.

library/network_connections.py
```diff
@@ -993,7 +995,8 @@ def connection_create(self, connections, idx, connection_current=None):
                     s_ip4.add_dns(d["address"])
             for s in ip["dns_search"]:
                 s_ip4.add_dns_search(s)
-
+            for s in ip["dns_options"]:
+                s_ip4.add_dns_option(s)
```
thom311, line 999: dns-options property is a list of strings.

Most properties that are a list of something don't differentiate between an empty list and an unset list. However, for dns-options that makes a difference:

* leaving the property unset (the default), means that the DNS options can be configured as default values in NetworkManager.conf
* leaving the property explicitly set to an empty list, means that there are not DNS options set (and the defaults are not used).

The role should possibly fully define a profile. At least, for the properties that it supports. Hence, you should first once call `s_ip4.clear_dns_options(True)` so that the profile has NO DNS options set.


For initscripts, that difference doesn't exist. However, we often want that an ifcfg file also works when using it in NetworkManager. For that reason, it might make sense to explicitly set `ifcfg["RES_OPTIONS"] = ''`. But that is less clearly the desired things...

liangwen12year: updated, thanks

library/network_connections.py
```diff
@@ -467,6 +467,8 @@ def ifcfg_create(
                 ifcfg["DNS" + str(idx + 1)] = dns["address"]
             if ip["dns_search"]:
                 ifcfg["DOMAIN"] = " ".join(ip["dns_search"])
+            if ip["dns_options"]:
+                ifcfg["RES_OPTIONS"] = ""
```
thom311, line 471: now this change is incorrect.

Anyway, `grep RES_OPTIONS -R /etc/sysconfig/` shows that initscripts don't support this option either. It's like with ipv6_disabled: the initscripts provider needs to reject specifying such settings.

liangwen12year: inside the VM, I get this:
```
[root@ibm-p8-kvm-03-guest-02 ~]# grep RES_OPTIONS -R /etc/sysconfig/ 
/etc/sysconfig/network-scripts/ifcfg-eth0-1:RES_OPTIONS="rotate timeout:1"
```

cathay4t:
@liangwen12year Somehow the CI stuck again. Please amend the patch with zero changes: `git commit --amend --date="$(date -R)"`, and force push again. So the CI could retrigger. Thanks.

richm:
> @liangwen12year Somehow the CI stuck again. Please amend the patch with zero changes: `git commit --amend --date="$(date -R)"`, and force push again. So the CI could retrigger. Thanks.

I don't think anything is stuck yet.  I'm monitoring the CI jobs - they are proceeding, just slowly.

thom311 (approved):

cathay4t:
@richm Thanks.

PR #342, opened 2021-01-13
Support ipv6.method disabled in network role

liangwen12year:


liangwen12year:
will add test cases soon

tests/unit/test_network_connections.py
```diff
@@ -143,6 +143,7 @@ def setUp(self):
                 "gateway4": None,
                 "route_metric4": None,
                 "auto6": True,
+                "ipv6_disabled": False,
```
tyll, line 146: `auto6:True` and `ipv6_disabled:False` sounds like an invalid combination.

library/network_connections.py
```diff
@@ -972,6 +976,10 @@ def connection_create(self, connections, idx, connection_current=None):
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "manual")
             else:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "ignore")
+            
+            if ip["ipv6_disabled"]:
```
tyll, line 980: I am not convinced that introducing `ipv6_disabled` is the best approach here. As this code shows, the decision tree becomes unclear. If it is set, the method will be first "ignore" and then updated to "disabled". This is confusing.

tests/unit/test_network_connections.py
thom311: the argvalidators are supposed to either validate (and reject such a setting) or to normalize it -- whatever is more sensible.

Other parts of the code then can rely that parameters have sensible values (also with respect to each other).

liangwen12year: I deleted or changed into `"ipv6_disabled": None,`, `tox -e py37` tests failed.
Maybe thomas is right, we should validate the value in argvalidators first.

library/network_connections.py
liangwen12year: updated

module_utils/network_lsr/argument_validator.py
```diff
@@ -531,6 +533,13 @@ def _validate_post(self, value, name, result):
                 raise ValidationError(
                     name, "'dhcp4_send_hostname' is only valid if 'dhcp4' is enabled"
                 )
+        if result["ipv6_disabled"] is True:
+            result["auto6"] = None
```
thom311, line 537: if "ipv6_disable" is on, then you should check that the user didn't set `auto6:yes` and didn't configure any IPv6 addresses. That would be conflicting and should be rejected.

Also, a major point of `ArgValidator` is to *normalize* the values. In yaml, a setting like `ipv6_disabled` has three possible values: `True`, `False` and unspecified (`None`). After `_validate_post()`, `None` should be normalized to either True or False.

module_utils/network_lsr/argument_validator.py
```diff
@@ -531,6 +533,13 @@ def _validate_post(self, value, name, result):
...
+        if result["auto6"] is not None:
```
thom311, line 538: at this point, `result["auto6"]` cannot be None.

liangwen12year: if `result["auto6"]` is not None. then the error `"'auto6' and 'ipv6_disabled' are mutually exclusive"`  will be invoked everytime based on the following testing code:
```yaml
        network_connections:
          - name: eth0
            type: ethernet
            ip:
              ipv6_disabled: true
```

module_utils/network_lsr/argument_validator.py
liangwen12year: like you mentioned in IRC,  if we don't specify any methods, then "auto" will be enabled by default (as ArgValidator already normalizes). then conflicts/error happened everytime when we set "ipv6_disable" on.

thom311: this sets `result["auto6"] = None`. That cannot be right because `_validate_post()` should normalize `auto6` to a boolean and not leave it undecided. It is only left undecided during `_validate_impl()`, so we can distinguish here between "unset" and explicitly set.

module_utils/network_lsr/argument_validator.py
thom311: above there is the code
```
        if result["auto6"] is None:  
            result["auto6"] = not any(
                [a for a in result["address"] if a["family"] == socket.AF_INET6]
            )
```
so here the condition `if result["auto6"] is not None:` is NEVER satisfied.

thom311: > so here the condition if result["auto6"] is not None: is NEVER satisfied.

Wrong of course. "is ALWAYS satisfied".

library/network_connections.py
thom311: > As this code shows, the decision tree becomes unclear. If it is set, the method will be first "ignore" and then updated to "disabled". This is confusing.

The decision tree becomes clear if `ArgValidator` resolves all inconsistencies. Yes, `"ipv6_disabled:True"` and `"auto6:True"` conflicts, but `ArgValidator` should reject such configuration as valid and guarantee that the result of the validation is consistent.

library/network_connections.py
```diff
@@ -966,12 +970,15 @@ def connection_create(self, connections, idx, connection_current=None):
...
 
-            if ip["auto6"]:
+            if ip["ipv6_disabled"]:
+                s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "disabled")
+            elif ip["auto6"]:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "auto")
             elif addrs6:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "manual")
             else:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "ignore")
```
thom311, line 980: in a follow-up patch, it should be considered that "ignore" gets dropped and instead have "link-local" here. Possibly "ignore" setting should not be configurable via the role. That would be a change in behavior and warrants a separate commit. But maybe you can add a `# TODO` comment now.

module_utils/network_lsr/argument_validator.py
liangwen12year: You said "After _validate_post(), None should be normalized to either True or False.", then `auto6` is either true or false , `ipv6_disabled` is either true or false. I did not see how we can achieve  `'auto6' and 'ipv6_disabled' are mutually exclusive`.

library/network_connections.py
thom311: done.

module_utils/network_lsr/argument_validator.py
thom311: I think the validation code is good now.

module_utils/network_lsr/argument_validator.py
thom311: Is it clear now? I think the code is now good, marking this question as "resolved". Please unresolve is something is wrong.

tests/unit/test_network_connections.py
thom311: > auto6:True and ipv6_disabled:False sounds like an invalid combination.

I think Till meant "auto6:True and ipv6_disabled:**True** sounds like an invalid combination."

Anyway, in the latest version of the code this is fixed and rejected as invalid.

library/network_connections.py
thom311: > I am not convinced that introducing ipv6_disabled is the best approach here.

Maybe. This question is about naming of parameters (and which form a parameter should have). The choice to have here a new boolean `ipv6_disabled` seems sensible to me. But no doubt, it could also be different.

Is this question stil unresolved?

library/network_connections.py
```diff
@@ -457,6 +457,10 @@ def ifcfg_create(
                 ifcfg["IPV6_AUTOCONF"] = "no"
             else:
                 ifcfg["IPV6INIT"] = "no"
+
+            if ip["ipv6_disabled"]:
+                ifcfg["IPV6_DISABLED"] = "yes"
```
thom311, line 462: `grep -R IPV6_DISABLED /etc/sysconfig/` shows that initscripts don't support this variable. This is an extension added by NetworkManager, but when the role writes ifcfg-rh files, it must write them so that initscripts can handle them.

I don't think that initscript support `ipv6_disabled`. At least not as part of ifcfg-files, you would have to write a sysctl files (and possibly reboot). So, it's fine if the ifcfg-rh backend does not support this option. See `validate_connection_one()` how to handle that.

library/network_connections.py
tyll: The problem I see is with the redundancy, for example auto6:yes implies ipv6_disabled:no. Therefore it would be nicer to find a solution where the configuration format ensures that there are no such conflicts but maybe it is not possible. I guess ipv6_link_local:yes|no has similar problems - but maybe it would be more descriptive about the impact of the option. Or are static IPs without link-local IPs a valid configuration for IPv6?

thom311: they are not. You always need a link-local IPv6 address (except with `ipv6_disabled=yes`)

In the current branch,
```
    ipv6_disabled = no
    auto6 = no
    addresses = []
```
translates to `ipv6.method=ignored`, but it really should be instead `ipv6.method=link-local`.

I don't see a problem with "redundancy". Sure, in settings like

```
   ipv6_disabled = no
   auto6 = yes
```
the `ipv6_disabled` is redundant. But you can already do
```
   ipv6_disabled = no
   auto6 = yes
   addresses = []
```
where both `addresses` and `ipv6_disabled` is redundant too. That seems not avoidable. It would be annoying to fail:
```
    auto6 = yes
    addresses = []
```
with: "ERROR: cannot configure an empty list of addresses because that redundant to not specifying the option."


On the other hand, conflicting settings are rejected as invalid by the branch.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1569,3 +1602,12 @@ def validate_connection_one(self, mode, connections, idx):
...
                 )
+
+        # initscripts does not support ipv6_disabled, so raise errors when network
+        # provider is initscripts
+        if connection["ip"]["ipv6_disabled"]:
+            if mode == self.VALIDATE_ONE_MODE_INITSCRIPTS:
+                raise ValidationError.from_connection(
+                    idx,
+                    "ipv6.method disabled is not supported by initscripts.",
+                )
```
liangwen12year, line 1613: @thom311 , I added the error handler for initscripts

thom311: looks good. But the error message doesn't seem very suitable. "ipv6.method disabled" is something of NetworkManager. The paramter in the role is called "ip.ipv6_disabled" and the error message should refer to that.

liangwen12year: already changed into `"ip.ipv6_disabled is not supported by initscripts."`

library/network_connections.py
thom311: these two lines are still here in the latest version.

liangwen12year: these two lines were removed

thom311 (approved):

cathay4t (changes requested):
CI is failing on the test case you added.

liangwen12year:
> CI is failing on the test case you added.
I resolved the conflcts,  and I will monitor CI to see if it will fail

liangwen12year:
> > CI is failing on the test case you added.
> 
> I resolved the conflcts, and I will monitor CI to see if it will fail

Looks good now!

ffmancera (approved):
LGTM, please could you rebase it? A fix has been introduced for the tox tests.

richm:
Looks like you have an ansible-lint issue - you are using the `shell` module - you should use `set -euo pipefail` as the first line of your shell for additional bash checks

liangwen12year:
> Looks like you have an ansible-lint issue - you are using the `shell` module - you should use `set -euo pipefail` as the first line of your shell for additional bash checks

thanks for letting me know, updated.

ffmancera (approved):
LGTM, thanks!

ffmancera:
[citest bad]

PR #345, opened 2021-01-20
remove ansible 2.7 support from molecule

richm:
ansible 2.7 has been deprecated
https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html
We need to remove ansible 2.7 from molecule testing in order to
* make room for ansible 2.10 testing
* move to molecule v3

ansible 2.8 and later support platform-python on el8 and later
so we don't have to handle that case explicitly by setting
ansible_python_interpreter for centos8 in molecule.yml


richm:
ok to merge?

richm:
ok to merge?

richm:
ok to merge?

PR #350, opened 2021-02-09
fix collection lookup problem

richm:
This fixes the following error:
```
/bin/sh: line 9: cd: /root/.ansible/collections: No such file or directory
```
The fix is to ensure the directory exists before attempting
to `cd` to the directory.

### DIFF
--- a/tests/tasks/get_modules_and_utils_paths.yml
+++ b/tests/tasks/get_modules_and_utils_paths.yml
@@ -50,6 +50,7 @@
       fi
     done
     for dir in {{ collection_paths | join(" ") }}; do
+      if [ ! -d "$dir" ]; then continue; fi
       cd "$dir"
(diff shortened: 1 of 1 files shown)


nhosoi (approved):
lgtm

I see. When we have multiple collection paths like this [/root/.ansible/collections /usr/share/ansible/collections], and the first path does not exist but the second one is the real collection path, we had this issue. And your fix solves the error.

nhosoi:
I believe the failure in "tox / python (3.6) (pull_request) Failing after 4m — python (3.6) " is from black in the collection tests. I could reproduce it locally. May I remove it from runcollection.sh in tox-lsr?

I tested the combination of this PR and tox-lsr with no black in runcollection.sh and verified the failure disappears. That implies this PR is ready to merge.

liangwen12year (approved):
looks good to me! Approved.

PR #351, opened 2021-02-09
use molecule v3, drop v2 - use tox-lsr 2.1.2

richm:
Switch to molecule v3.  This drops support for v2.
This requires tox-lsr 2 or later.
Upgrade to tox-lsr 2.1.2 to pick up fix for collection dependency
on python six.

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -3,9 +3,7 @@
 dependency:
   name: galaxy
 driver:
-  name: docker
-lint:
-  enabled: false
+  name: ${LSR_MOLECULE_DRIVER:-docker}
 platforms:
   - name: centos-6
     image: registry.centos.org/centos:6
@@ -28,8 +26,6 @@ platforms:
 provisioner:
   name: ansible
   log: true
-  lint:
-    enabled: false
   playbooks:
     converge: ../../tests/tests_default.yml
 scenario:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -3,7 +3,7 @@ name: tox
 on:  # yamllint disable-line rule:truthy
   - pull_request
 env:
-  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@1.0.2"
+  TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.1.2"
   LSR_ANSIBLES: 'ansible==2.8.* ansible==2.9.*'
   LSR_MSCENARIOS: default
   # LSR_EXTRA_PACKAGES: "libdbus-1-dev libgirepository1.0-dev python3-dev"


richm:
@nhosoi same problem here with `black` and collection converted python code

nhosoi:
Sadly, it seems `# noqa:E501` is no longer available (maybe, not supported in the new version?)

As @rjeffman suggested in https://github.com/linux-system-roles/certificate/pull/78, instead of adding `# noqa E501` at the end of the problematic line, we have to wrap them with `# fmt: off` and `# fmt: on`.

Is it ok to open a pr against  richm:molecule-v3-tox-lsr-2.1.2, @richm?

richm:
> Sadly, it seems `# noqa:E501` is no longer available (maybe, not supported in the new version?)
> 
> As @rjeffman suggested in [linux-system-roles/certificate#78](https://github.com/linux-system-roles/certificate/pull/78), instead of adding `# noqa E501` at the end of the problematic line, we have to wrap them with `# fmt: off` and `# fmt: on`.

The problem with that is that we want `fmt: on` for all "original" source code.  We only want to use `fmt: off` for the collection converted source code.  I'm not sure how to do that.  It would be quite difficult I think to make the lsr_role2collection code smart enough to insert these in the correct places in the converted code.

> 
> Is it ok to open a pr against richm:molecule-v3-tox-lsr-2.1.2, @richm?

not at this time.  I think we should merge this PR as-is, then figure out the collection/black problem.

nhosoi:
Just a note...
>  We only want to use fmt: off for the collection converted source code.

This is ideal, but we skipped a long line error by setting `# noqa:E501` in the original source code... (Right, fmt: off would skip all the tests, so the consequence is wider, though...)

richm:
> Just a note...
> 
> > We only want to use fmt: off for the collection converted source code.
> 
> This is ideal, but we skipped a long line error by setting `# noqa:E501` in the original source code... (Right, fmt: off would skip all the tests, so the consequence is wider, though...)

right - my concern is that we are too aggressive in suppressing errors - I think suppressing line length checking, on a line-by-line basis, is fine - and even suppressing line length checking in an entire file is fine - but `black` formatting is tricky

molecule/default/molecule.yml
```diff
@@ -3,9 +3,7 @@
 dependency:
   name: galaxy
 driver:
-  name: docker
-lint:
-  enabled: false
+  name: ${LSR_MOLECULE_DRIVER:-docker}
```
liangwen12year, line 6: what is the `LSR_MOLECULE_DRIVER` stands for? Is there any link available for me to get reference?

nhosoi: The variable id described in https://github.com/linux-system-roles/tox-lsr/#using-setenv-and-environment-variables.

LSR stands for Linux System Roles. :)

nhosoi (approved):
lgtm
A tox failure should be fixed by https://github.com/linux-system-roles/tox-lsr/pull/22.

molecule/default/molecule.yml
richm: > what is the `LSR_MOLECULE_DRIVER` stands for? Is there any link available for me to get reference?

so you can do something like `LSR_MOLECULE_DRIVER=podman tox -e molecule` to use `podman` instead of the default `docker`

PR #354, opened 2021-02-16
Fixing interface not found failure in tests_eth_dns_support.yml

liangwen12year:


liangwen12year:
For future reference: we should create a dedicated test interface and remove the interface after completing the test.  
The following link is sample code for creating a test interface:
 https://github.com/linux-system-roles/network/blob/main/tests/playbooks/tests_ethernet.yml#L22

ffmancera (approved):
LGTM, thanks!

PR #355, opened 2021-02-19
Add support for ETHTOOL Ring option in Network Role

liangwen12year:


richm:
please rebase to latest `main` branch

richm:
If this PR is for a new feature, could you add something to the README.md to describe the new feature?

library/network_connections.py
```diff
@@ -2015,21 +2042,47 @@ def _check_ethtool_setting_support(self, idx, connection):
...
+        # if specified_features and not hasattr(NM, "SettingEthtool"):
+        #     self.log_fatal(idx, "ethtool.features specified but not supported by NM")
+
+        # for feature, setting in specified_features.items():
+        #     nm_feature = nm_provider.get_nm_ethtool_feature(feature)
+        #     if not nm_feature:
+        #         self.log_fatal(
+        #             idx, "ethtool feature %s specified but not support by NM" % feature
+        #         )
+
```
ffmancera, line 2085: Why is this code commented? If it is not needed, please remove it.

module_utils/network_lsr/argument_validator.py
```diff
@@ -584,7 +585,10 @@ def __init__(self):
         ArgValidatorDict.__init__(
             self,
             name="ethtool",
-            nested=[ArgValidator_DictEthtoolFeatures()],
+            nested=[
+                ArgValidator_DictEthtoolFeatures(),
+                ArgValidator_DictEthtoolSetring()
```
ffmancera, line 590: I think the correct name is `ArgValidator_DictEthtoolSetRing()` according to the naming use on the project. Could you confirm me this? @tyll 

Thanks!

module_utils/network_lsr/argument_validator.py
```diff
@@ -800,6 +804,38 @@ def __init__(self):
         )
 
 
+
```
ffmancera, line 807: I think this line is not needed.

module_utils/network_lsr/argument_validator.py
```diff
@@ -800,6 +804,38 @@ def __init__(self):
...
+        )
+        self.default_value = dict(
+            [(k, v.default_value) for k, v in self.nested.items()]
+        )
+
+
+
+
+
+
```
ffmancera, line 836: Please, remove this empty lines, they are not needed. Thanks

module_utils/network_lsr/nm_provider.py
```diff
@@ -21,3 +22,18 @@ def get_nm_ethtool_feature(name):
 
     feature = getattr(Util.NM(), name, None)
     return feature
+
+
+def get_nm_ethtool_setring(name):
```
ffmancera, line 27: Maybe `get_nm_ethtool_set_ring`?

module_utils/network_lsr/nm_provider.py
```diff
@@ -21,3 +22,18 @@ def get_nm_ethtool_feature(name):
...
+    """
+    Translate ethtool set-ring into Network Manager name
```
ffmancera, line 29: `Translate ethtool set-ring option into NetworkManager attribute name` I think is a better description.

ffmancera (changes requested):
Hi! Thanks for this work. Please, could check why the CI is failing? In addition, it would be nice to have a detailed description in the commit message.

Thank you!

module_utils/network_lsr/nm_provider.py
tyll: Is `set` really part of option? It seems to be just `ring` (compare to feature).

module_utils/network_lsr/argument_validator.py
tyll: Actually, it seems to me that `ArgValidator_DictEthtoolRing` would be the correct name.

module_utils/network_lsr/nm_provider.py
tyll: Just `ring`

tyll:
@liangwen12year this should be rather easy to add but the PR is very outdated. It seems to contain changes that were already added.

module_utils/network_lsr/nm_provider.py
liangwen12year: Changed into `Translate ethtool ring option into NetworkManager attribute name`

module_utils/network_lsr/argument_validator.py
liangwen12year: Empty line removed.

module_utils/network_lsr/argument_validator.py
liangwen12year: Empty line removed.

examples/ethtool_setring.yml
```diff
@@ -0,0 +1,25 @@
...
+              set-ring:
+                rx: 128
+                rx_mini: 128
+                rx_jumbo: 128
+                tx: 128
+
+
+
+
+
```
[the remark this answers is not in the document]
liangwen12year: Space removed.

module_utils/network_lsr/nm_provider.py
liangwen12year: Changed into `get_nm_ethtool_ring`.

module_utils/network_lsr/argument_validator.py
liangwen12year: Changed into `ArgValidator_DictEthtoolRing`.

library/network_connections.py
liangwen12year: The commented code was removed.

tests/ensure_provider_tests.py
```diff
@@ -76,6 +76,10 @@
         MINIMUM_VERSION: "'1.25.1'",
         "comment": "# NetworkManager 1.25.1 introduced ethtool coalesce support",
     },
+    "playbooks/tests_ethtool_ring.yml": {
+        MINIMUM_VERSION: "'1.25.2'",
+        "comment": "# NetworkManager 1.25.2 introduced ethtool ring support",
+    },
```
liangwen12year, line 82: Get hint from
```
[wenliang@localhost NetworkManager]$ git grep -w 'NMEthtoolRingState' 1.25.1-dev 
[wenliang@localhost NetworkManager]$ git grep -w 'NMEthtoolRingState' 1.25.2-dev 
1.25.2-dev:src/devices/nm-device.c:     NMEthtoolRingState *ring;
1.25.2-dev:src/devices/nm-device.c:     gs_free NMEthtoolRingState *ring = NULL;
1.25.2-dev:src/devices/nm-device.c:     NMEthtoolRingState ring_old;
1.25.2-dev:src/devices/nm-device.c:     NMEthtoolRingState ring_new;
1.25.2-dev:src/platform/nm-platform-utils.c:                  NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform-utils.c:                            NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform-utils.c:                  const NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform-utils.c:                            const NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform-utils.h:                                     NMEthtoolRingState *ring);
1.25.2-dev:src/platform/nm-platform-utils.h:                                     const NMEthtoolRingState *ring);
1.25.2-dev:src/platform/nm-platform.c:                                   NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform.c:                              const NMEthtoolRingState *ring)
1.25.2-dev:src/platform/nm-platform.h:typedef struct _NMEthtoolRingState NMEthtoolRingState;
1.25.2-dev:src/platform/nm-platform.h:                                            NMEthtoolRingState *ring);
1.25.2-dev:src/platform/nm-platform.h:                                       const NMEthtoolRingState *ring);
```

library/network_connections.py
```diff
@@ -428,6 +428,21 @@ def ifcfg_create(
                 connection["interface_name"],
                 " ".join(configured_coalesce),
             )
+        ethtool_ring = connection["ethtool"]["ring"]
+        configured_ring = []
+        for ring, setting in ethtool_ring.items():
+            if setting is not None:
+                if isinstance(setting, bool):
+                    setting = int(setting)
```
tyll, line 436: this does not seem to make sense

liangwen12year: The data type for `ethtool.ring-rx`, `ethtool.ring-tx`, `ethtool.ring-rx-mini`, `ethtool.ring-rx-jumbo` value should be integer.
It is not necessary though, argvalidator will raise an error if the corresponding value is not integer:
```
TASK [linux-system-roles.network : Configure networking connection profiles] *****************************************************************************************************************************************************************
task path: /home/wenliang/network/tests/roles/linux-system-roles.network/tasks/main.yml:81
fatal: [Fedora-Cloud-Base-33-20210106.0.x86_64.qcow2]: FAILED! => {"_invocation": {"module_args": {"__debug_flags": "", "connections": [{"autoconnect": false, "ethtool": {"ring": {"rx": 128.6}}, "ip": {"auto6": false, "dhcp4": false}, "name": "testnic1", "type": "ethernet"}], "force_state_change": false, "ignore_errors": false, "provider": "nm"}}, "changed": false, "msg": "fatal error: configuration error: connections[0].ethtool.ring.rx: must be an integer number but is '128.6'", "stderr": "\n", "stderr_lines": [""]}
```

liangwen12year: Actually the above code snippet is necessary, if we define the connection profile as the following, argvalidator will not raise any error

Code:
```yaml
          vars:
            network_connections:
              - name: "{{ interface }}"
                type: ethernet
                autoconnect: no
                ip:
                  dhcp4: no
                  auto6: no
                ethtool:
                  ring:
                    rx: True
        - name: Get profile's ring options
          command: nmcli -g ethtool.ring-rx c show {{ interface }}
          register: with_ring
        - name: Assert ring options set in profile
```

Result:
```
TASK [Assert ring options set in profile] ****************************************************************************************************************************************************************************************************
task path: /home/wenliang/network/tests/playbooks/tests_ethtool_ring.yml:67
fatal: [Fedora-Cloud-Base-33-20210106.0.x86_64.qcow2]: FAILED! => {
    "assertion": "with_ring.stdout == '128'",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

liangwen12year:
> Hi! Thanks for this work. Please, could check why the CI is failing? In addition, it would be nice to have a detailed description in the commit message.
> 
> Thank you!

Detailed description added, PR was rebased, conflict resolved, CI failure fixed.

library/network_connections.py
thom311: >  argvalidator will not raise any error

That is because of how `ArgValidatorNum._validate_impl()` is implemented:

```python
        v = None
        try:
            if isinstance(value, self.numeric_type):
                v = value
            else:
                v2 = self.numeric_type(value)
                # ^^^^^^^
                if isinstance(value, Util.STRING_TYPE) or v2 == value:
                    v = v2
```
It tries to convert the value into a number. If it is not desirable to convert bools here (it probably is not), then fix ArgValidatorNum's implementation (note that this is potentially a break in the API, but probably ok).


---

but I don't understand why
```python
+                if isinstance(setting, bool):
+                    setting = int(setting)
```
would ever be necessary. ArgValidatorNum should normalize this to either `None` or `int` type. If not, **that** is the bug, and needs to be investigated.

thom311: ah, that was already in code before:

```python
                 if isinstance(setting, bool):
                     setting = int(setting)
                 configured_coalesce.append(
                     "%s %s" % (coalesce.replace("_", "-"), setting)
```
That's wrong! `ArgValidatorNum` MUST normalize the value to either `None` or `int` (and it does already, or why not?). No need to assume that `ArgValidator` would not work properly and handle cases that in fact must not happen.

Likewise,
```
+                    if setting is None:
+                        s_ethtool.option_set(nm_ring, None)
+                    else:
+                        s_ethtool.option_set_uint32(nm_ring, int(setting))
                                                            # ^^^ WRONG
```

thom311: Rational: we want ArgValidator to normalize things and rely on it's result. It has an "API" or expected behavior, where we trust that it does perform the desired normalization. For the `ArgValidatorNum` setting, we do expect that it does what it's supposed to do, and don't try to handle cases where it would not (which would be due to a bug).

In other words, we should not add code to handle situations, that can only happen in face of a bug -- if a bug would be there, we should fix the bug instead.

liangwen12year: Indeed, following code in `argvalidatorNum` will normalize the value into `None` or `int` type. Even if we define the value to be `Bool`(probably will never be the case),   `self.numeric_type(Bool)` will also normalize the value to be either 1 or 0. 
Let's remove `if isinstance(setting, bool): setting = int(setting)`. Thanks for clarifying.
```
        v = None
        try:
            if isinstance(value, self.numeric_type):
                v = value
            else:
                v2 = self.numeric_type(value)
                # ^^^^^^^
                if isinstance(value, Util.STRING_TYPE) or v2 == value:
                    v = v2
```

liangwen12year:
> If this PR is for a new feature, could you add something to the README.md to describe the new feature?

Description was added here, https://github.com/linux-system-roles/network/blob/b1f89f4ad86101bcb82cb1a6cec193d4b3062983/README.md#ethtool

library/network_connections.py
```diff
@@ -428,6 +428,19 @@ def ifcfg_create(
                 connection["interface_name"],
                 " ".join(configured_coalesce),
             )
+        ethtool_ring = connection["ethtool"]["ring"]
+        configured_ring = []
+        for ring, setting in ethtool_ring.items():
+            if setting is not None:
```
thom311, line 434: yes, right. But as I tried to say earlier, above is also
```python
         for coalesce, setting in ethtool_coalesce.items():
             if setting is not None:
                 if isinstance(setting, bool):
                     setting = int(setting)
                 configured_coalesce.append(
```
with the same problem. Please fix this pre-existing code in a separate commit.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1008,6 +1009,31 @@ def __init__(self):
...
 
 
+class ArgValidator_DictEthtoolRing(ArgValidatorDict):
+    def __init__(self):
+        ArgValidatorDict.__init__(
+            self,
+            name="ring",
+            nested=[
+                ArgValidatorNum(
+                    "rx", val_min=0, val_max=UINT32_MAX, default_value=None
```
thom311, line 1019: at https://github.com/linux-system-roles/network/pull/355#discussion_r615212067, I suggest to fix ArgValidatorNum() to reject boolean values. In a separate commit too, and with a comment in the commit message that this bug fix potentially breaks previously "working" playbooks (but realistically, they were not working, because setting 0 or 1 was unlikely intended).

library/network_connections.py
liangwen12year: IMHO, I do not think we should remove  ` if isinstance(setting, bool): setting = int(setting)` for coalesce setting. Because the data type for `adaptive_rx` and `adaptive_tx` in Network role is bool, see
```python
                ArgValidatorBool("adaptive_rx", default_value=None),
                ArgValidatorBool("adaptive_tx", default_value=None),
```
But In NetworkManager, we have to set it to either 0 or 1 when modify the `ethtool.coalesce-adaptive-rx` and `ethtool.coalesce-adaptive-tx` setting.
```bash
[root@localhost ~]# nmcli connection modify testnic1 ethtool.coalesce-adaptive-rx True
Error: failed to modify ethtool.coalesce-adaptive-rx: 'True' is out of range [0, 4294967295].# Is anything wrong with the error msg?
[root@localhost ~]# nmcli connection modify testnic1 ethtool.coalesce-adaptive-rx 1
```

library/network_connections.py
```diff
@@ -948,6 +961,14 @@ def connection_create(self, connections, idx, connection_current=None):
...
                     else:
                         s_ethtool.option_set_uint32(nm_coalesce, int(setting))
+            for ring, setting in connection["ethtool"]["ring"].items():
+                nm_ring = nm_provider.get_nm_ethtool_ring(ring)
+
+                if nm_ring:
+                    if setting is None:
+                        s_ethtool.option_set(nm_ring, None)
+                    else:
+                        s_ethtool.option_set_uint32(nm_ring, setting)
```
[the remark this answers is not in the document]
liangwen12year: yes, right. But as I tried to say earlier, above is also
```python
         for coalesce, setting in ethtool_coalesce.items():
             if setting is not None:
                 if isinstance(setting, bool):
                     setting = int(setting)
                 configured_coalesce.append(
```
with the same problem. Please fix this pre-existing code in a separate commit.

module_utils/network_lsr/argument_validator.py
liangwen12year: Separated commit was created and test cases was also added.

tests/ensure_provider_tests.py
tyll: Actually, the tests should also run for the initscripts provider since that is getting support, too. So limiting this only to NM is wrong but also wrong for the other ethtool settings. :-/

tests/playbooks/tests_ethtool_ring.yml
```diff
@@ -0,0 +1,102 @@
...
+                  dhcp4: no
+                  auto6: no
+        - name: Get profile's ring options
+          command: nmcli -g ethtool.ring-rx c show {{ interface }}
+          register: no_ring
+        - name: "ASSERT: The profile does not contain ring options"
+          assert:
+            that: no_ring.stdout | length == 0
+        - name: >-
+            TEST: I can set rx.
```
tyll, line 48: If possible, it would be great to set all four settings.

library/network_connections.py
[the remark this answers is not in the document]
tyll: Great observation @liangwen12year - there might be a problem with initscripts support for coalesce, though, since I guess it needs on or off for boolean values instead of 1/0 (not sure).

tests/ensure_provider_tests.py
liangwen12year: Yes, the Ethtool Ring option was supported by initscripts[1], How do you know Coalesce option was also supported by initscripts? 

```bash
[root@localhost ~]# grep -R ETHTOOL_OPTS $(rpm -ql network-scripts)
/etc/sysconfig/network-scripts/network-functions:    for opts in $ETHTOOL_OPTS ; do
/etc/sysconfig/network-scripts/ifcfg-testnic1:ETHTOOL_OPTS="-C testnic1 adaptive-rx 1"
/etc/sysconfig/network-scripts/network-functions:    for opts in $ETHTOOL_OPTS ; do
/usr/share/doc/network-scripts/sysconfig.txt:    ETHTOOL_OPTS=...
/usr/share/doc/network-scripts/sysconfig.txt:        ETHTOOL_OPTS="speed 100 duplex full autoneg off"
/usr/share/doc/network-scripts/sysconfig.txt:      ETHTOOL_OPTS="-K ${DEVICE} tso on; -G ${DEVICE} rx 256 tx 256"
/usr/share/doc/network-scripts/sysconfig.txt:    ETHTOOL_OPTS=...
/usr/share/doc/network-scripts/sysconfig.txt:        ETHTOOL_OPTS="speed 100 duplex full autoneg off"
/usr/share/doc/network-scripts/sysconfig.txt:      ETHTOOL_OPTS="-K ${DEVICE} tso on; -G ${DEVICE} rx 256 tx 256"
```

tyll: Take a look at the lines after

```
/etc/sysconfig/network-scripts/network-functions:    for opts in $ETHTOOL_OPTS ; do
```

tests/playbooks/tests_ethtool_ring.yml
liangwen12year: Four settings have been set.

tests/unit/test_network_connections.py
```diff
@@ -376,6 +376,12 @@ def test_validate_int(self):
         v = network_lsr.argument_validator.ArgValidatorNum("state", required=True)
         self.assertValidationError(v, None)
 
+        v = network_lsr.argument_validator.ArgValidatorNum(
+            "rx", val_min=0, val_max=0xFFFFFFFF, default_value=None
+        )
```
tyll, line 390: Is the new validator required here? Can you reuse the previous one?

module_utils/network_lsr/argument_validator.py
```diff
@@ -271,7 +271,9 @@ def __init__(  # pylint: disable=too-many-arguments
     def _validate_impl(self, value, name):
         v = None
         try:
-            if isinstance(value, self.numeric_type):
```
tyll, line 274: Please provide more context in the commit message. Currently, it needs more context to understand why it is mentioning 0 or 1.

module_utils/network_lsr/argument_validator.py
```diff
@@ -271,7 +271,9 @@ def __init__(  # pylint: disable=too-many-arguments
...
+            if isinstance(value, bool):
+                pass
+            elif isinstance(value, self.numeric_type):
                 v = value
             else:
```
tyll, line 278: Seems cleaner to me to avoid pass and make this

`elif not isinstance(value, bool)`

@thom311 what other casts ccould happen here? I guess from other numeric ones to the one that is used here and from string types to the numeric one are expected. Dicts/lists/tuples raise an exception with int at least.

liangwen12year: `if isinstance(True, int)` is evaluated to `True`.

module_utils/network_lsr/argument_validator.py
liangwen12year: I explained as: 
```
In `./library/network_connections.py`, for ethtool coalesce setting,
there is unnecessary code snippet `if isinstance(setting, bool): setting = int(setting)`,
which convert the setting value from boolean into integer.
`int(False)` is evaluated to 0 , `int(True)` is evaluated to 1.
Therefore, we need to fix `ArgValidatorNum()` to reject boolean values
when integer values are expected for the setting. This bug fix
potentially breaks previously "working" playbooks (but realistically,
they were not working, because setting 0 or 1 was unlikely intended).
```

module_utils/network_lsr/argument_validator.py
richm: https://stackoverflow.com/questions/60769919/isinstancefalse-int-returns-true

tests/unit/test_network_connections.py
liangwen12year: New validator is not required, I reused the previous validator, thanks.

module_utils/network_lsr/argument_validator.py
thom311: This explanation does not seem right.

This is not about coalesce setting or what the mentioned code snippet does. It's primarily about how `ArgValidatorNum` should behave.

`ArgValidatorNum` was always supposed to give a value of type `self.numeric_type` -- except the default value (which might be `None` or even something else altogether).

There was a bug with 
```python
            if isinstance(value, self.numeric_type):
                 v = value
```
because `bool` is a subclass of `int` in python (which I was not aware of). But I don't thik that a `bool` should be accepted as a valid input for a number -- at least not by default (maybe in some cases). The bug was it would wrongly accept a boolean value, and it would wrongly pass it on (and not normalize it to `numeric_type`).


how about:
```python
            if self.numeric_type is not bool and isinstance(value, bool):
                # bool can (probably) be converted to self.numeric_type,
                # but here we don't want to accept a boolean value.
                pass
            elif isinstance(value, self.numeric_type):
                 v = self.numeric_type(value)
             else:
                 v2 = self.numeric_type(value)
                 if isinstance(value, Util.STRING_TYPE) or v2 == value:
                     v = v2
```

liangwen12year: Previously, in `ArgvalidaorNum` we did not normalize the boolean value into `numeric_type`. And those playbooks can still run without raising any error just because we did not wrongly pass boolean value when `numeric_type` is expected. Therefore, this bug fix did not break any playbooks at all.

tests/ensure_provider_tests.py
```diff
@@ -72,10 +72,6 @@
         MINIMUM_VERSION: "'1.20.0'",
         "comment": "# NetworKmanager 1.20.0 added support for forgetting profiles",
     },
-    "playbooks/tests_ethtool_coalesce.yml": {
-        MINIMUM_VERSION: "'1.25.1'",
-        "comment": "# NetworkManager 1.25.1 introduced ethtool coalesce support",
-    },
```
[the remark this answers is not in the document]
liangwen12year: New PR was created to address the above problem, thanks.

module_utils/network_lsr/argument_validator.py
tyll: Interesting, thank you. Does it make sense to change `isinstance(value, self.numeric_type)` to `type(value) == self.numeric_type`, then?

liangwen12year: Yes, I think we could. With the current implementation, we can also reject bool safely. Either way is fine.
```python
            if self.numeric_type is not bool and isinstance(value, bool):
                # bool can (probably) be converted to self.numeric_type,
                # but here we don't want to accept a boolean value.
                pass
            elif isinstance(value, self.numeric_type):
                 v = self.numeric_type(value)
            else:
                 v2 = self.numeric_type(value)
                 if isinstance(value, Util.STRING_TYPE) or v2 == value:
                     v = v2
```

tests/ensure_provider_tests.py
```diff
@@ -89,6 +85,15 @@
     },
     "playbooks/tests_wireless_plugin_installation.yml": {},
 }
+# NM_DEMANDING_TESTS is used to store the test playbooks which are demanding for NM
+# minimum version or extra running condition, test playbooks in NM_DEMANDING_TESTS can
+# also run with initscripts provider
+NM_DEMANDING_TESTS = {
```
tyll, line 91:
```suggestion
NM_CONDITIONAL_TESTS = {
```

tests/ensure_provider_tests.py
```diff
@@ -114,11 +119,19 @@
 def create_nm_playbook(test_playbook):
     fileroot = os.path.splitext(os.path.basename(test_playbook))[0]
     nm_testfile = fileroot + "_nm.yml"
+    if NM_DEMANDING_TESTS.get(test_playbook, {}) != {}:
```
tyll, line 122:
```suggestion
    if test_playbook in NM_CONDITIONAL_TESTS:
```

tests/ensure_provider_tests.py
```diff
@@ -114,11 +119,19 @@
...
+        minimum_nm_version = NM_DEMANDING_TESTS.get(test_playbook, {}).get(
```
tyll, line 123:
```suggestion
        minimum_nm_version = NM_DEMANDING_TESTS[test_playbook].get(
```

If you check already if the key is present in the dictionary, there is no need to use `.get()`.

tests/ensure_provider_tests.py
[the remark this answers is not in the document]
tyll: It seems to me that you added more commits to this PR but did not create a new PR.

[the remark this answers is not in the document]
liangwen12year: Right. Thanks for mentioning.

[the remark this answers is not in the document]
liangwen12year: New PR, https://github.com/linux-system-roles/network/pull/386

tests/ensure_provider_tests.py
liangwen12year: Changed into `NM_CONDITIONAL_TESTS`.

tests/ensure_provider_tests.py
liangwen12year: Removed the unnecessary `.get()`.

tests/ensure_provider_tests.py
liangwen12year: Changed into `if test_playbook in NM_CONDITIONAL_TESTS:`

tests/ensure_provider_tests.py
```diff
@@ -72,10 +72,6 @@
         MINIMUM_VERSION: "'1.20.0'",
         "comment": "# NetworKmanager 1.20.0 added support for forgetting profiles",
     },
-    "playbooks/tests_ethtool_coalesce.yml": {
-        MINIMUM_VERSION: "'1.25.1'",
-        "comment": "# NetworkManager 1.25.1 introduced ethtool coalesce support",
```
tyll, line 77: Seems like this needs the new conditional tests support

liangwen12year: Will change it after that https://github.com/linux-system-roles/network/pull/386 was merged.

tyll: please just remove this commit from this PR, then.

liangwen12year: This commit was removed. I think that I may need to add the commit to #386 then.

liangwen12year:
I think the reason that CI failed is because we have not add ethtool ring test into NM conditional test category(https://github.com/linux-system-roles/network/pull/386).

library/network_connections.py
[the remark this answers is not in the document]
cathay4t: Please open a new issue or PR for tracking the coalesce efforts instead of this PR.

cathay4t (changes requested):
Please move the argument validator patches into new PR.

liangwen12year:
> Please move the argument validator patches into new PR.

New PR for argument validator patches, https://github.com/linux-system-roles/network/pull/387

library/network_connections.py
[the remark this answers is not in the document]
liangwen12year: New issue created, https://github.com/linux-system-roles/network/issues/388

jharuda:
[citest pending]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

README.md
```diff
@@ -514,6 +525,11 @@ kernel and device, changing some options might not be supported.
       tx_usecs_high: 0  # optional mininum=0 maximum=0xffffffff
       tx_usecs_irq: 0  # optional mininum=0 maximum=0xffffffff
       tx_usecs_low: 0  # optional mininum=0 maximum=0xffffffff
+    ring:
+      rx: 0  # optional mininum=0 maximum=0xffffffff
+      rx_jumbo: 0  # optional mininum=0 maximum=0xffffffff
+      rx_mini: 0  # optional mininum=0 maximum=0xffffffff
+      tx: 0  # optional mininum=0 maximum=0xffffffff
```
ffmancera, line 532: Is this correct? Are we allowing hex values? IMO, `4294967295` would be better.

ffmancera (approved):
Other than my nit, it looks good. Thanks! Please, could you include more information on the commit message. One ethtool ring example would be great, thanks!

README.md
liangwen12year: Hex values are allowed.

liangwen12year:
> Other than my nit, it looks good. Thanks! Please, could you include more information on the commit message. One ethtool ring example would be great, thanks!

Added ring example to the commit message.Thanks.

cathay4t (approved):

PR #361, opened 2021-02-25
Tests: Use more conscious language

tyll:

### DIFF
--- a/tests/playbooks/tests_bond_deprecated.yml
+++ b/tests/playbooks/tests_bond_deprecated.yml
@@ -4,9 +4,9 @@
   vars:
     controller_profile: bond0
     controller_device: nm-bond
-    slave1_profile: bond0.0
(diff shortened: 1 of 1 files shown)


ffmancera:
Hi @tyll, I agree with you but the idea of using "slave" there is to specify these tests are using "slave" words. As you can see it is the deprecated one. The others are using inclusive words. What do you think? Thanks.

tyll:
> Hi @tyll, I agree with you but the idea of using "slave" there is to specify these tests are using "slave" words. As you can see it is the deprecated one. The others are using inclusive words. What do you think? Thanks.

This is not a test to check that "slave" still works, this is the test to check that "master" still works and the change only affects test-internal variable names and a description.

ffmancera (approved):
LGTM, thanks!

PR #356, opened 2021-02-21
Enable Enhanced Open (OWE) support in network role

liangwen12year:
Signed-off-by: Wen Liang <liangwen12year@gmail.com>

PR #344, opened 2021-01-20
Enable SAE (WPA3-Personal) support in Network role

liangwen12year:
Signed-off-by: Wen Liang <liangwen12year@gmail.com>

PR #311, opened 2020-10-29
Fix initscripts parameter names for declaring static ipv6 connections

jaedolph:

PR #362, opened 2021-03-03
fix most ansible-test issues, suppress the rest

richm:
Automation Hub, and possibly Galaxy in the future, require the
collection to be screened with `ansible-test sanity` among other
checks.  The role had a number of issues:
* Use `AssertionError` instead of `assert`
* Use of `logging` module not in accordance with standards, but these
  are ok and the errors were suppressed
* Several import errors which are ok because they are checked
  elsewhere
* __init__.py in the module_utils directories must be empty, so a
  new file myerror.py was added to move the code from __init__.py
  * NOTE: network_lsr/nm/__init__.py is not empty
* The documentation block in the module was not properly constructed
  or formatted.
* shellcheck issues, including removing unused files
* use `unused` instead of `_` (underscore) for variables that are
  unused


library/network_connections.py
```diff
@@ -1911,7 +1933,12 @@ def run_prepare(self):
 
             name = connection["name"]
             if not name:
-                assert connection["persistent_state"] == "absent"
+                if not connection["persistent_state"] == "absent":
```
nhosoi, line 1936: Just from my curiosity..., is
`if not connection["persistent_state"] == "absent":`
equivlalent to
`if connection["persistent_state"] != "absent":`
?

richm: > Just from my curiosity..., is
> `if not connection["persistent_state"] == "absent":`
> equivlalent to
> `if connection["persistent_state"] != "absent":`
> ?

Yes

nhosoi (approved):
lgtm

library/network_connections.py
```diff
@@ -772,6 +777,9 @@ def connection_compare(
         if compare_flags is None:
             compare_flags = NM.SettingCompareFlags.IGNORE_TIMESTAMP
 
+        # this is really odd, but I'm assuming there is a double-negative
+        # here for a reason . . . .
+        # pylint: disable=unneeded-not
         return not (not (con_a.compare(con_b, compare_flags)))
```
tyll, line 783: @thom311 can you help here? Do you remember why we have this?

library/network_connections.py
```diff
@@ -1607,12 +1617,16 @@ def connections(self):
...
                 CheckMode.DRY_RUN,
                 CheckMode.PRE_RUN,
                 CheckMode.REAL_RUN,
-            ]
+            ]:
+                raise AssertionError(
+                    "invalid value {0} for self.check_mode".format(self.check_mode)
+                )
             c = []
+            # pylint: disable=blacklisted-name
```
tyll, line 1629:
```suggestion
            # disable check for bad names:
            # pylint: disable=C0102
```

The name for the C0102 check does not match our conscious language initiative, therefore using the code with an explanation instead of the name seems to be better here. What is the bad variable name here? `_`? Is there a better pattern to do something x times than a `for _ in range()`? Maybe we can get rid of the warning completelely.

library/network_connections.py
```diff
@@ -1988,6 +2015,7 @@ def _check_ethtool_setting_support(self, idx, connection):
                     idx, "ethtool.%s specified but not supported by NM", specified
                 )
 
+            # pylint: disable=blacklisted-name
```
tyll, line 2018: see the other comment about this change

module_utils/network_lsr/ethtool.py
```diff
@@ -46,6 +50,7 @@ def get_perm_addr(ifname):
             res = ecmd.tobytes()
         except AttributeError:  # tobytes() is not available in python2
             res = ecmd.tostring()
+        # pylint: disable=blacklisted-name
         _, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
```
tyll, line 54: see the other comment about this change

module_utils/network_lsr/nm/active_connection.py
```diff
@@ -21,6 +25,7 @@ def deactivate_active_connection(nm_ac, timeout, check_mode):
         return False
     if not check_mode:
         main_loop = client.get_mainloop(timeout)
+        # pylint: disable=logging-format-interpolation
         logging.debug(
```
tyll, line 29: Seems to me that pylint is right here and instead of disabling the check, using `logging.debug("Deactivating %s ...", nm_ac.get_id())` would be the better choice here and for the other instances.

tests/ensure_provider_tests.py
```diff
@@ -120,7 +118,7 @@ def create_nm_playbook(test_playbook):
         EXTRA_RUN_CONDITION, ""
     )
     if extra_run_condition:
-        extra_run_condition = "{}{}\n".format(
+        extra_run_condition = "{0}{1}\n".format(
```
tyll, line 121: Why is this needed? The test helper script does not need to support old python versions, does it?

tests/get_coverage.sh
```diff
@@ -32,10 +31,12 @@ call_ansible() {
 }
 
 remote_coverage_dir="$(mktemp -d /tmp/remote_coverage-XXXXXX)"
-trap "rm -rf '${remote_coverage_dir}'" EXIT
+trap 'rm -rf "${remote_coverage_dir}"' EXIT
```
tyll, line 34: Is this change correct? AFAIU, this will make `$remote_coverage_dir` to be evaluated when the script exits but it should use the value it has at the time the trap is configured.

tests/merge_coverage.sh
```diff
@@ -23,7 +23,7 @@ export COVERAGE_FILE="${1}"
 shift
 
 tempdir="$(mktemp -d /tmp/coverage_merge-XXXXXX)"
-trap "rm -rf '${tempdir}'" EXIT
```
tyll, line 26: See the other comment about this quoting change.

tyll (changes requested):
Please avoid at least introducing words that do not align with our conscious language initiative and it also seems that logging should be improved instead of ignoring the warning.

library/network_connections.py
thom311: yes, it can be removed.

richm: Does that mean I can change this to 
```
   return con_a.compare(con_b, compare_flags)
```
?

richm: Done.

library/network_connections.py
richm: I rewrote the code to avoid this issue - check it out

library/network_connections.py
richm: I rewrote the code to avoid this issue e.g. use `for option in specified.keys()`

module_utils/network_lsr/nm/active_connection.py
richm: I rewrote all of the calls to `logging` methods to use `%s`

tests/ensure_provider_tests.py
richm: I rewrote this to avoid the warning

tests/get_coverage.sh
richm: shellcheck assumes that the correct thing to do is to use single quotes around the entire `trap` command, and only expand it, including any variables used by the `trap` command, at `trap` evaulation time.  https://github.com/koalaman/shellcheck/wiki/SC2064 - but we know what we're doing here, and we want the variable expanded at the point where the `trap` is declared.  So we follow the directions:
```

Exceptions

If you don't care that the trap code is expanded early because the commands/variables won't change during execution of the script, or because you want to use the current and not the future values, then you can ignore this message.
```
so we will tell shellcheck to be quiet

module_utils/network_lsr/argument_validator.py
```diff
@@ -72,6 +76,7 @@ def connection_get_non_absent_names(connections):
 
 class ValidationError(MyError):
     def __init__(self, name, message):
+        # pylint: disable=non-parent-init-called
         Exception.__init__(self, name + ": " + message)
```
tyll, line 80: Is it possible to use `super()` here instead of ignoring the warning? https://docs.python.org/2.7/library/functions.html#super

module_utils/network_lsr/ethtool.py
```diff
@@ -46,6 +50,8 @@ def get_perm_addr(ifname):
             res = ecmd.tobytes()
         except AttributeError:  # tobytes() is not available in python2
             res = ecmd.tostring()
+        # disable check for bad names:
+        # pylint: disable=C0102
         _, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
```
tyll, line 55: Is `_` now considered to be bad? Then this would allow to avoid this AFAICS:

```suggestion
        size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)[1:]
```

but I am not 100% sure that this is better to read.

tyll (commented):
Thank you, I only have two comments.

module_utils/network_lsr/argument_validator.py
richm: Part of the problem is that the superclass is `MyError` but the code is calling `Exception.__init__` - I have never seen a constructor bypass the superclass and use an ancestor class instead - I figured this was odd enough not to try to "fix".  But I'll see if I can use `super` here.

module_utils/network_lsr/ethtool.py
richm: > Is `_` now considered to be bad?

Apparently so?  Which I don't understand - it is the standard, or at least de facto, way to ignore a return value from a function.  And there seems to be no way to tell pylint to accept it.

> Then this would allow to avoid this AFAICS:
> 
> but I am not 100% sure that this is better to read.

Another alternative
```python
size, perm_addr = struct.unpack("I%is" % MAX_ADDR_LEN, res[4:])
```

richm: There doesn't seem to be a format specifier or modifier that I can use to say "match an `I` here but don't return the value" - https://docs.python.org/2/library/struct.html

richm: @tyll so what do you want to do?
* keep the `_` with the pylint exception
* use `struct.unpack()[1:]
* use `struct.unpack(res[4:])`
* something else?

tyll: I would like to know why `_` is considered bad. It is not in the list of bad names in https://github.com/linux-system-roles/network/blob/main/pylintrc#L353 and when I run pylint locally, the C0102 warning does not appear - it does when I switch `_` to `foo`. Do you have any idea where it comes from? Does tox-lsr override this somehow?

richm: > Does tox-lsr override this somehow?

I think it must be `ansible-test` - I'm not sure why ansible would care that `_` is being used in a module.  I'll have to dig into the ansible-test source code to figure out how it is invoking pylint.

richm: https://github.com/ansible/ansible/blob/stable-2.9/test/lib/ansible_test/_data/sanity/pylint/config/ansible-test.cfg#L28
I've asked the ansible team why - waiting on a reply

tyll: In their custom check script they wrote to use `dummy` instead of `_`: [PR](https://github.com/ansible/ansible/pull/54140/commits/7928dba817271802077b95881b33209bfb2ffb69#diff-b91efbf85963e341f0494813aa21ca602b60097cf080310de1f707acebfef2bb), see test/sanity/code-smell/no-underscore-variable.py

richm: Ansible devs say "its also 'standard for gettext'" https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-underscore-variable.html - Ansible reserves the right to use `_` for its own future use

Ansible devs suggest to use `dummy` - so
```
+        dummy, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
```
this passes with `tox -e pylint`
I also checked if using the word "dummy" for a variable name is problematic with any conscious language initiatives - I haven't found any

@tyll ok to use `dummy` here?

tyll: @richm dummy is fine by me, thanks

module_utils/network_lsr/argument_validator.py
richm: `super` works - see https://github.com/linux-system-roles/network/pull/362/commits/e70dd1b04722fc2f617bdb2745f54b6312f56c18

module_utils/network_lsr/ethtool.py
richm: fixed in https://github.com/linux-system-roles/network/pull/362/commits/e70dd1b04722fc2f617bdb2745f54b6312f56c18

richm:
Rebase report - what effect will merging this PR have on other PRs:
* #363 - rebases cleanly
* #356 - needs rebase to latest `main` branch - then minor changes
* #355 - needs rebase to latest `main` branch - then minor changes
* #344 - needs rebase to latest `main` branch - then minor changes
* #311 - needs rebase to latest `main` branch
* #309 - needs rebase to latest `main` branch

richm:
so wdyt?  Can we merge this PR, and I help PR authors rebase and fixup their PRs?

richm:
ok to merge?

library/network_connections.py
```diff
@@ -1408,7 +1416,8 @@ def check_mode_set(self, check_mode, connections=None):
             or (c == CheckMode.PRE_RUN and check_mode in [CheckMode.REAL_RUN])
             or (c == CheckMode.REAL_RUN and check_mode in [CheckMode.DONE])
             or (c == CheckMode.DRY_RUN and check_mode in [CheckMode.DONE])
-        )
+        ):
+            raise AssertionError("check_mode value is incorrect {0}".format(c))
```
liangwen12year, line 1420: should this be `raise AssertionError("updating check_mode value from {0} into {1} is incorrect".format(c, check_mode))` ?

richm: fixed https://github.com/linux-system-roles/network/pull/362/commits/2bfe053a5a4011a8002cd5ec43e62b7a8489ed80#diff-6b3f5f6149ac1330c5e2c832286b63b3597158a998f2244ddbec2045ace1f794R1420

module_utils/network_lsr/nm_provider.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: BSD-3-Clause
 """ Support for NetworkManager aka the NM provider """
 
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
```
liangwen12year, line 7: Is 'type' the default 'metaclass' in python ?

richm: > Is 'type' the default 'metaclass' in python ?

https://docs.ansible.com/ansible/2.9/dev_guide/testing/sanity/metaclass-boilerplate.html

liangwen12year (approved):
LGTM!

ffmancera:
@richm There are 2 black formatting errors. Other than that, it looks good to me. :-)

richm:
> @richm There are 2 black formatting errors. Other than that, it looks good to me. :-)

fixed

richm:
@tyll have all of your requested changes been addressed?

tyll:
> @tyll have all of your requested changes been addressed?

yes, thank you.

module_utils/network_lsr/ethtool.py
```diff
@@ -46,7 +50,7 @@ def get_perm_addr(ifname):
             res = ecmd.tobytes()
         except AttributeError:  # tobytes() is not available in python2
             res = ecmd.tostring()
-        _, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
+        dummy, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
```
tyll, line 53: The inclusive open source community orientation training at

https://training.linuxfoundation.org/training/inclusive-open-source-community-orientation-lfc102/


recommends in

https://trainingportal.linuxfoundation.org/learn/course/inclusive-open-source-community-orientation-lfc102/making-change/making-change

to avoid the term "Dummy Value" and use "Placeholder Value" instead. I do not see an explanation for this.

richm: I guess I have to register to see the contents of that link?  At any rate - what is the suggestion?
```python
         placeholder, size, perm_addr = struct.unpack("II%is" % MAX_ADDR_LEN, res)
```
?

thom311: why do you want to replace `_` in the first place? I though that is pretty common and pythonic.

An alternative name might be `unused`. While I am surprised to hear "dummy" is discouraged, `unused` anyway seems a better name.

richm: > why do you want to replace `_` in the first place? I though that is pretty common and pythonic.

https://github.com/linux-system-roles/network/pull/362#discussion_r589668462
tl;dr - `_` is also used for i18n, and Ansible wants to reserve `_` for future use.

> 
> An alternative name might be `unused`. While I am surprised to hear "dummy" is discouraged, `unused` anyway seems a better name.

"dummy" is a derogatory word which can be seen as non-inclusive

`unused` works for me

richm: I'll note that "dummy" is used in many places in the network role:
```
> git grep -c dummy
examples/dummy_simple.yml:4
library/network_connections.py:1
module_utils/network_lsr/argument_validator.py:1
pylintrc:2
tests/ensure_provider_tests.py:1
tests/playbooks/tests_dummy.yml:5
tests/tasks/create_dummy_profile.yml:1
tests/tasks/manage_test_interface.yml:9
tests/tests_dummy_nm.yml:2
tests/tests_helpers_and_asserts.yml:2
```

tyll: > I guess I have to register to see the contents of that link? At any rate - what is the suggestion?

You can easily login with your Red Hat Google account if you like. The course was recommended in the past and you can get a Fedora badge if you complete it.

> I'll note that "dummy" is used in many places in the network role:

It is also an official network interface type, see https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking/#dummy so we cannot get it completely removed.

> 
> ```
> > git grep -c dummy
> examples/dummy_simple.yml:4
> library/network_connections.py:1
> module_utils/network_lsr/argument_validator.py:1
> pylintrc:2

This is a good find since it shwos that `unused` is configured as an expectedly not used variable name, so we can use that.

> tests/ensure_provider_tests.py:1
> tests/playbooks/tests_dummy.yml:5
> tests/tasks/create_dummy_profile.yml:1
> tests/tasks/manage_test_interface.yml:9
> tests/tests_dummy_nm.yml:2
> tests/tests_helpers_and_asserts.yml:2
> ```

richm:
[citest]

jharuda:
[citest]

richm:
[citest]

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
+tests/network/ensure_provider_tests.py compile-2.7!skip
```
tyll, line 1: It would be nice to sort this list to ensure there are no duplicates or contradictions.

richm: done

tyll:
Not sure if this is related to other changes, but there are now CI errors (Tox Python 3.6):

https://github.com/linux-system-roles/network/pull/362/checks?check_run_id=2290367054
```text
Running sanity test 'import' with Python 2.6
541
ERROR: Found 4 import issue(s) on python 2.6 which need to be resolved:
542
ERROR: plugins/module_utils/network_lsr/nm/active_connection.py:13:0: traceback: ImportError: No module named gi (at plugins/module_utils/network_lsr/nm/client.py:13:0)
543
ERROR: plugins/module_utils/network_lsr/nm/client.py:13:0: traceback: ImportError: No module named gi
544
ERROR: plugins/module_utils/network_lsr/nm/connection.py:13:0: traceback: ImportError: No module named gi (at plugins/module_utils/network_lsr/nm/client.py:13:0)
545
ERROR: plugins/module_utils/network_lsr/nm/provider.py:11:0: traceback: ImportError: No module named gi (at plugins/module_utils/network_lsr/nm/client.py:13:0)
```

gi does not exist on Python 2.6 and is also not needed there. This is a module needed to access the NetworkManager API. Not sure where this is coming from, seems to be related to the conversion to collections.

Also, there are some warnings, not sure what causes them, maybe it is related to broken symlinks?
https://github.com/linux-system-roles/network/pull/362/checks?check_run_id=2290367054#step:5:257

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
...
+plugins/module_utils/network_lsr/nm/connection.py import-3.5!skip
+plugins/module_utils/network_lsr/nm/connection.py import-3.6!skip
+plugins/module_utils/network_lsr/nm/connection.py import-3.7!skip
+plugins/module_utils/network_lsr/nm/connection.py import-3.8!skip
+plugins/module_utils/network_lsr/nm/provider.py import-2.7!skip
+plugins/module_utils/network_lsr/nm/provider.py import-3.5!skip
+plugins/module_utils/network_lsr/nm/provider.py import-3.6!skip
+plugins/module_utils/network_lsr/nm/provider.py import-3.7!skip
+plugins/module_utils/network_lsr/nm/provider.py import-3.8!skip
+plugins/module_utils/network_lsr/utils.py shebang!skip
```
tyll, line 25: Shebang can be removed.

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
...
+plugins/modules/network_connections.py validate-modules:missing-examples
+plugins/modules/network_connections.py validate-modules:missing-gplv3-license
+plugins/modules/network_connections.py validate-modules:no-default-for-required-parameter
+plugins/modules/network_connections.py validate-modules:parameter-type-not-in-doc
+plugins/modules/network_connections.py validate-modules:undocumented-parameter
+tests/network/covstats shebang!skip
```
tyll, line 31: Seems this just needs the whitespace to be removed from the script to fix this.

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
...
+tests/network/ensure_provider_tests.py future-import-boilerplate!skip
+tests/network/ensure_provider_tests.py metaclass-boilerplate!skip
+tests/network/ensure_provider_tests.py shebang!skip
+tests/network/get_coverage.sh shebang!skip
+tests/network/get_total_coverage.sh shebang!skip
+tests/network/integration/conftest.py future-import-boilerplate!skip
+tests/network/integration/conftest.py metaclass-boilerplate!skip
+tests/network/integration/test_ethernet.py future-import-boilerplate!skip
+tests/network/integration/test_ethernet.py metaclass-boilerplate!skip
+tests/network/merge_coverage.sh shebang!skip
```
tyll, line 43: Seems this just needs the whitespace to be removed from the script to fix this.

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
+plugins/module_utils/network_lsr/__init__.py shebang!skip
+plugins/module_utils/network_lsr/argument_validator.py shebang!skip
+plugins/module_utils/network_lsr/myerror.py shebang!skip
```
tyll, line 3: The shebangs can be removed from these files and then they don't need to be skipped.

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
...
+plugins/module_utils/network_lsr/nm/__init__.py empty-init!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-2.7!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.5!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.6!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.7!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.8!skip
```
tyll, line 9: Are these the imports from ansible.module_utils? Are they needed for 2.6, too?

.sanity-ansible-ignore-2.9.txt
```diff
@@ -0,0 +1,47 @@
...
+plugins/modules/network_connections.py validate-modules:parameter-type-not-in-doc
+plugins/modules/network_connections.py validate-modules:undocumented-parameter
+tests/network/covstats shebang!skip
+tests/network/ensure_provider_tests.py compile-2.7!skip
+tests/network/ensure_provider_tests.py compile-3.5!skip
+tests/network/ensure_provider_tests.py future-import-boilerplate!skip
+tests/network/ensure_provider_tests.py metaclass-boilerplate!skip
+tests/network/ensure_provider_tests.py shebang!skip
+tests/network/get_coverage.sh shebang!skip
+tests/network/get_total_coverage.sh shebang!skip
```
tyll, line 38: Seems these just needs the whitespace to be removed from the script to fix this.

tyll (commented):
Most shebang exceptions could be cleaned up.

tyll:
The commit message is not accurate anymore since a few things were done differently, now. If you like, you can move some of the changes like the python code in a separate PR that we can merge to reduce the amount of changes piling up in this PR.

.sanity-ansible-ignore-2.9.txt
richm: apparently not, if you are talking about the `shebang`

.sanity-ansible-ignore-2.9.txt
richm: apparently not . . .

.sanity-ansible-ignore-2.9.txt
richm: apparently not . . .

.sanity-ansible-ignore-2.9.txt
tyll: This exception is about what is described in https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/shebang.html, isn't it? It states `#!/bin/bash` as a valid Shebang but the file uses `#! /bin/bash`. Not sure what I am missing here.

tyll: `#! /bin/bash` is valid for the Linux kernel, though.

richm: This is the error, for example:
```
ERROR: tests/network/covstats:1:1: unexpected non-module shebang: b'#!/bin/bash'
See documentation for help: https://docs.ansible.com/ansible/2.9/dev_guide/testing/sanity/shebang.html
```
I think the problem is that ansible-test doesn't understand executable files outside of the expected module/plugin directories.  So it thinks that `tests/covstats` is a module since it is executable and has a shebang, but it isn't in the right directory.   This is because we need to have these helper scripts but there isn't a good place to keep them.  There isn't a general Ansible role/collection convention for helper scripts used during build/test time.  The ssh role uses `.dev-tools` for this - https://github.com/linux-system-roles/ssh/tree/master/.dev-tools - I think a dotfile directory is a good idea - we used to use `.travis/` for these sorts of scripts.

richm:
> The commit message is not accurate anymore since a few things were done differently, now. If you like, you can move some of the changes like the python code in a separate PR that we can merge to reduce the amount of changes piling up in this PR.

Based on the recent changes, including updates to the PR description and commit message, do we still need to split this?

.sanity-ansible-ignore-2.9.txt
tyll: The scripts are currently meant for a human to be used so hiding them in a dotdir seems to be wrong. I am fine with skipping them for now.

tyll:
> > The commit message is not accurate anymore since a few things were done differently, now. If you like, you can move some of the changes like the python code in a separate PR that we can merge to reduce the amount of changes piling up in this PR.
> 
> Based on the recent changes, including updates to the PR description and commit message, do we still need to split this?

I assumed the CI failures would cause more pain, I am fine with merging this as-is once the CI is finished.

tyll (approved):
:+1:

tyll:
[citest bad]

PR #363, opened 2021-03-07
Specify PCI address to configure profile

liangwen12year:
Proposed schema:

```
match:
    driver: ...
    name: ... 
    kernel_argument: ...
    path: ... (this will be implemented here)
    mac: ...
```

PR #364, opened 2021-03-17
Support specifying mac address without quotation mark

liangwen12year:

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -14,6 +14,9 @@ from ansible.module_utils.network_lsr.myerror import MyError  # noqa:E501
 from ansible.module_utils.network_lsr.utils import Util  # noqa:E501
 
 UINT32_MAX = 0xFFFFFFFF
+# a MAC address for type ethernet requires 6 octets
+# a MAC address for type infiniband requires 20 octets
+MAC_ADDR_OCTETS = [6, 20]
 
 
 class ArgUtil:
@@ -416,11 +419,25 @@ class ArgValidatorIP(ArgValidatorStr):
 
 
 class ArgValidatorMac(ArgValidatorStr):
-    def __init__(self, name, force_len=None, required=False, default_value=None):
+    def __init__(
+        self, name, force_len=MAC_ADDR_OCTETS, required=False, default_value=None
+    ):
         ArgValidatorStr.__init__(self, name, required, default_value, None)
         self.force_len = force_len
 
     def _validate_impl(self, value, name):
+        if type(value) == int:
+            try:
+                value = Util.num_to_mac(value, self.force_len)
+            except MyError:
+                raise ValidationError(
+                    name, "value '%s' is not a valid MAC address" % (value)
+                )
+        if not isinstance(value, Util.STRING_TYPE):
+            raise ValidationError(
+                name,
+                "must be a string and value should be quoted, but is '%s', " % (value),
+            )
         v = ArgValidatorStr._validate_impl(self, value, name)
         try:
             addr = Util.mac_aton(v, self.force_len)
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -3361,6 +3361,41 @@ class TestValidator(unittest.TestCase):
             self.assertTrue("port_type" in connection)
             self.assertTrue("slave_type" not in connection)
 
+    def test_mac_address_argvalidator(self):
+        """
+        Test that argvalidator for validating mac address is correctly defined.
+        """
+        validator = network_lsr.argument_validator.ArgValidatorMac("mac")
+        self.assertEqual(validator.validate(41135085296), "52:54:00:12:34:56")
+        self.assertEqual(validator.validate("00:00:5e:00:53:5d"), "00:00:5e:00:53:5d")
+        self.assertEqual(validator.validate("00:00:00:00:00:00"), "00:00:00:00:00:00")
+        self.assertEqual(validator.validate("ff:ff:ff:ff:ff:ff"), "ff:ff:ff:ff:ff:ff")
+        self.assertEqual(
+            validator.validate(
+                "80:00:00:6d:fe:80:00:00:00:00:00:00:00:02:55:00:70:33:cf:01"
+            ),
+            "80:00:00:6d:fe:80:00:00:00:00:00:00:00:02:55:00:70:33:cf:01",
+        )
+        self.assertEqual(
+            validator.validate(
+                "01:02:03:04:05:06:07:08:09:0A:01:02:03:04:05:06:07:08:09:11"
+            ),
+            "01:02:03:04:05:06:07:08:09:0a:01:02:03:04:05:06:07:08:09:11",
+        )
+        self.assertValidationError(validator, 1234)
+        self.assertValidationError(validator, "aa:bb")
+        self.assertValidationError(validator, sys.maxsize)
+        self.assertValidationError(validator, 0)
+        self.assertValidationError(
+            validator, "80:00:00:6d:fe:80:00:00:00:00:00:00:00:02:55:00:70:33:cf:"
+        )
+        self.assertValidationError(
+            validator, "80:00:00:6d:fe:80:00:00:00:00:00:00:00:02:55:00:70:33:cf:xx"
+        )
+        self.assertValidationError(
+            validator, "80:00:00:6d:fe:80:00:00:00:00:00:00:00:02:55:000:70:33:cf:01"
+        )
+
 
 @my_test_skipIf(nmutil is None, "no support for NM (libnm via pygobject)")
 class TestNM(unittest.TestCase):
--- a/module_utils/network_lsr/utils.py
+++ b/module_utils/network_lsr/utils.py
@@ -255,7 +255,7 @@ class Util:
         if i == 1:
             raise MyError("not a valid MAC address: '%s'" % (mac_str))
         if force_len is not None:
-            if force_len != len(b):
+            if not any(a for a in force_len if a == len(b)):
                 raise MyError(
                     "not a valid MAC address of length %s: '%s'" % (force_len, mac_str)
                 )
@@ -271,6 +271,24 @@ class Util:
     @staticmethod
     def mac_norm(mac_str, force_len=None):
         return Util.mac_ntoa(Util.mac_aton(mac_str, force_len))
+
+    @staticmethod
+    def num_to_mac(num, force_len=None):
+        mac = []
+        s_num = num
+        while num:
+            num, mod = divmod(num, 60)
+            if mod < 10:
+                mac.insert(0, "0" + str(mod))
+            else:
+                mac.insert(0, str(mod))
+        mac_addr = ":".join(mac)
+        if force_len is not None:
+            if not any(a for a in force_len if a == len(mac)):
+                raise MyError(
+                    "not a valid number '%d' for MAC address: '%s'" % (s_num, mac_addr)
+                )
+        return mac_addr
 
     @staticmethod
     def boolean(arg):


thom311:
can you show an example how this is used?

What prompted this merge-request, is there an issue that you can link to?

```
   validator._validate_impl(41135085296, "mac"), "52:54:00:12:34:56"
```

Why does `41135085296` correspond to `"52:54:00:12:34:56"`, or better, where did you get this (to me unknown) algorithm with `divmod(num, 60)`? That doesn't make sense to me.

liangwen12year:
> can you show an example how this is used?
> 
> What prompted this merge-request, is there an issue that you can link to?
> 
> ```
>    validator._validate_impl(41135085296, "mac"), "52:54:00:12:34:56"
> ```
> 
> Why does `41135085296` correspond to `"52:54:00:12:34:56"`, or better, where did you get this (to me unknown) algorithm with `divmod(num, 60)`? That doesn't make sense to me.

The issue:
https://github.com/linux-system-roles/network/issues/365

module_utils/network_lsr/argument_validator.py
```diff
@@ -418,6 +418,8 @@ def __init__(self, name, force_len=None, required=False, default_value=None):
         self.force_len = force_len
 
     def _validate_impl(self, value, name):
+        if type(value) == int:
+            value = Util.num_to_mac(value, self.force_len)
```
liangwen12year, line 422: if we replace `self.force_len` into 6 here:
the following test cases failed:
```
validator._validate_impl(1234, "mac")
ansible.module_utils.network_lsr.MyError: not a valid sexagesimal number '1234' for MAC address: '20:34'
validator._validate_impl(sys.maxsize, "mac")
ansible.module_utils.network_lsr.MyError: not a valid sexagesimal number '9223372036854775807' for MAC address: '15:15:13:34:32:31:55:20:15:30:07'
```

thom311:
> The issue:
> #365

Btw, would have been good to link the background information upfront.


That doesn't seem a good idea. It allows nonsense like:

```
   mac: 1
```
which IMO should be rejected as error.

Also, if you read https://stackoverflow.com/a/52732369/354393 you'll see that number larger or equal than 60 are treated as carrier to the next digit. So, I would guess that "00:00:00:00:60" is the same as "00:00:00:01:00", which seems very wrong.

thom311:
> Also, if you read https://stackoverflow.com/a/52732369/354393 you'll see that number larger or equal than 60 are treated as carrier to the next digit. So, I would guess that "00:00:00:00:60" is the same as "00:00:00:01:00", which seems very wrong.

Ah wrong, it doesn't carry over. The link says "Since your other MAC-address, 84:19:14:15:86:58, has a number after the first number that is over 59 (86 that is), it is not considered as a sexagesimal number and is therefore treated literally as a string.".

But that seems not something that the role should try to work around. The user has to set the MAC address as string and not the role interpreting integers as hours:minutes:seconds...

liangwen12year:
> > The issue:
> > #365
> 
> Btw, would have been good to link the background information upfront.
> 
> That doesn't seem a good idea. It allows nonsense like:
> 
> ```
>    mac: 1
> ```
> 
> which IMO should be rejected as error.
> 
> Also, if you read https://stackoverflow.com/a/52732369/354393 you'll see that number larger or equal than 60 are treated as carrier to the next digit. So, I would guess that "00:00:00:00:60" is the same as "00:00:00:01:00", which seems very wrong.

`mac: 1` was indeed rejected as error by the function `num_to_mac()`

richm:
Are you absolutely guaranteed that you can always convert a mac address specified like `mac: 52:54:00:12:34:56` into a unique number, and that number can always be converted back into the same exact mac address specified in the original YAML?
If you are not absolutely guaranteed, then you will just have to require users to specify the mac address as a string in any of the various ways to specify a string in YAML.

liangwen12year:
> > Also, if you read https://stackoverflow.com/a/52732369/354393 you'll see that number larger or equal than 60 are treated as carrier to the next digit. So, I would guess that "00:00:00:00:60" is the same as "00:00:00:01:00", which seems very wrong.
> 
> Ah wrong, it doesn't carry over. The link says "Since your other MAC-address, 84:19:14:15:86:58, has a number after the first number that is over 59 (86 that is), it is not considered as a sexagesimal number and is therefore treated literally as a string.".
> 
> But that seems not something that the role should try to work around. The user has to set the MAC address as string and not the role interpreting integers as hours:minutes:seconds...

if the mac is 

> Are you absolutely guaranteed that you can always convert a mac address specified like `mac: 52:54:00:12:34:56` into a unique number, and that number can always be converted back into the same exact mac address specified in the original YAML?
> If you are not absolutely guaranteed, then you will just have to require users to specify the mac address as a string in any of the various ways to specify a string in YAML.

the function `num_to_mac()` literally is to remediate yaml's peculiarity that "string literal consists of a series of one or two-digit numbers delimited by colons and all numbers but the first are between 0 and 59, it would be interpreted as a sexagesimal number, and would be converted automatically to the equivalent number of seconds".  I am guaranteed that the function `num_to_mac()` will always convert it back into mac address  specified in the original YAML. Also, I understand the concern from @thom311,  like specifying arbitrary integer( e.g. `mac: 1`), but we also have `force_len` to raise error for invalid interger value

liangwen12year:
> > Also, if you read https://stackoverflow.com/a/52732369/354393 you'll see that number larger or equal than 60 are treated as carrier to the next digit. So, I would guess that "00:00:00:00:60" is the same as "00:00:00:01:00", which seems very wrong.
> 
> Ah wrong, it doesn't carry over. The link says "Since your other MAC-address, 84:19:14:15:86:58, has a number after the first number that is over 59 (86 that is), it is not considered as a sexagesimal number and is therefore treated literally as a string.".
> 
> But that seems not something that the role should try to work around. The user has to set the MAC address as string and not the role interpreting integers as hours:minutes:seconds...

if the mac is 

> Are you absolutely guaranteed that you can always convert a mac address specified like `mac: 52:54:00:12:34:56` into a unique number, and that number can always be converted back into the same exact mac address specified in the original YAML?
> If you are not absolutely guaranteed, then you will just have to require users to specify the mac address as a string in any of the various ways to specify a string in YAML.

the function `num_to_mac()` literally is to remediate yaml's peculiarity that "string literal consists of a series of one or two-digit numbers delimited by colons and all numbers but the first are between 0 and 59, it would be interpreted as a sexagesimal number, and would be converted automatically to the equivalent number of seconds".  I am guaranteed that the function `num_to_mac()` will always convert it back into mac address  specified in the original YAML. Also, I understand the concern from @thom311,  like specifying arbitrary integer( e.g. `mac: 1`), but we also have `force_len` to raise error for invalid interger value

liangwen12year:
> Are you absolutely guaranteed that you can always convert a mac address specified like `mac: 52:54:00:12:34:56` into a unique number, and that number can always be converted back into the same exact mac address specified in the original YAML?
> If you are not absolutely guaranteed, then you will just have to require users to specify the mac address as a string in any of the various ways to specify a string in YAML.

And it is very good suggestion that always remind users to specify the mac address as a string in any of the various ways to specify a string in YAML.

README.md
```diff
@@ -266,7 +266,7 @@ By default, profiles are created with autoconnect enabled.
 The `mac` address is optional and restricts the profile to be usable only on
 devices with the given MAC address. `mac` is only allowed for `type`
 `ethernet` or `infiniband` to match a non-virtual device with the
-profile.
+profile. The value of `mac` address is string and should use with quotation mark.
```
tyll, line 269: Please put the documentation update in its own PR, then we can merge it by itself.

module_utils/network_lsr/argument_validator.py
```diff
@@ -196,7 +196,14 @@ def __init__(  # pylint: disable=too-many-arguments
 
     def _validate_impl(self, value, name):
         if not isinstance(value, Util.STRING_TYPE):
-            raise ValidationError(name, "must be a string but is '%s'" % (value))
+            if name[-3:] == "mac":
```
tyll, line 199: This belongs in an ArgValidatorMac class.

tests/unit/test_network_connections.py
```diff
@@ -3260,6 +3260,15 @@ def test_set_deprecated_slave_type(self):
             self.assertTrue("port_type" in connection)
             self.assertTrue("slave_type" not in connection)
 
+    def test_mac_address_argvalidator(self):
+        """
+        Test that argvalidator for validating mac address is correctly defined.
+        """
+        validator = network_lsr.argument_validator.ArgValidatorMac("mac")
+        self.assertEqual(
+            validator._validate_impl(41135085296, "mac"), "52:54:00:12:34:56"
```
tyll, line 3269: This should probably call validator.validate() instead of the internal method `_validate_impl()`.

tests/unit/test_network_connections.py
```diff
@@ -3260,6 +3260,15 @@ def test_set_deprecated_slave_type(self):
             self.assertTrue("port_type" in connection)
             self.assertTrue("slave_type" not in connection)
 
+    def test_mac_address_argvalidator(self):
```
tyll, line 3263: This should probably be updated to test for ethernet or infiniband or both.

README.md
liangwen12year: Seperate PR for docs,
https://github.com/linux-system-roles/network/pull/367

tests/unit/test_network_connections.py
liangwen12year: We can not know the connection type inside `ArgValidatorMac`, it may not be  achievable to have the unit tests for ethernet or infiniband respectively(based on connection type),

But we still have the sanity check which based on connection type (ethernet or infiniband)  
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1484

module_utils/network_lsr/argument_validator.py
```diff
@@ -413,11 +413,23 @@ def _validate_impl(self, value, name):
 
 
 class ArgValidatorMac(ArgValidatorStr):
-    def __init__(self, name, force_len=None, required=False, default_value=None):
+    def __init__(self, name, force_len=[6, 20], required=False, default_value=None):
```
cathay4t, line 416: Instead of using magic number. Please use constant and make a comment for the constant why 6 and 20 matters.

cathay4t (changes requested):
Besides the magic number, this PR is breaking the API!

Previously, the unquoted mac address works well when mac address contains [a-f], but your patch change it to raise exception.

Your title says you are supporting unquoted mac address, but the code looks like you are raise exception for it.

module_utils/network_lsr/argument_validator.py
thom311: or just don't use this force-len argument. The correct length depends on the layer 2 type, and `ArgValidatorMac` does not know whether it's used in an ethernet or infiniband profile.

The length was already validated at a later pointer, where the knowledge about the layer2 type is present and a good message can be raised (like "and ethernet MAC address must be 6 octets long").

tests/unit/test_network_connections.py
liangwen12year: I called `validator.validate()` instead

module_utils/network_lsr/argument_validator.py
liangwen12year: I validated in the ArgValidatorMac class as following:
```
        if type(value) == int:
            try:
                value = Util.num_to_mac(value, self.force_len)
            except MyError:
                raise ValidationError(
                    name, "value '%s' is not a valid MAC address" % (value)
                )
        if not isinstance(value, Util.STRING_TYPE):
            raise ValidationError(
                name,
                "must be a string and value should be quoted, but is '%s', " % (value),
            )
```

liangwen12year:
> Besides the magic number, this PR is breaking the API!
> 
> Previously, the unquoted mac address works well when mac address contains [a-f], but your patch change it to raise exception.
> 
> Your title says you are supporting unquoted mac address, but the code looks like you are raise exception for it.

For  the unquoted mac address containing [a-f], it will not raise an exception, it will work as before

module_utils/network_lsr/argument_validator.py
liangwen12year: I used a constant `MAC_ADDR_OCTETS = [6, 20]` instead. See
https://github.com/linux-system-roles/network/blob/e98e8d1bc64259f711b338771ed12479b7eb091a/module_utils/network_lsr/argument_validator.py#L14

ffmancera:
I do not like the magic numbering and do not see the problem on specifying the mac address as string. But, I am curious, why is this needed? And why not in other projects like nmstate? @cathay4t @tyll I assume it is because this is missing some schema..

cathay4t:
@liangwen12year This PR is still in `DRAFT` state. And the git commit has no commit comment explaining why patch is required and how this patch is fixing the issue.

module_utils/network_lsr/argument_validator.py
```diff
@@ -11,6 +11,9 @@
 from ansible.module_utils.network_lsr.utils import Util  # noqa:E501
 
 UINT32_MAX = 0xFFFFFFFF
+# a MAC address for type ethernet requires 6 octets
+# a MAC address for type infiniband requires 20 octets
+MAC_ADDR_OCTETS = [6, 20]
```
cathay4t, line 16: This is wrong.

The kernel has no limitation minimum MAC address length.
And the maximum is not 20, but in `/usr/include/linux/netdevice.h`:

#define MAX_ADDR_LEN	32		/* Largest hardware address length */

That is 32 octets and 64 hex string characters.

cathay4t (changes requested):

module_utils/network_lsr/argument_validator.py
liangwen12year: The reason that I define `MAC_ADDR_OCTETS = [6, 20]` is because I saw the validation check here,
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1483

And in NM,it was also clear indicated,
```
     * ---ifcfg-rh---
     * property: mac-address
     * variable: HWADDR
     * description: IBoIP 20-byte hardware address of the device (in traditional
     *    hex-digits-and-colons notation).
     *    Note that for initscripts this is the current MAC address of the device as found
     *    during ifup. For NetworkManager this is the permanent MAC address. Or in case no
     *    permanent MAC address exists, the MAC address initially configured on the device.
     * example: HWADDR=01:02:03:04:05:06:07:08:09:0A:01:02:03:04:05:06:07:08:09:11
     * ---end---
     
     /* ---keyfile---
     * property: mac-address
     * format: usual hex-digits-and-colons notation
     * description: MAC address in traditional hex-digits-and-colons notation
     *   (e.g. 00:22:68:12:79:A2), or semicolon separated list of 6 bytes (obsolete)
     *   (e.g. 0;34;104;18;121;162)
     * ---end---
```

liangwen12year:
> @liangwen12year This PR is still in `DRAFT` state. And the git commit has no commit comment explaining why patch is required and how this patch is fixing the issue.

Added commit message to explain why patch is required and how this patch is fixing the issue.

jharuda:
[citest pending]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

cathay4t:
@liangwen12year This PR is still in `DRAFT` mode. Is it ready for review?

liangwen12year:
> @liangwen12year This PR is still in `DRAFT` mode. Is it ready for review?

It is still in `DRAFT` mode, because we need to discuss if this patch is actually needed.

richm:
[citest pending]

richm:
any update on this?

tyll:
This is currently not a priority, therefore it is just a draft PR. We can revisit this after #363 are handled and the current CI problems are handled. Would you prefer to have the PR closed for now?

richm:
> This is currently not a priority, therefore it is just a draft PR. We can revisit this after #363 are handled and the current CI problems are handled. Would you prefer to have the PR closed for now?

No, this is fine - it is marked as WIP/Draft

PR #367, opened 2021-03-23
Docs: update 'mac' to suggest user using quotation mark

liangwen12year:

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,11 @@ By default, profiles are created with autoconnect enabled.
 The `mac` address is optional and restricts the profile to be usable only on
 devices with the given MAC address. `mac` is only allowed for `type`
 `ethernet` or `infiniband` to match a non-virtual device with the
-profile.
+profile. The value of the `mac` address needs to be specified in hexadecimal notation
+using colons (for example: `mac: "00:00:5e:00:53:5d"`). To avoid YAML parsing mac
+addresses as integers in sexagesimal (base 60) notation (see
+<https://yaml.org/spec/1.1/#id858600>), it is recommended to always quote the value
+with double quotes and sometimes it is necessary.
 
 - For `NetworkManager`, `mac` is the permanent MAC address, `ethernet.mac-address`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -266,7 +266,7 @@ By default, profiles are created with autoconnect enabled.
 The `mac` address is optional and restricts the profile to be usable only on
 devices with the given MAC address. `mac` is only allowed for `type`
 `ethernet` or `infiniband` to match a non-virtual device with the
-profile.
+profile. The value of `mac` address is string and should use with quotation mark.
```
tyll, line 269: Please add more context for the reader to understand the problem (that yaml might parse it as a float in sexagesimal notation) and/or link to a good external resource that explains this - there might be something on the Ansible website or the yaml spec.

thom311: also update `ArgValidatorMac` to reject values of type integer!

thom311: I guess, that already happens... so just make sure that the error message is sensible.

liangwen12year: updated

README.md
```diff
@@ -266,7 +266,9 @@ By default, profiles are created with autoconnect enabled.
...
+Because `mac` address value is separated by colon, if not quoted, YAML may parse it as
+a integer in sexagesimal(base 60) form. (For more details, check https://yaml.org/spec/1.1/#id858600)
```
tyll, line 271: Thank you. There are a few grammar errors in the string and I realized that the description does not really tell that the address needs to use colons (sometimes hyphens are used, too). Here is a suggestion:

```suggestion
profile. The value of the `mac` address needs to be specified in hexadecimal notation using colons (for example: `mac: "00:00:5e:00:53:5d"`). To avoid YAML parsing mac addresses as integers in sexagesimal (base 60) notation (see https://yaml.org/spec/1.1/#id858600), it is recommended to always quote the value with double quotes and sometimes it is necessary.
```

jharuda:
[citest bad]

jharuda:
[citest bad]

README.md
liangwen12year: updated

tyll:
LGTM

cathay4t (approved):

liangwen12year:
@cathay4t ,rebased!

PR #368, opened 2021-03-24
CHANGELOG: update changelog for 1.3.0 release

liangwen12year:

### DIFF
--- a/.mdl_style.rb
+++ b/.mdl_style.rb
@@ -5,3 +5,5 @@ rule 'MD003', :style => :setext_with_atx
 rule 'MD013', :line_length => 88
 # https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix
 rule 'MD029', :style => :ordered
+# https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md024---multiple-headers-with-the-same-content
+rule "MD024", :allow_different_nesting => true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,26 @@
 Changelog
 =========
 
+[1.3.0] - 2021-04-08
+--------------------
+
+### Changes
+
+- Use inclusive language
+  - `slave` is deprecated in favor of `port`
+  - `master` is deprecated in favor of `controller`
+
+### New features
+
+- Support disabling IPv6
+- Support `dns_options` when using one or more IPv4 nameservers
+- Support Ethtool coalesce settings
+- Support dummy interfaces
+
+### Bug fixes
+
+- Fix static IPv6 support for initscripts provider
+
 [1.2.0] - 2020-08-26
 --------------------
 


CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
 # Changelog
+## [1.2.1] - 2021-03-24
+### Changes
+- Use inclusive language
+Use inclusive language in docs, variable names, and more
```
tyll, line 5: This needs more explanation.

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
 # Changelog
+## [1.2.1] - 2021-03-24
```
tyll, line 2: This should probably be 1.3.0 since there are new features and bigger changes for inclusive language

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
 # Changelog
+## [1.3.0] - 2021-03-24
+### Changes
+- Use inclusive language
+Use inclusive language in docs, variable names, and more
+
+### New features
+- Support ipv6.method disabled (8318686)
+- Support dns-options (880b7ab)
```
tyll, line 9: Please explain which options and try to use more formal language instead of technical terms - also the actual setting would be with underscore and not with dashes. To me the commit IDs seem to be disturbing.

- Support to configure DNS options to ... and ...

Also noticed that the README does not explain which DNS options can be set and what they do and how they are written. Are these the same options that resolv.conf(5) describes? Then this could be written. If there are differences they should be explained.

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+- Add support for ethtool coalesce settings (b586d11)
```
tyll, line 10:
```suggestion
- Ethtool coalesce settings support

File: CHANGELOG.md [hunk 5]
```diff
@@ -1,4 +1,18 @@
 # Changelog
+## [1.3.0] - 2021-03-24
+### Changes
```
ffmancera, line 3: Instead of changes, use `Breaking changes` and `Bug fixes`. For the inclusive language each parameter, should have attached its own description and should be on `Breaking Changes`

CHANGELOG.md
ffmancera: Explain each parameter deprecated in favor of the new one.

ffmancera (changes requested):

CHANGELOG.md
tyll: FYI: the previous changelog entry used "Changes" not "Breaking Changes" since the role does not really break things at this time.

thom311: Right!! I almost wanted to comment "but the role doesn't break API, does it"?

CHANGELOG.md
liangwen12year: They are the same DNS options as resolv.conf(5) describing

liangwen12year: I will create a new PR describing the available DNS options

CHANGELOG.md
ffmancera: Well, IMO, deprecating something is a "breaking change" even if we are not dropping it in this version, it will be in the future.. But I am fine then to use "Changes" if the previous entry did.

CHANGELOG.md
liangwen12year: But currently, we only support DNS options for **IPv4**

CHANGELOG.md
liangwen12year: updated, thanks

CHANGELOG.md
thom311: "deprecating" does not necessarily mean that it will be dropped in the future (https://en.wikipedia.org/wiki/Deprecation)

Taking deprecation as an excuse to break API later, is IMO bad (but common). We should not do that.

In any case, it would be a breaking change only in the future when it gets actually dropped.


Maybe one could say: but now I get a deprecation warning, and that already broke my setup. E.g. if libnm marks API as deprecated, you may get a compile time error with `-Wdeprecated-declarations=error`. Now one might argue whether that really counts as broken or still within acceptable limits. But if that is the concern, then the deprecation warning is too aggressive, and should be toned down. But not say: "well of course it broke, we depreacted it!".

CHANGELOG.md
thom311: DNS should not be specific to the address family. NetworkManager gets this wrong with `ipv4.dns-options` and `ipv6.dns-options`.

How to correctly abstract that in the role, is a different question.

CHANGELOG.md
```diff
@@ -1,4 +1,38 @@
...
+- Use inclusive language
+Use inclusive language in docs, variable names, function names and parameter names
+  - "slave" is deprecated in favor of "port"
+  - "master" is deprecated in favor of "controller"
+
+### New features
+- Support ipv6.method disabled
+- Support dns_options for IPv4 config
+
+  The list of supported DNS options is:
```
tyll, line 13: This is too verbose for the changelog

CHANGELOG.md
```diff
@@ -1,4 +1,38 @@
...
+## [1.3.0] - 2021-03-24
+### Changes
+- Use inclusive language
+Use inclusive language in docs, variable names, function names and parameter names
+  - "slave" is deprecated in favor of "port"
+  - "master" is deprecated in favor of "controller"
+
+### New features
+- Support ipv6.method disabled
+- Support dns_options for IPv4 config
```
tyll, line 11:
```suggestion
- Support dns_options when using one or more IPv4 nameservers
```

CHANGELOG.md
tyll: Using deprecation is not an excuse but a warning so that users can start to migrate away from it.

CHANGELOG.md
liangwen12year: How about now?

CHANGELOG.md
```diff
@@ -1,4 +1,36 @@
 # Changelog
+## [1.3.0] - 2021-03-24
+### Changes
+- Use inclusive language
+Use inclusive language in docs, variable names, function names and parameter names
```
tyll, line 5: This is repeating `Use inclusive language`, mentioning "docs, variable names, ..." is too much detail in my opition.

CHANGELOG.md
```diff
@@ -1,4 +1,36 @@
...
+  - "slave" is deprecated in favor of "port"
+  - "master" is deprecated in favor of "controller"
+
+### New features
+- Support ipv6.method disabled
```
tyll, line 10:
```suggestion
- Support disabling IPv6
```

CHANGELOG.md
tyll: Please remove the list of options from here.

CHANGELOG.md
```diff
@@ -1,4 +1,36 @@
...
+  - `no-reload`, disables automatic reloading of a changed configuration file.
+  - `no-tld-query`, attempt not to resolve an unqualified name as if it were a top level domain(TLD).
+  - `rotate`, round-robin selection of name servers from among those listed.
+  - `single-request`, this option makes glibc perform the IPv6 and IPv4 requests sequentially.
+  - `single-request-reopen`, open a new socket for the second request if two requests from same port are handled incorrectly.
+  - `timeout:n`, the amount of time the resolver will wait for a response from a remote name server.
+  - `trust-ad`, this option controls the AD bit behavior of the stub resolver.
+  - `use-vc`, this option forces the use of TCP for DNS resolutions.
+
+- Ethtool coalesce settings support
```
tyll, line 28:
```suggestion
- Support Ethtool coalesce settings
```

cathay4t (approved):

CHANGELOG.md
liangwen12year: I changed into "Support Ethtool coalesce settings".

CHANGELOG.md
liangwen12year: The list of DNS options was removed.

CHANGELOG.md
liangwen12year: Thanks, I changed into "Support disabling IPv6".

CHANGELOG.md
liangwen12year: I changed into 
```
- Use inclusive language
  - "slave" is deprecated in favor of "port"
  - "master" is deprecated in favor of "controller"
  ```

File: CHANGELOG.md (hunk 7 above)
Comment (liangwen12year) [reply CHANGELOG.md]: I changed into "Support dns_options when using one or more IPv4 nameservers".

**Comment by liangwen12year** [comment]
@cathay4t , rebased. Thanks.

**Comment by tyll** [comment]
This will probably need some fixes when #378 is merged.

File: CHANGELOG.md [hunk 11]
```diff
@@ -1,4 +1,19 @@
 # Changelog
+## [1.3.0] - 2021-03-24
+### Changes
+- Use inclusive language
+  - "slave" is deprecated in favor of "port"
+  - "master" is deprecated in favor of "controller"
```
tyll, line 6: Please use backticks for the technical terms instead of double quotes.

liangwen12year: I replaced double quotes with the backticks.

.mdl_style.rb
```diff
@@ -5,3 +5,5 @@
 rule 'MD013', :line_length => 88
 # https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix
 rule 'MD029', :style => :ordered
+# https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md024---multiple-headers-with-the-same-content
+rule "MD024", :allow_different_nesting => true
```
tyll, line 9: This should be its own commit with an explanatory commit message.

CHANGELOG.md
```diff
@@ -1,6 +1,26 @@
 Changelog
 =========
 
+[1.3.0] - 2021-03-24
```
tyll, line 4: You will need to update the date here when the actual release is published. @ffmancera when is this planned?

.mdl_style.rb
liangwen12year: I created a new commit for the above change

tyll: Please explain in the commit message why we need this change.

liangwen12year: I explained as following:
```
`markdownlint` check failed on CHANGELOG.md with error "Multiple
headers with the same content", because the `markdownlint` parameter
`allow_different_nesting` defaults to be `false`. To allow header
duplication under different nesting level, we need to explicitly set
`allow_different_nesting` to be true.
```

CHANGELOG.md
ffmancera: In the moment it is approved. Maybe we can say today? :-)

ffmancera (approved):
LGTM, thanks!

.mdl_style.rb
tyll: I would prefer an explanation that gives the technical reason why `allow_different_nesting=True` is necessary. Using the default value of `False` as reason is not a convincing reason. It is similar to argue that a function was added because it was not there in the beginning.

A possible alternative would be, to give you an example:

Markdownlint requires each section headers to be different by default but the Changelog needs to contain multiple sections with the same name, therefore enable `allow_different_nesting` to support this use case.

I will not block this PR on this, though.

CHANGELOG.md
liangwen12year: I updated into `[1.3.0] - 2021-04-08`

liangwen12year: I updated into `[1.3.0] - 2021-04-08`

.mdl_style.rb
liangwen12year: I updated with the examples that you provide, thanks.

tyll:
#311 seems to be ready to be merged, too. So if it is merged before this one, the changelog needs to be updated.

tyll (approved):
LGTM

PR #369, opened 2021-03-24
do not setup centos-6 repos; enable EPEL where needed

richm:
The CI system correctly sets up centos-6 repos, so the network role
should not do it.
Make sure EPEL is correctly enabled where needed.


jharuda:
[citest]

jharuda:
[citest]

jharuda:
[citest]

richm:
[citest bad]

tests/tasks/el_repo_setup.yml
```diff
@@ -1,26 +0,0 @@
...
-
-      [extras]
-      name=CentOS-$releasever - Extras
-      baseurl=https://vault.centos.org/6.10/extras/$basearch/
-      gpgcheck=1
-      gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-6
-  when:
-    - ansible_distribution == 'CentOS'
-    - ansible_distribution_major_version == '6'
-- include_tasks: enable_epel.yml
```
ffmancera, line 26: Is this not needed for running the tests locally using a centos6 image? Thanks!

richm: > Is this not needed for running the tests locally using a centos6 image? Thanks!

It is needed for running the tests locally using a centos6 image.  But I don't think we want it in the test, in this form, since the CI environment already does this.

cathay4t: @richm 
This is hiding information in a black box CI and I have to explain when anyone hit this problem locally.

Is there any problem caused by old code?

PR #371, opened 2021-03-24
add test cleanup/state restore for certain tests; add test tags

richm:
Add test state save/state restore for cleanup for certain tests
that we run on a single machine in CI.  Add tags for other tests
that we use in CI.

### DIFF
--- a/tests/playbooks/tests_802_1x.yml
+++ b/tests/playbooks/tests_802_1x.yml
@@ -1,5 +1,10 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
+- name: Save host state
+  hosts: all
+  tasks:
+    - import_tasks: tasks/save_state.yml
+
 - hosts: all
   vars:
     interface: 802-1x-test
@@ -122,3 +127,8 @@
               command: update-ca-trust
           tags:
             - "tests::cleanup"
+
+- name: Restore host state
+  hosts: all
+  tasks:
+    - import_tasks: tasks/restore_state.yml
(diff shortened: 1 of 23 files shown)


tests/playbooks/tests_802_1x.yml
```diff
@@ -122,3 +127,8 @@
               command: update-ca-trust
           tags:
             - "tests::cleanup"
+
+- name: Restore host state
```
tyll, line 131: For other cleanup tasks, we added them in an always block to ensure that they always run. Do we need this here, too?

tests/playbooks/tests_bond.yml
```diff
@@ -8,6 +13,8 @@
     dhcp_interface1: test1
     port2_profile: bond0.1
     dhcp_interface2: test2
+  tags:
+    - "tests::expfail"
```
tyll, line 17: Why is this expected to fail? Would be good to explain/link to an issue/BZ to track this.

tests/tasks/restore_state.yml
```diff
@@ -0,0 +1,24 @@
...
+    path: /etc/sysconfig/network
+    state: absent
+  when:
+    - etc_sysconfig_network_stat is defined
+    - not etc_sysconfig_network_stat.stat.exists
+
+- name: Restore services
+  import_tasks: restore_services_state.yml
+
+- name: reload NetworkManager
```
tyll, line 15: Is it a problem when NetworkManager or other services are installed during the test and not removed/disabled afterwards?

jharuda:
[citest]

jharuda:
[citest]

PR #372, opened 2021-03-28
Fix: support dns_search and dns_options for all address family

liangwen12year:


library/network_connections.py
```diff
@@ -993,16 +995,19 @@ def connection_create(self, connections, idx, connection_current=None):
...
-                if d["family"] == socket.AF_INET:
-                    s_ip4.add_dns(d["address"])
-            for s in ip["dns_search"]:
-                s_ip4.add_dns_search(s)
-            s_ip4.clear_dns_options(True)
-            for s in ip["dns_options"]:
-                s_ip4.add_dns_option(s)
-
+                if not ipv4_is_disabled:
+                    if d["family"] == socket.AF_INET:
```
tyll, line 999: This will silently ignore IPv4 DNS settings if IPv4 is disabled but this needs to create a failure.

liangwen12year: make sense, thanks

ffmancera (changes requested):
Please could you explain why is this needed on the commit message? Thanks!

library/network_connections.py
liangwen12year: Code updated.
I suddenly realized that we should raise errors when `dns_search` (or `dns_options`) is defined but dns nameserver is not defined. Right?

liangwen12year:
> Please could you explain why is this needed on the commit message? Thanks!

Explained, thanks.

library/network_connections.py
```diff
@@ -993,16 +995,24 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                        s_ip4.add_dns(d["address"])
+                        for s in ip["dns_search"]:
+                            s_ip4.add_dns_search(s)
+                        s_ip4.clear_dns_options(True)
+                        for s in ip["dns_options"]:
+                            s_ip4.add_dns_option(s)
+                else:
+                    if d["family"] == socket.AF_INET:
```
thom311, line 1007: the current version iterates over all `ip["dns"]`, and whenever it find a suitable IPv6 address, it does:
```
    for s in ip["dns_search"]:
        s_ip6.add_dns_search(s)
```
That's not really a problem, because `add_dns_search()` will not add duplicate entries. But it still seems odd. Can we do this outside of the loop?

How about to unwrap this loop to:

```python
    for d in ip["dns"]:
        if d["family"] != socket.AF_INET6:
            continue
        if ipv6_is_disabled:
            raise MyError(
                         "when IPv6 is disabled, DNS setting for IPv6 nameservers is not allowed."
            )
        s_ip6.add_dns(d["address"])
    if not ipv6_is_disabled:
       for s in ip["dns_search"]:
           s_ip6.add_dns_search(s)
        s_ip6.clear_dns_options(True)
        for s in ip["dns_options"]:
            s_ip6.add_dns_option(s)
```
(which has a different behavior, of also configuring the DNS option/searches, if there are not IPv6 name servers. But that's fine, you might get the nameserver via DHCP).

liangwen12year: @thom311 , do you mean that we can get nameserver no matter what ? Even if `ip["dns"] == []`, we can always get nameserver via DHCP ? So, as long as `ipv6.method != disabled`, we should configure the DNS option/searches.

liangwen12year: @thom311 , do you mean that we can get nameserver no matter what ? Even if `ip["dns"] == []`, we can always get nameserver via DHCP ? So, as long as `ipv6.method != disabled`, we should configure the DNS option/searches.

thom311: yes, that's what I meant.

library/network_connections.py
```diff
@@ -1026,8 +1036,19 @@ def connection_create(self, connections, idx, connection_current=None):
                     NM.SETTING_IP_CONFIG_ROUTE_METRIC, ip["route_metric6"]
                 )
             for d in ip["dns"]:
```
tyll, line 1038:
```suggestion
            for nameserver in ip["dns"]:
```

library/network_connections.py
```diff
@@ -1026,8 +1036,19 @@ def connection_create(self, connections, idx, connection_current=None):
...
-                if d["family"] == socket.AF_INET6:
-                    s_ip6.add_dns(d["address"])
+                if d["family"] != socket.AF_INET6:
+                    continue
+                if ipv6_is_disabled:
+                    raise MyError(
+                        "when IPv6 is disabled, DNS setting for IPv6 nameservers is not allowed."
```
tyll, line 1043:
```suggestion
                        "IPv6 needs to be enabled to support IPv6 nameservers."
```

library/network_connections.py
```diff
@@ -1026,8 +1036,19 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                    )
+                s_ip6.add_dns(d["address"])
+            if not ipv6_is_disabled:
+                for s in ip["dns_search"]:
```
tyll, line 1047:
```suggestion
                for search_domains in ip["dns_search"]:
```

library/network_connections.py
```diff
@@ -1026,8 +1036,19 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                    s_ip6.add_dns_search(s)
+                s_ip6.clear_dns_options(True)
+                for s in ip["dns_options"]:
```
tyll, line 1050:
```suggestion
                for option in ip["dns_options"]:
```

library/network_connections.py
```diff
@@ -1025,9 +1035,20 @@ def connection_create(self, connections, idx, connection_current=None):
                 s_ip6.set_property(
                     NM.SETTING_IP_CONFIG_ROUTE_METRIC, ip["route_metric6"]
                 )
-            for d in ip["dns"]:
-                if d["family"] == socket.AF_INET6:
-                    s_ip6.add_dns(d["address"])
+            for nameserver in ip["dns"]:
+                if nameserver["family"] != socket.AF_INET6:
+                    continue
```
tyll, line 1040: Why are you using the negation with continue here? It seems to me to be easier to understand to use a non-negated if statement here without continue.

library/network_connections.py
```diff
@@ -1025,9 +1035,20 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                if ipv6_is_disabled:
+                    raise MyError(
+                        "IPv6 needs to be enabled to support IPv6 nameservers."
+                    )
+                s_ip6.add_dns(nameserver["address"])
+            if not ipv6_is_disabled:
+                for search_domains in ip["dns_search"]:
+                    s_ip6.add_dns_search(search_domains)
```
tyll, line 1048:
```suggestion
                for search_domain in ip["dns_search"]:
                    s_ip6.add_dns_search(search_domain)
```

library/network_connections.py
thom311: sure.

The "continue" is not too bad IMO, because then you walk through the loop and abort it early (when a certain condition is satisfied). The alternative leads to deeper nested code/block, which IMO hurt readability.

Of course, not a big issue either way. But that was my reasoning.

library/network_connections.py
```diff
@@ -1025,9 +1035,20 @@ def connection_create(self, connections, idx, connection_current=None):
...
-            for d in ip["dns"]:
-                if d["family"] == socket.AF_INET6:
-                    s_ip6.add_dns(d["address"])
+            for nameserver in ip["dns"]:
+                if nameserver["family"] != socket.AF_INET6:
+                    continue
+                if ipv6_is_disabled:
+                    raise MyError(
+                        "IPv6 needs to be enabled to support IPv6 nameservers."
+                    )
```
tyll, line 1044: This might be better checked in validate_post, @thom311 what do you think? I believe to remember that you might have suggested this in the past, too.

library/network_connections.py
tyll: If the check for disabled IPv6 is done in the argument validator (which might be the actual place to do this), then the for loop is just one statement, though.

library/network_connections.py
liangwen12year: I changed into `search_domain`

library/network_connections.py
thom311: good point. Errors should be caught as early as possible, and if a constraint only depends on settings from the input YAML, then such misconfiguration should already be caught when validating/normalizing the YAML. Tha's what `ArgValidator` does. In the first step, its validation only depends on the input YAML alone.

library/network_connections.py
thom311: True!! When I originally suggested the snippet, I thought there would be more inside the loop.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1802,3 +1802,22 @@ def validate_connection_one(self, mode, connections, idx):
...
+            if not connection["ip"]["dhcp4"] and not addrs4:
+                raise ValidationError.from_connection(
+                    idx,
+                    "IPv4 needs to be enabled to support IPv4 nameservers.",
+                )
+            if connection["ip"]["ipv6_disabled"]:
+                raise ValidationError.from_connection(
+                    idx,
+                    "IPv6 needs to be enabled to support IPv6 nameservers.",
+                )
```
liangwen12year, line 1823: @thom311 , in network role, there is no explicit option like `ipv4_disabled`, how do you like the validation 
```
            if not connection["ip"]["dhcp4"] and not addrs4:
                raise ValidationError.from_connection(
                    idx,
                    "IPv4 needs to be enabled to support IPv4 nameservers.",
                )
 ```

File: library/network_connections.py (hunk 7 above)
Comment (liangwen12year) [reply library/network_connections.py]: I changed into non-negated if statement without `continue`:
```
            for nameserver in ip["dns"]:
                if nameserver["family"] == socket.AF_INET6:
                    s_ip6.add_dns(nameserver["address"])
```

File: library/network_connections.py (hunk 3 above)
Comment (liangwen12year) [reply library/network_connections.py]: `d` was changed into `nameserver`

File: library/network_connections.py (hunk 4 above)
Comment (liangwen12year) [reply library/network_connections.py]: I changed into "IPv6 needs to be enabled to support IPv6 nameservers."

File: library/network_connections.py (hunk 5 above)
Comment (liangwen12year) [reply library/network_connections.py]: `s` was changed into `search_domain`

File: library/network_connections.py (hunk 6 above)
Comment (liangwen12year) [reply library/network_connections.py]: `s` was changed into `option`

File: library/network_connections.py (hunk 9 above)
Comment (liangwen12year) [reply library/network_connections.py]: I added the following into function `validate_connection_one()`: 

```
        if connection["ip"]["dns_options"] or connection["ip"]["dns_search"]:
            addrs4 = list(
                [
                    address
                    for address in connection["ip"]["address"]
                    if address["family"] == socket.AF_INET
                ]
            )
            if not connection["ip"]["dhcp4"] and not addrs4:
                raise ValidationError.from_connection(
                    idx,
                    "IPv4 needs to be enabled to support IPv4 nameservers.",
                )
            if connection["ip"]["ipv6_disabled"]:
                raise ValidationError.from_connection(
                    idx,
                    "IPv6 needs to be enabled to support IPv6 nameservers.",
                )
```

File: module_utils/network_lsr/argument_validator.py [hunk 11]
```diff
@@ -1802,3 +1802,23 @@ def validate_connection_one(self, mode, connections, idx):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+        # when IPv4 or IPv6 is disabled, settings for ip.dns_options and
+        # ip.dns_search are not allowed
+        if connection["ip"]["dns_options"] or connection["ip"]["dns_search"]:
+            addrs4 = list(
+                [
+                    address
```
thom311, line 1810: minor comment: here you use a list comprehension with `[]`. That already generates a list. The code `list([ ... ])` generates a second list (throwing away the first). Maybe python can optimize that, IDK. But it also is verbose to read.

Also you could write `any_addrs4 = any( [list comprehension] )` as you don't actually need the entire list `addrs4`. When you do that, you could use a generator comprehension instead of a list comprehension. The difference is that `any()` only needs to iterate until it find one true value. It does not need the entire list, and neither does it need to generate the enitre list in the first place.

Meaning, I'd do:

```python
has_addrs4 = any( a["family"] == AF_INET for a in connection["ip"]["address"] )
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -1802,3 +1802,23 @@ def validate_connection_one(self, mode, connections, idx):
...
+            )
+            if not connection["ip"]["dhcp4"] and not addrs4:
+                raise ValidationError.from_connection(
+                    idx,
+                    "IPv4 needs to be enabled to support IPv4 nameservers.",
+                )
+            if connection["ip"]["ipv6_disabled"]:
+                raise ValidationError.from_connection(
+                    idx,
+                    "IPv6 needs to be enabled to support IPv6 nameservers.",
```
thom311, line 1823: the error message talks about "nameservers", but here we are inside a
```
if connection["ip"]["dns_options"] or connection["ip"]["dns_search"]:
```
block. That is, it does not check the nameservers, but other DNS settings.

More importantly, setting "dns_search" and/or "dns_option" is fine if **any** address family is enabled. The current code would reject setting these DNS settings, if any of the address families is disabled.

module_utils/network_lsr/argument_validator.py
liangwen12year: Then the `list()` here is also redundant,
https://github.com/linux-system-roles/network/blob/main/library/network_connections.py#L972

thom311: it is!

module_utils/network_lsr/argument_validator.py
liangwen12year: I do not understand "More importantly, setting "dns_search" and/or "dns_option" is fine if any address family is **enabled**. " Should this be **disabled** instead?

thom311: it is OK set set dns-search/dns-options, if at least one of the address families is enabled.

The current code checks that both are enabled.

module_utils/network_lsr/argument_validator.py
liangwen12year: I changed into
```
            has_addrs4 = any(
                address["family"] == socket.AF_INET
                for address in connection["ip"]["address"]
            )
```

module_utils/network_lsr/argument_validator.py
liangwen12year: I just realized the validation for dns and dns_search/dns_options should be different:
```
        # Setting ip.dns is not allowed when corresponding IP method for that
        # nameserver is disabled
        for nameserver in connection["ip"]["dns"]:
            has_addrs4 = any(
                address["family"] == socket.AF_INET
                for address in connection["ip"]["address"]
            )
            if nameserver["family"] == socket.AF_INET and (
                not connection["ip"]["dhcp4"] and not has_addrs4
            ):
                raise ValidationError.from_connection(
                    idx,
                    "IPv4 needs to be enabled to support IPv4 nameservers.",
                )
            if (
                nameserver["family"] == socket.AF_INET6
                and connection["ip"]["ipv6_disabled"]
            ):
                raise ValidationError.from_connection(
                    idx,
                    "IPv6 needs to be enabled to support IPv6 nameservers.",
                )
        # when IPv4 and IPv6 are disabled, setting ip.dns_options or
        # ip.dns_search is not allowed
        if connection["ip"]["dns_search"] or connection["ip"]["dns_options"]:
            has_addrs4 = any(
                address["family"] == socket.AF_INET
                for address in connection["ip"]["address"]
            )
            if (not connection["ip"]["dhcp4"] and not has_addrs4) and connection["ip"][
                "ipv6_disabled"
            ]:
                raise ValidationError.from_connection(
                    idx,
                )
```

library/network_connections.py
ffmancera: +1

liangwen12year:
@thom311 , should we all more dns options specifically for ipv6 nameservers in argvalidator ? see https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/libnm-core-impl/nm-setting-ip-config.c#L36

thom311:
> should we all more dns options specifically for ipv6 nameservers in argvalidator ?

I don't understand the question. What do you mean?

thom311:
lgtm

liangwen12year:
> lgtm

thx

module_utils/network_lsr/argument_validator.py
```diff
@@ -1806,3 +1810,57 @@ def validate_connection_one(self, mode, connections, idx):
...
+        # Setting ip.dns is not allowed when corresponding IP method for that
+        # nameserver is disabled
+        for nameserver in connection["ip"]["dns"]:
+            has_addrs4 = any(
+                address["family"] == socket.AF_INET
+                for address in connection["ip"]["address"]
+            )
+            if nameserver["family"] == socket.AF_INET and (
+                not connection["ip"]["dhcp4"] and not has_addrs4
+            ):
```
tyll, line 1822: Would be great to introduce a helper function that reduces the repeated code:

```suggestion
            if nameserver["family"] == socket.AF_INET and not _ipv4_enabled(connection):
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -1806,3 +1810,57 @@ def validate_connection_one(self, mode, connections, idx):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+        # Setting ip.dns is not allowed when corresponding IP method for that
+        # nameserver is disabled
+        for nameserver in connection["ip"]["dns"]:
```
tyll, line 1815: It would be great to have unit tests to cover the scenarios here.

module_utils/network_lsr/argument_validator.py
liangwen12year: Helper funciton `def _ipv4_enabled(connection)` created

module_utils/network_lsr/argument_validator.py
liangwen12year: Added unit tests function `test_dns_argvalidator()`

tests/unit/test_network_connections.py
```diff
@@ -3312,6 +3312,114 @@ def test_dns_options_argvalidator(self):
             validator.validate(true_testcase_12)["dns_options"], ["use-vc"]
         )
 
+    def test_dns_argvalidator(self):
+        """
+        Test that configuring ip.dns is not allowed when corresponding IP method for
+        that nameserver is disabled.
+        """
+        input_connection1 = [
```
tyll, line 3320: Please use more descriptive names here, what is this testing?

tests/unit/test_network_connections.py
```diff
@@ -3312,6 +3312,114 @@ def test_dns_options_argvalidator(self):
...
+            input_connection1,
+            0,
+        )
+        self.assertRaises(
+            ValidationError,
+            validator.validate_connection_one,
+            "nm",
+            input_connection2,
+            0,
+        )
```
tyll, line 3421: These should be two test cases, one for IPv4 and one for IPv6

module_utils/network_lsr/argument_validator.py
```diff
@@ -1733,6 +1737,13 @@ def _validate_post(self, value, name, result):
     VALIDATE_ONE_MODE_INITSCRIPTS = "initscripts"
 
     def validate_connection_one(self, mode, connections, idx):
```
tyll, line 1739: @thom311 what is the purpose of `validate_connection_one()`? Is this the right place for this check? What does `one` relate to, here? Is it one provider?

tests/unit/test_network_connections.py
```diff
@@ -3312,6 +3312,114 @@ def test_dns_options_argvalidator(self):
...
+            }
+        ]
+        validator = network_lsr.argument_validator.ArgValidator_ListConnections()
+        self.assertRaises(
+            ValidationError,
+            validator.validate_connection_one,
+            "nm",
+            input_connection1,
+            0,
+        )
```
tyll, line 3414: This can be simplified by using something like

```
validator.validate_connection_one,  "nm", validator.validate(ipv6_dns_without_ipv6_config), 0)
ipv6_dns_without_ipv6_config = MINIMAL_EXAMPLE
```

module_utils/network_lsr/argument_validator.py
thom311: It "validate"s "one" "connection", while also taking into account the provider/mode.

It's noteworthy that it validates one connection (`connections[idx]`), because `ArgValidator_ListConnections` in general is about a list of connections.

The earlier validation steps done by ArgValidator are independent of the mode.

tyll: Why are those separated? `validate_connection_one()` is only used in a loop to validate all connections.

thom311: because the caller want's to know which of the connections was invalid. If `validate_connection_all()` would only throw an exception, it would need to somehow propagate `idx`. It seems better to have smaller, self-contained functions, that do something understandable. E.g. it's understandable that `validate_connection_one()` does either nothing or fail with a `ValidationError` on error. It's less clear how it would behave, when validating all connections at once.

But yes, it could also be different...

tyll: There are already a lot of validations happening in `.validate()` before `validdate_connetion_one()` is called and there the index is reported in the exception message:
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1696

thom311: `ArgValidator_ListConnections._validate_post()` checks interdependencies between profiles, so necessarily that processes the entire list of connection. On the other hand, `ArgValidator._validate_impl()` indeed loads (and validates) each element one-by-one, indpendently. But in general, the API or behavior of those two could not reasonably be any different and if they need to report the `idx`, then the should attach it to `ValidationError` (and kind of they already do, via the `name` and `error_message`).

The API cannot be reasonably different than what it is, is because ArgValidator should be nestable/composable, so there is no place where somebody would drive the creation/validation of individual list elements (otherwise, you'd have to know that the element you are creating is a `ArgValidatorList` type). Also, `.validate()` not only validates but normalizes (modifies) the data.

On the other hand, `validate_connection_one()` does not modify/normalize anything. Also, the caller is well aware that they want to specifically validate a connection of `ArgValidator_ListConnections`. It is not abstract "ArgValidator.validate_one_list_element()`, because the caller necessarily knows what `ArgValidator_ListConnections` and `validate_connection_one()` does. As such, the API can be different, and it is.


Of you dislike that, please open a merge request.

tests/unit/test_network_connections.py
liangwen12year: We can rely on `Validator.validate()` to normalize and fill unspecified values.

tests/unit/test_network_connections.py
liangwen12year: Created separated test cases:
`def test_ipv6_dns_without_ipv6_config(self)`
`def test_ipv4_dns_without_ipv4_config(self)`

tests/unit/test_network_connections.py
liangwen12year: Test cases were simplified.

tests/unit/test_network_connections.py
liangwen12year: I used the descriptive name for connections:
```python
        ipv4_dns_without_ipv4_config = [
            {
                "name": "test_ipv4_dns",
                "type": "ethernet",
                "ip": {
                    "auto6": True,
                    "dhcp4": False,
                    "dns": ["198.51.100.5"],
                },
            }
        ]
        
        ipv6_dns_without_ipv6_config = [
            {
                "name": "test_ipv6_dns",
                "type": "ethernet",
                "ip": {
                    "ipv6_disabled": True,
                    "dhcp4": True,
                    "dns": ["2001:db8::20"],
                },
            }
        ]
```

library/network_connections.py
cathay4t: @liangwen12year Any update on this?

cathay4t (changes requested):
Missing code for checking whether ipv6 dns option has ipv6 name server defined.

liangwen12year:
> Missing code for checking whether ipv6 dns option has ipv6 name server defined.

We can always get nameserver via DHCP as long as ipv6.method != disabled, so only raise errors when ipv6 dns option clearly defined and ipv6 is disabled. See the checking here, https://github.com/linux-system-roles/network/pull/372/commits/d9d02119ed1ee0f7977a03fd751e8736a27b1166#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1846

Did you mean that missing testing code here? Yes,  I will add the testing code.

library/network_connections.py
liangwen12year: I think the correct  validation is:
- raise errors when ipv6 dns option is defined and ipv6 is disabled
- raise errors when `dns_search` (or `dns_option`) is defined but both ipv6 and ipv4 is disabled.

Bear in mind that we do not need to clearly define nameserver, we can always get nameserver via DHCP as long as corresponding ip method is not disabled. Correct me if I was wrong , @thom311 ?

liangwen12year:
> > Missing code for checking whether ipv6 dns option has ipv6 name server defined.
> 
> We can always get nameserver via DHCP as long as ipv6.method != disabled, so only raise errors when ipv6 dns option clearly defined and ipv6 is disabled. See the checking here, [d9d0211#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1846](https://github.com/linux-system-roles/network/commit/d9d02119ed1ee0f7977a03fd751e8736a27b1166#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1846)
> 
> Did you mean that missing testing code here? Yes, I will add the testing code.

Testing code added here, https://github.com/linux-system-roles/network/pull/372/commits/496bbde4d3b03ee924397b5a914ea5a80d24e6c3#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359R3363

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

ffmancera (approved):
LGTM, thanks! @cathay4t do you agree with the changes?

cathay4t (approved):

cathay4t:
Looks good to me. Waiting rebase.

liangwen12year:
> Looks good to me. Waiting rebase.

Rebased. Thanks.

cathay4t:
[citest bad]

PR #373, opened 2021-03-29
[th/cleanup-default-val-in-argvalidator] two minor cleanup patches

thom311:


module_utils/network_lsr/argument_validator.py
```diff
@@ -353,16 +353,16 @@ def _validate_impl(self, value, name):
...
                 continue
             if validator.required:
                 raise ValidationError(name, "missing required key '%s'" % (setting))
-            default_value = validator.get_default_value()
-            if (
-                not self.all_missing_during_validate
-                and default_value is not ArgValidator.MISSING
-            ):
-                result[setting] = default_value
+            if not self.all_missing_during_validate:
```
tyll, line 361: Since this changes the logic, is there a bug that is fixed by this? Why does this need to be changed?

thom311: it slightly changes the logic. It now avoids calling `validator.get_default_value()`, when we already know that it's not used (`self.all_missing_during_validate`). It's unlikely that either has a visible effect.

It still feels more correct to only get the information, if we also are going to use it. The logic is also how someone would think. I mean, I would think "the default value is only requested `if not self.all_missing_during_validate`". That is easier to understand than "the default value is always requested, but only used if `if not self.all_missing_during_validate`.

tyll: ok, thanks

liangwen12year:
LGTM!

liangwen12year:
Please rebase it.

liangwen12year (approved):

ffmancera (approved):
LGTM, please rebase it. Thanks!

liangwen12year:
Looks like this PR can not be merged.
image

ffmancera (dismissed):
Hi, the commit message title is too long. Could you fit it in 50 characters or less? Thanks!

tyll:
> Looks like this PR can not be merged.
> image

This setting was not meant to be enabled. Disabled it.

tyll:
[citest bad]

PR #375, opened 2021-04-02
update to tox-lsr 2.4.0 - add support for ansible-test sanity with docker

richm:
set `LSR_ANSIBLE_TEST_DOCKER: "true"` so that we run ansible-test sanity
with docker
This should fix issues we are seeing with tox ansible-test failures


nhosoi (approved):
lgtm

tyll (approved):
:+1:

PR #377, opened 2021-04-07
use lsr-gh-action-py26 for python 2.6 unit tests

richm:
Use the new https://github.com/linux-system-roles/lsr-gh-action-py26 for running unit tests on python 2.6.  This uses an EL6 VM so more closely corresponds to the managed hosts that still use python 2.6.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -43,36 +43,9 @@ jobs:
           esac
           TOXENV="$toxenvs" lsr_ci_runtox
   python-26:
-    runs-on: ubuntu-16.04
-    env:
-      PY26URL: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-2.6.tar.bz2
-      PY26TARFILE: "python-2.6.tar.bz2"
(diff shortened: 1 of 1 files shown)


tyll:
DCO is missing, otherwise LGTM

.github/workflows/tox.yml
```diff
@@ -42,36 +42,9 @@ jobs:
...
-          lsr_ci_preinstall
-          pip list
-      - name: Run tox tests
-        run: |
-          set -euo pipefail
-          source "$PY26VENVPATH/bin/activate"
-          set -x
-          tox -e py26,coveralls26
+      - name: Run py26 tests
+        uses: richm/lsr-gh-action-py26@setuptools
```
tyll, line 50: Oh, did not notice this before, this should be `linux-system-roles/lsr-gh-action-py26`, I guess.

tyll (changes requested):

.github/workflows/tox.yml
richm: And version `1.0.1`
fixed

tyll (approved):
Thank you, :+1:

PR #383, opened 2021-04-14
Document gateway4 and gateway6 setting in README.md

liangwen12year:

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,10 @@ The IP configuration supports the following options:
     When using a caching DNS plugin (dnsmasq or systemd-resolved in NetworkManager.conf)
     then "edns0" and "trust-ad" are automatically added.
 
+- `gateway4` and `gateway6`
+
+    The default gateway for IPv4 (`gateway4`) or IPv6 (`gateway6`) packets.
+
 - `route_metric4` and `route_metric6`
 
     For `NetworkManager`, `route_metric4` and `route_metric6` corresponds to the


README.md
```diff
@@ -387,6 +387,13 @@ The IP configuration supports the following options:
     When using a caching DNS plugin (dnsmasq or systemd-resolved in NetworkManager.conf)
     then "edns0" and "trust-ad" are automatically added.
 
+- `gateway4` and `gateway6`
+
+    The gateway associated with the IPv4 configuration and IPv6 configuration. This is
+    only meaningful if the corresponding IPv4 addresses or IPv6 addresses is also set.
```
tyll, line 393: This sentence is not grammatically correct since addresses is plural and "is" is singular.
Also, I guess the gateway is still useful when using any auto configuration or with only a single address.

README.md
```diff
@@ -387,6 +387,13 @@ The IP configuration supports the following options:
...
+    The gateway's main purpose is to control the next hop of the standard default route
```
tyll, line 394: If you write "main purpose" it implies that there are other purposes. What other purposes are there?

README.md
```diff
@@ -387,6 +387,13 @@ The IP configuration supports the following options:
...
+    on the device.
```
tyll, line 395:
```suggestion
    The default gateway for IPv4 (`gateway4`) or IPv6 (`gateway6`) packets.
```

tyll (approved):
:+1:

README.md
liangwen12year: The reason that I considered IPv4 addresses as singular is because in NM `addresses`  is a property[1]. Maybe I should quote the "addresses"( IPv4 "addresses" ):

image

[1] https://developer.gnome.org/NetworkManager/stable/nm-settings-nmcli.html

tyll: if you mean the property, a possible way to write it is: Specifying a gateway only works if the IPv6 `addresses` property is set.

On a technical level, I guess setting a gateway and enabling DHCP might also be correct but I might be wrong.

README.md
liangwen12year: "main purpose" removed.

tyll (approved):

PR #387, opened 2021-05-02
Normalize numeric value and reject bool arguments for ArgValidatorNum

liangwen12year:


jharuda:
[citest pending]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

ffmancera (changes requested):
Thanks for this, but could you please fix the commit message formatting? [PATCH 1/2] is used when generating multiple patch from `git format-patch`, but we do not use it when sending the patch directly to a branch in order to create a PR.

liangwen12year:
> Thanks for this, but could you please fix the commit message formatting? [PATCH 1/2] is used when generating multiple patch from `git format-patch`, but we do not use it when sending the patch directly to a branch in order to create a PR.

Commit message formatting fixed.

cathay4t:
@liangwen12year Please also update the comment 0 of this PR and title to help the reviewer on what and how this PR is fixing.

cathay4t (dismissed):
LGTM. Thanks!

module_utils/network_lsr/argument_validator.py
```diff
@@ -271,8 +271,12 @@ def __init__(  # pylint: disable=too-many-arguments
     def _validate_impl(self, value, name):
         v = None
         try:
-            if isinstance(value, self.numeric_type):
-                v = value
+            if self.numeric_type is not bool and isinstance(value, bool):
```
cathay4t, line 274: The class is called `ArgValidatorNum`, the `self.numeric_type is not bool`  should always true.
Did I miss anything?

thom311: `ArgValidatorNum` is an API, where the user can specify the numeric type. You are correct, that no user currently sets this to bool, so the condition is never satisfied.

However, implementing it also serves as a form of documentation for the fact that `bool` in python is a subtype of `int`, which requires some special considerations here. And, beside "documentation",  it just does the right thing in those cases too.

But yes, it could be dropped too!

liangwen12year: `self.numeric_type is not bool` was dropped.

liangwen12year:
> @liangwen12year Please also update the comment 0 of this PR and title to help the reviewer on what and how this PR is fixing.

Comment and title was updated. Thanks.

cathay4t (approved):
LGTM. Thanks

PR #392, opened 2021-05-14
tag 802-1x, bond tests as expfail; bridge test cleanup

richm:
The 802-1x test will fail on platforms where `hostapd` is not available,
so tag that test to make it skippable.

The initscripts bridge test does not clean up properly, leaving the
device around which causes the nm test to fail.  Explicitly remove
the device for cleanup.

The bond tests do not work when being run sequentially on the same
machine as in the smoke test scenario, so tag those tests so that
they can be skipped.

### DIFF
--- a/tests/playbooks/tests_802_1x.yml
+++ b/tests/playbooks/tests_802_1x.yml
@@ -3,6 +3,8 @@
 - hosts: all
   vars:
     interface: 802-1x-test
+  tags:
+    - tests::expfail
   tasks:
     - name: "INIT: 802.1x tests"
       debug:
--- a/tests/playbooks/tests_bond.yml
+++ b/tests/playbooks/tests_bond.yml
@@ -8,6 +8,8 @@
     dhcp_interface1: test1
     port2_profile: bond0.1
     dhcp_interface2: test2
+  tags:
+    - tests::expfail
   tasks:
     - name: "INIT Prepare setup"
       debug:
--- a/tests/playbooks/tests_bridge.yml
+++ b/tests/playbooks/tests_bridge.yml
@@ -49,6 +49,16 @@
     profile: "{{ interface }}"
     task: tasks/assert_profile_absent.yml
 
+- name: Remove test bridge
+  hosts: all
+  tags:
+    - tests::cleanup
+  tasks:
+    - name: Remove the test interface
+      command: ip link delete {{ interface | quote }}
+      ignore_errors: yes
+      changed_when: false
+
 # FIXME: Devices might still be left when profile is absent
 # - import_playbook: run_tasks.yml
 #   vars:
--- a/tests/tests_bond_deprecated_initscripts.yml
+++ b/tests/tests_bond_deprecated_initscripts.yml
@@ -10,5 +10,6 @@
         network_provider: initscripts
       tags:
         - always
+        - tests::expfail
 
 - import_playbook: playbooks/tests_bond_deprecated.yml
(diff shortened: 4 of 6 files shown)


tests/playbooks/tests_802_1x.yml
```diff
@@ -3,6 +3,8 @@
 - hosts: all
   vars:
     interface: 802-1x-test
+  tags:
+    - tests::expfail
```
tyll, line 7: In my opinion, it would be better to describe which platforms we want to skip as we do this for the wireless tests: https://github.com/linux-system-roles/network/blob/main/tests/ensure_provider_tests.py#L83

Also, if this is just because there is no EPEL 9, yet, then we also need to track this to ensure it will be enabled again at the right time.

tests/playbooks/tests_bond.yml
```diff
@@ -8,6 +8,8 @@
     dhcp_interface1: test1
     port2_profile: bond0.1
     dhcp_interface2: test2
+  tags:
+    - tests::expfail
```
tyll, line 12: Is there a Bugzilla/issue to track this?

tests/playbooks/tests_bridge.yml
```diff
@@ -49,6 +49,16 @@
...
+  hosts: all
+  tags:
+    - tests::cleanup
+  tasks:
+    - name: Remove the test interface
+      command: ip link delete {{ interface | quote }}
+      ignore_errors: yes
+      changed_when: false
+
 # FIXME: Devices might still be left when profile is absent
```
tyll, line 62: The fixme can be removed when this change is added.

tests/tests_bond_deprecated_initscripts.yml
```diff
@@ -10,5 +10,6 @@
         network_provider: initscripts
       tags:
         - always
+        - tests::expfail
```
tyll, line 13: Since this is a generated file, it should break the custom test: https://github.com/linux-system-roles/network/blob/main/.travis/custom.sh but it does not seem to be running according to https://github.com/linux-system-roles/network/pull/392/checks?check_run_id=2580561690#step:5:27

tests/playbooks/tests_bond.yml
richm: > Is there a Bugzilla/issue to track this?

I'm not sure - is it https://bugzilla.redhat.com/show_bug.cgi?id=1842614 or https://bugzilla.redhat.com/show_bug.cgi?id=1915017 or https://bugzilla.redhat.com/show_bug.cgi?id=1918210 or https://bugzilla.redhat.com/show_bug.cgi?id=1918252 ?

tests/tests_bond_deprecated_initscripts.yml
richm: > Since this is a generated file, it should break the custom test: https://github.com/linux-system-roles/network/blob/main/.travis/custom.sh but it does not seem to be running according to https://github.com/linux-system-roles/network/pull/392/checks?check_run_id=2580561690#step:5:27

Not sure why it isn't listed at https://github.com/linux-system-roles/network/blob/main/.github/workflows/tox.yml#L40-L43

tests/playbooks/tests_802_1x.yml
richm: It has to do with the availability of the `hostapd` package, which comes from epel, and the downstream tests do not or cannot use epel

cathay4t:
@richm We had many fixed to tests recently, do we still need this PR?

PR #396, opened 2021-06-24
nm: Fix the incorrect change indication when apply the same config twice

cathay4t:
## Patch 1

When applying the same network connections twice, the second apply still
shows `changed: true`.

The root cause:
 * When user never asked about ethtool configuration, network-role
   will generate an all-default `NM.SettingEthtool` and pass it to
   NetworkManager daemon. But NetworkManager discard it when saving to
   ifcfg plugin as ifcfg plugin will not keep empty ethtool option.

 * During second apply, the `NM.SimpleConnection.compare` will return
   False indicating configuration changed because of on-disk connection
   has no ethtool setting while pending connection does.

To fix it, we just remove the all-default `NM.SettingEthtool`.

## Patch 2

Fix the incorrect change indication for dns option

When applying with `dhcp4: "no"` or `auto6: "no"`, we get incorrect
change indication even when network connection was not changed.

The root cause is the `NM.SettingIPConfig.clear_dns_options(True)` will
create an empty list which will be discard by ifcfg plugin.
The follow up `NM.Connection.compare()` will show configuration changed
as dns option entry missing.

Fixed by remove dns option completely before appending.


thom311 (dismissed):

thom311:
fix in NM here: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/905

liangwen12year (approved):
LGTM, nice patch.

tyll:
It would be great to have a test for this that applies a certain profile twice that triggered this error and asserts that the module does not report "changed" for the second time the profile is applied.

tests/tests_change_indication_on_repeat_run.yml
```diff
@@ -0,0 +1,57 @@
...
+        - import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: "{{ interface }}"
+                state: up
+                type: ethernet
+                ip:
+                  dhcp4: "no"
+                  auto6: "no"
```
tyll, line 38: This variable can be set in line 5 for the playbook then it does not need to be repeated and it would ensure that the profiles are always the same.

cathay4t: I had trouble on this. Could you provide an example?

cathay4t:
@liangwen12year @thom311 I have included another patch fixing the DNS option change indication problem. Please review it again.

tests/tests_change_indication_on_repeat_run.yml
richm: > This variable

Which variable?

> I had trouble on this.

What did you try, and what was the result?

tests/tests_change_indication_on_repeat_run.yml
```diff
@@ -0,0 +1,85 @@
...
+              - name: "{{ interface }}"
+                state: up
+                type: ethernet
+                ip:
+                  address:
+                    - 192.0.2.2/24
+          register: __network_connections_result
+        - name: Assert change:false
+          assert:
+            that: __network_connections_result.changed == false
```
richm, line 42: `that: not __network_connections_result is changed`

tests/tests_change_indication_on_repeat_run.yml
```diff
@@ -0,0 +1,85 @@
...
+              - name: "{{ interface }}"
+                state: up
+                type: ethernet
+                ip:
+                  dhcp4: "no"
+                  auto6: "no"
+          register: __network_connections_result
+        - name: Assert change:false
+          assert:
+            that: __network_connections_result.changed == false
```
richm, line 70: see above

tests/tests_change_indication_on_repeat_run.yml
```diff
@@ -0,0 +1,85 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# This file was generated by ensure_provider_tests.py
+---
+- hosts: all
+  vars:
+    interface: testnic1
+    type: veth
```
tyll, line 7:
```suggestion
    type: veth
    network_connections:
      - name: testnic1
        state: up
        type: ethernet
        ip:
          address:
            - 192.0.2.2/24
```

cathay4t: Placing `network_connections` in top `vars` section does not works.  How can I use it in the `import_role`?

richm: > Placing `network_connections` in top `vars` section does not works. How can I use it in the `import_role`?

Can you use `include_role` instead of `import_role`?

tests/tests_change_indication_on_repeat_run.yml
cathay4t: I tried to place the `network_connections` at the host level variable, and then remove that from `import_role` task. Not work.
Even with this in `import_role` task will not work:

```yaml
    - block:
        - import_role:
          vars:
            - network_connections: " {{ network_connections }} "
          register: __network_connections_result
```

richm: not sure what the problem is - here is an example playbook calling a role with a variable defined at the host level - https://gist.github.com/richm/e9d338d8b39254b84b90fd2bf0b0b0a1

tests/tests_change_indication_on_repeat_run.yml
cathay4t: Yes. The `include_role` did the magic.

cathay4t: Thanks a lot.

cathay4t:
@tyll @richm Patch updated to use vars to mitigate the duplication in test code.

tests/tests_change_indication_on_repeat_run.yml
```diff
@@ -0,0 +1,71 @@
...
+  name: Test change indication on repeat run
+  tasks:
+    - include_tasks: tasks/manage_test_interface.yml
+      vars:
+        state: present
+    - include_tasks: tasks/assert_device_present.yml
+    - block:
+        - block:
+          - include_role:
```
tyll, line 16:
```suggestion
            - include_role:
```
Please adjust the indentation as reported  by Github actions.

spetrosi: Also, yamllint does not like tasks without names. So ensure to add a name for each task for code readability.

library/network_connections.py
```diff
@@ -976,6 +976,16 @@ def connection_create(self, connections, idx, connection_current=None):
                     else:
                         s_ethtool.option_set_uint32(nm_ring, setting)
 
+            # * When user never asked about ethtool configuration, network-role
+            #   will generate an all-default `NM.SettingEthtool` and pass it to
+            #   NetworkManager daemon. But NetworkManager discard it when saving to
+            #   ifcfg plugin as ifcfg plugin will not keep empty ethtool option.
+            # * The follow up `NM.SimpleConnection.compare()` will indicate
+            #   configuration changed as new ethtool setting found.
+            # * To workaround this, we just remove the all-default NM.SettingEthtool.
```
tyll, line 985: Here are some ideas to improve the wording. I am not 100% happy with my attempt, though.

```suggestion
            # * When users did not specify any ethtool configuration, this module
            #   will generate an default `NM.SettingEthtool` object and pass it to
            #   NetworkManager. But NetworkManager cannot serialize this when using the ifcfg
            #   plugin but treats this as no `NM.SettingEthtool` object.
            # * The following `NM.SimpleConnection.compare()` will therefore identify a
            #   difference in the configuration.
            # * To workaround this, remove the default NM.SettingEthtool object.
```

library/network_connections.py
```diff
@@ -1034,7 +1044,11 @@ def connection_create(self, connections, idx, connection_current=None):
                     s_ip4.add_dns(nameserver["address"])
             for search_domain in ip["dns_search"]:
                 s_ip4.add_dns_search(search_domain)
-            s_ip4.clear_dns_options(True)
+            # NetworkManager ifcfg plguin will discard empty dns option which
+            # cause follow up NM.Connection.compare() raise false alarm
+            # Use False here to ask NetworkManager remove dns option completely instead
+            # of keeping an empty list
+            s_ip4.clear_dns_options(False)
```
tyll, line 1051: It seems that False is the default so maybe remove the argument and forget that it exists. I don't really understand the difference between True and False for this argument to be honest.

tyll (approved):
:+1:

tyll (approved):
:+1:

tyll:
[citest bad]

tyll:
[citest bad]

library/network_connections.py
cathay4t: True: use [] as dns option.
False: Remove the dns option entry from setting.

tests/tests_change_indication_on_repeat_run.yml
cathay4t: Thanks for the suggestion. The code is working now.

cathay4t:
[citest bad]

cathay4t:
The CI failure is unrelated. Force merging.

PR #399, opened 2021-07-14
test: Fix the bond test on DHCP

cathay4t:
The `tests_bond_nm.yml` test fails at 50% rate due to bond interface
failed to retrieve IP address from DHCP server.

The root cause of this is NetworkManager by default create default
connections for new managed interface which place the DHCP server
interfaces into IPv4.auto method which then prevent DHCP server running
on these ports.

The fix is mark DHCP server ports(veth endpoint and bridge) as
unmanaged before link up.

### DIFF
--- a/tests/tasks/create_test_interfaces_with_dhcp.yml
+++ b/tests/tasks/create_test_interfaces_with_dhcp.yml
@@ -8,21 +8,24 @@
 
 - name: Create test interfaces
   shell: |
-    # NM to see veth devices starting with test* as managed after ip add..
-    echo 'ENV{ID_NET_DRIVER}=="veth",\
-          ENV{INTERFACE}=="test*", \
-          ENV{NM_UNMANAGED}="0"' >/etc/udev/rules.d/88-veth.rules
-    udevadm control --reload-rules
-    udevadm settle --timeout=5
-
-    # Setuptwo devices with IPv4/IPv6 auto support
     ip link add {{ dhcp_interface1 }} type veth peer name {{ dhcp_interface1 }}p
-    ip link set {{ dhcp_interface1 }}p up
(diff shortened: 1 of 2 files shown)


tests/tasks/create_test_interfaces_with_dhcp.yml
```diff
@@ -9,20 +9,23 @@
 - name: Create test interfaces
   shell: |
     # NM to see veth devices starting with test* as managed after ip add..
```
tyll, line 11: please also remove the comment when you remove the code

tests/tasks/create_test_interfaces_with_dhcp.yml
```diff
@@ -9,20 +9,23 @@
...
-    udevadm control --reload-rules
-    udevadm settle --timeout=5
-
     # Setuptwo devices with IPv4/IPv6 auto support
     ip link add {{ dhcp_interface1 }} type veth peer name {{ dhcp_interface1 }}p
-    ip link set {{ dhcp_interface1 }}p up
     ip link add {{ dhcp_interface2 }} type veth peer name {{ dhcp_interface2 }}p
+    if [ -x /usr/bin/nmcli ];then
+      nmcli d set {{ dhcp_interface1 }} managed true
+      nmcli d set {{ dhcp_interface2 }} managed true
```
tyll, line 17: is this correct in case we are testing the initscripts backend?

cathay4t: As long as NetworkManager daemon is running, we should do that.

Let me change it to check NM daemon status.

richm:
needs rebase

cathay4t:
Closing as patch has been included in PR #408 showing I actually fixed the bond tests.

PR #400, opened 2021-07-14
test: Restore the resolv.conf for bond dhcp test

cathay4t:
The `tests_bond_initscripts.yml` will leave the VM holding the testing
DNS entry (nameserver 192.0.2.1) which break network connection.

The root cause is initscript will use dhcp-client for bond DHCP testing
which override the /etc/resolv.conf file. After test bond interface been
removed, the initscript will not restore the DNS settings.

The fix is backup the /etc/resolv.conf and restore it on clean up for
initscript provider.

### DIFF
--- a/tests/playbooks/tests_bond.yml
+++ b/tests/playbooks/tests_bond.yml
@@ -19,6 +19,10 @@
     - import_tasks: tasks/assert_device_present.yml
       vars:
         interface: "{{ dhcp_interface2 }}"
+    - name: "Backup the /etc/resolv.conf for initscript"
+      command: cp -vf /etc/resolv.conf /etc/resolv.conf.bak
+      when:
+        - network_provider == "initscripts"
     - block:
         - name: "TEST Add Bond with 2 ports"
           debug:
@@ -93,5 +97,9 @@
             - command: ip link del {{ controller_device }}
               ignore_errors: true
             - import_tasks: tasks/remove_test_interfaces_with_dhcp.yml
+            - name: "Restore the /etc/resolv.conf for initscript"
+              command: mv -vf /etc/resolv.conf.bak /etc/resolv.conf
+              when:
+                - network_provider == "initscripts"
           tags:
             - "tests::cleanup"


tests/playbooks/tests_bond.yml
```diff
@@ -19,6 +19,10 @@
     - import_tasks: tasks/assert_device_present.yml
       vars:
         interface: "{{ dhcp_interface2 }}"
+    - name: "Backup the /etc/resolv.conf for initscript"
+      command: cp -vf /etc/resolv.conf /etc/resolv.conf.bak
+      when:
+        - network_provider == "initscripts"
```
richm, line 25: I think this task needs a `changed_when: false` but I'm not sure why it wasn't flagged by ansible-lint

tests/playbooks/tests_bond.yml
```diff
@@ -93,5 +97,9 @@
             - command: ip link del {{ controller_device }}
               ignore_errors: true
             - import_tasks: tasks/remove_test_interfaces_with_dhcp.yml
+            - name: "Restore the /etc/resolv.conf for initscript"
+              command: mv -vf /etc/resolv.conf.bak /etc/resolv.conf
+              when:
+                - network_provider == "initscripts"
```
richm, line 103: I think this task needs a `changed_when: false` but I'm not sure why it wasn't flagged by ansible-lint

richm:
[citest bad]

richm:
needs rebase

tests/playbooks/tests_bond.yml
cathay4t: Why `changed_when` is required here?

richm: I guess it doesn't matter since this is in test code, and so will always report `changed` for the task, and will never be run again, so it won't run afoul of idempotency.

tests/playbooks/tests_bond.yml
richm: I guess it doesn't matter since this is in test code, and so will always report `changed` for the task, and will never be run again, so it won't run afoul of idempotency.

richm (approved):
lgtm

PR #402, opened 2021-07-15
fix: pytest not reproducible in RHEL8

liangwen12year:
Previously when we run `tests_integration_pytest.yml` in RHEL 8, device
becoming unmanaged and NM `ifup` command failed to bring up the
connection when running pytest with initscripts, because network scripts
are deprecated in RHEL8 and they are no longer provided by default.

To fix that, install `network-scripts` package to utilize the legacy
`ifup` command.

### DIFF
--- a/tests/playbooks/integration_pytest_python3.yml
+++ b/tests/playbooks/integration_pytest_python3.yml
@@ -131,6 +131,10 @@
                 var: playbook_run.stdout_lines
 
         - block:
+            - name: install network-scripts when running pytest with initscripts
+              package:
+                name: network-scripts
+                state: present
             - name: Run pytest with initscripts
               command: >
                 pytest


tests/integration/test_ethernet.py
```diff
@@ -106,6 +106,8 @@ def test_static_ip_with_ethernet(testnic1, provider, network_lsr_nm_mock):
             "ip": {"address": [ip_address]},
         }
     ]
+    if provider == "initscripts":
+        subprocess.check_call(["dnf", "-y", "install", "network-scripts"])
```
richm, line 110: I think this should be done at the ".yml" level - not sure where - in one of these places:
- tests/tests_integration_pytest.yml
- tests/playbooks/integration_pytest_python3.yml
- tests/tests_default_initscripts.yml

That way you can easily adjust the name of the package to install based on the platform, and use the `package:` module

liangwen12year: Sounds good, updated.

cathay4t (approved):

richm:
lgtm - test passes on centos-8 and rhel-9 (when run manually)

richm:
[citest bad]

richm:
[citest bad]

PR #403, opened 2021-07-15
[th/arg-validator-default-value] cleanup and minor fixes for default-value handling of ArgValidator 

thom311:


cathay4t:
@thom311 I see 11 patch with zero comments except title.
Could you provide simple words about what this PR try to solve or what benefit will this PR bring?

thom311:
what should I do to move this forward?

tyll:
> what should I do to move this forward?

The CI for tox/Python 2.7 needs to be fixed at least.

module_utils/network_lsr/argument_validator.py
```diff
@@ -457,6 +457,11 @@ def __init__(self, name, family=None, required=False):
         self.family = family
 
     def _validate_impl(self, value, name):
+        # Note that ArgValidatorIPAddr accepts both a string and
+        # a dictionary (with "address" and "prefix") as input.
+        #
+        # But the validation will always unfold this into a dictionary
+        # containing "address", "family", ad "prefix".
```
tyll, line 476: Since it is a description of the class, it seems like it would fit better in a docstring of the class. What do you think?

module_utils/network_lsr/argument_validator.py
```diff
@@ -92,12 +92,12 @@ class ArgValidator:
     def __init__(self, name=None, required=False, default_value=None):
         self.name = name
         self.required = required
-        self.default_value = default_value
+        self._default_value = default_value
 
     def get_default_value(self):
```
tyll, line 97: Since you are changing this, it seems to be more pythonic to me to make `default_value` a property, then.

module_utils/network_lsr/argument_validator.py
```diff
@@ -347,7 +347,7 @@ def _validate_impl(self, value, name):
                 raise ValidationError(name, "invalid key '%s'" % (setting))
             if isinstance(validator, ArgValidatorDeprecated):
                 setting = validator.deprecated_by
-                validator = self.nested.get(setting, None)
```
tyll, line 355: There is something wrong in the commit message: `"validator.deprecated_by" is not use input.` not sure what you mean exactly.

Also, it would be nice to keep the summary short enough that it is not wrapped.

module_utils/network_lsr/argument_validator.py
```diff
@@ -311,11 +311,17 @@ def _validate_impl(self, value, name):
...
+class ArgValidatorDeprecated(ArgValidator):
     def __init__(self, name, deprecated_by):
-        self.name = name
+        ArgValidator.__init__(self, name, default_value=ArgValidator.MISSING)
         self.deprecated_by = deprecated_by
 
+    def _validate_impl(self, value, name):
+        # The Deprecated validator is not a full fledged validator that can
+        # also validate settings. All it does is a marker to rename/alias
+        # the validator with another one.
```
tyll, line 322: This is a description of the class and therefore should be in a docstring for the class in my opinion, what do you think?

module_utils/network_lsr/argument_validator.py
```diff
@@ -364,9 +364,9 @@ def _validate_impl(self, value, name):
             if validator.required:
                 raise ValidationError(name, "missing required key '%s'" % (setting))
             if not self.all_missing_during_validate:
-                default_value = validator.get_default_value()
-                if default_value is not ArgValidator.MISSING:
-                    result[setting] = default_value
+                v = validator.get_default_value()
+                if v is not ArgValidator.MISSING:
+                    result[setting] = v
```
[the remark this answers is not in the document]
tyll: Especially in this case using "v" for "default_value" when there is also a validator is messing with my brain since it is assuming that "v" means "validator" here after this change. If you dislike "default_value" here, then "default" might be a good alternative.

module_utils/network_lsr/argument_validator.py
thom311: That `@property` makes attributes visually indistinguishable from a method call, seems undesirable.

Granted, in python the distinction between what looks like an access to an attribute, versus an underlying function call is probably subtle. Important is that `self.get_default_value()` makes it more obvious that something gets computed and that it returns a new value. In fact, it's important that if this function returns either an immutable value or a clone that is owned by the user (that is, `self._default_value = list()` would be wrong, it should be `self._default_value = list`).

Also, again, the term "default_value" is for different purposes. I want to be able to search the source for the callers of this function using a unique name.

module_utils/network_lsr/argument_validator.py
thom311: There could be a class docstring. This comment was not intended to be a documentation of the class. It's a note about some special behavior that happens right here, during `validate()`. I reader of the code might be interested, how `validate()` behaves, that's why the note is there. For that aspect, I wouldn't want to look at the class docstring.

Also the docstring of the class should describe that this is in general a validator for an IP address, and the forms in which it accepts the input and normalizes the output. I am not strongly against such a documentation (if somebody wants to add it), but I'd rather read the code (that is hopefully easy to read and has a note/comment about the pitfalls/specialties), than a prosaic description about it.

By that argument this entire commit shouldn't be there. I'll drop it.

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
thom311: There are other short variables like `k`, `v`, `kk`. As you might guess, I don't have the distaste for them, if they are  temporary variable, that is used in a very narrow context (spanning very few lines of code).

I would prefer `val`, as `default` sounds like a reserved keyword (which it isn't, in Python). But I don't really mind. Renamed to `default`!!

module_utils/network_lsr/argument_validator.py
thom311: my problem with this (as before), is that it's hard to write a good, complete docstring for the class. I also find that many docstrings are redundant and line-noise, and the human would better read the code instead. So, when I would set up to write one, it rather be good.

The requirements for a good code comment are smaller, but also they can more specifically highlight a certain, immediate point. Instead of the much larger goal to document the entire class.

So, when making suggestions how to improve the documentation, please be very specific (or provide a patch), because I lack the imagination for how a good class docstring could be.

module_utils/network_lsr/argument_validator.py
```diff
@@ -316,26 +311,35 @@ def _validate_impl(self, value, name):
         raise ValidationError(name, "must be an boolean but is '%s'" % (value))
 
 
-class ArgValidatorDeprecated:
+class ArgValidatorDeprecated(ArgValidator):
```
tyll, line 314:
```suggestion
class ArgValidatorDeprecated(ArgValidator):
"""

ArgValidatorDeprecated is only used as a marker to indicate that a setting is deprecated by another setting. The validator that contains a deprecated setting is responsible for processing this and the replacement setting needs to perform the validation.

"""
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -316,26 +311,35 @@ def _validate_impl(self, value, name):
...
     def __init__(self, name, deprecated_by):
-        self.name = name
+        ArgValidator.__init__(self, name, default_value=ArgValidator.MISSING)
         self.deprecated_by = deprecated_by
 
+    def _validate_impl(self, value, name):
+        # The Deprecated validator is not a full fledged validator that can
+        # also validate settings. All it does is a marker to rename/alias
+        # the validator with another one.
+        raise MyError("Calling validate on a deprecated setting is unexpected")
```
tyll, line 323:
```suggestion
        raise MyError("Deprecated settings need to be validated by the replacement setting.")
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -347,7 +347,7 @@ def _validate_impl(self, value, name):
                 raise ValidationError(name, "invalid key '%s'" % (setting))
             if isinstance(validator, ArgValidatorDeprecated):
                 setting = validator.deprecated_by
-                validator = self.nested.get(setting, None)
+                validator = self.nested[setting]
```
liangwen12year, line 364: I noticed that the line above it also raised a error,  should we also raise an error here ? 
```python
            try:
                validator = self.nested[setting]
            except KeyError:
                raise ValidationError(name, "invalid key '%s'" % (setting))
```

thom311: no. If `self.nested[setting]` would access a non-existing key, then that is not a validation error. Instead, it would be a user error of how the `ArgValidator` instance was set up.

Certainly not a `ValidationError`, because it's not an error validating the input. Maybe a more suitable exception would be useful, but why? If this happens, it's a bug in the program. That's really bad already, and the only solution for that is fixing the bug, not raising a better exception. The `KeyError` we'd get here is find, it will lead to a crash, we'll see it, and will then fix the bug.

PR #404, opened 2021-07-15
tests: allow executing `pytest test_network_connections.py` directly

thom311:
Unit tests are important for development, and the ability to run them
directly, without tox etc.

Fake the environment so that `test_network_connections.py` can be run
directly via
```
  $ pytest tests/unit/test_network_connections.py -v
```

### DIFF
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import copy
 import itertools
+import os
 import pprint as pprint_
 import socket
 import sys
@@ -15,9 +16,12 @@ except ImportError:  # py2
 
 sys.modules["ansible.module_utils.basic"] = mock.Mock()
 
-# pylint: disable=import-error, wrong-import-position
+try:
+    import network_lsr
+except ImportError:
+    sys.path.append(os.path.join(os.path.dirname(__file__), "fake_env"))
+    import network_lsr
 
-import network_lsr
 import network_lsr.argument_validator
 from network_connections import IfcfgUtil, NMUtil, SysUtil, Util
 from network_lsr.argument_validator import ValidationError
--- a/tests/unit/fake_env/ansible/module_utils/network_lsr
+++ b/tests/unit/fake_env/ansible/module_utils/network_lsr
@@ -1,0 +1,1 @@
+../../../../module_utils/network_lsr
--- a/tests/unit/fake_env/network_connections.py
+++ b/tests/unit/fake_env/network_connections.py
@@ -1,0 +1,1 @@
+../../../library/network_connections.py
--- a/tests/unit/fake_env/network_lsr
+++ b/tests/unit/fake_env/network_lsr
@@ -1,0 +1,1 @@
+../../../module_utils/network_lsr


tests/unit/test_network_connections.py
```diff
@@ -15,7 +16,10 @@
 
 sys.modules["ansible.module_utils.basic"] = mock.Mock()
 
-# pylint: disable=import-error, wrong-import-position
+try:
+    import network_lsr
+except ModuleNotFoundError:
```
richm, line 21: looks like this breaks py27 support - `ModuleNotFoundError` not supported - why not use `ImportError`?

thom311:
`pytest tests/unit/test_nm_provider.py` already was working, but it does not require the symlinks in `fake_env`.

Hm. I wonder whether this patch could also function without those fake_env symlinks. Possibly, but I couldn't get it to work. Suggestion welcome.

tests/unit/test_network_connections.py
thom311: done.

tyll:
https://github.com/linux-system-roles/network/commit/9c86ff6f767ee83e09dc9260c8f75b144c8ec1b4#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359 removed a similar workaround that setup the sys.path to the parent directories but I think that broke because of the change to collections so this might break here, too.

thom311:
> [9c86ff6#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359](https://github.com/linux-system-roles/network/commit/9c86ff6f767ee83e09dc9260c8f75b144c8ec1b4#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359) removed a similar workaround that setup the sys.path to the parent directories but I think that broke because of the change to collections so this might break here, too.

I don't understand what "collections" are. Anyway, the patch still works with `tox -e py38` but also with `pytest` directly. What do you mean "so this might break here too".

I don't mind about a particular solution. What solution do you suggest instead?

richm:
> > [9c86ff6#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359](https://github.com/linux-system-roles/network/commit/9c86ff6f767ee83e09dc9260c8f75b144c8ec1b4#diff-b06f1e1bf991f123e64cf9dde6dcf9574bb279a720609a570e8af27cc2d9a359) removed a similar workaround that setup the sys.path to the parent directories but I think that broke because of the change to collections so this might break here, too.
> 
> I don't understand what "collections" are.

It is the new packaging format that Ansible is using going forward.  The "legacy" roles format (i.e. the format of this repo) will be deprecated at some point.  We must support collections in some use cases (e.g. Automation Hub). Though it may seem to be an unnecessary, useless burden to impose on system roles developers, we must support both the legacy roles format and the new collections format.  https://docs.ansible.com/ansible/latest/user_guide/collections_using.html

In fact, I can appreciate the fact that there is a lot of seemingly unnecessary "stuff" around system roles development - seemingly trivial things like flake8, black, ansible-lint, etc. that cause errors for particularly minor infractions - but some of these are imposed upon us by various Ansible gatekeepers e.g. Galaxy import, Automation Hub import, and so on - so we have to play by their rules.

> Anyway, the patch still works with `tox -e py38` but also with `pytest` directly. What do you mean "so this might break here too".

Not sure, but it was pretty tricky to make pytest work in tox *and* in the test playbook, *and* on all supported platforms, *and* in both the legacy roles format and the collections format.

> 
> I don't mind about a particular solution. What solution do you suggest instead?

Let's see what happens in the integration CI test runs first.

thom311:
@richm squashed the flake8 fix

tests/unit/test_network_connections.py
```diff
@@ -15,9 +16,12 @@
 
 sys.modules["ansible.module_utils.basic"] = mock.Mock()
 
-# pylint: disable=import-error, wrong-import-position
+try:
+    import network_lsr
+except ImportError:
+    sys.path.append(os.path.join(os.path.dirname(__file__), "fake_env"))
```
cathay4t, line 22: Placing soft link in code folder is not a good choice to me.

Please create the link at runtime.

cathay4t (changes requested):

tests/unit/test_network_connections.py
thom311: Could you elaborate why is that not a good choice? This will naturally only run on Linux, and symlinks work there.

Writing files to the source tree while running a unit test does not seem a good choice either -- except in cases where the unit test needs to write files and has a special scratch directory for that.

cathay4t: 1. I have preference on clean code base. Having unnecessary soft link is not clean.
2. I used to face a loop softlink issue in other project, so I dislike softlink with `../../../../`.
3. Put it in runtime creation code could explains why this link is required.

thom311: I don't know how to create the symlinks at test time in a way that is not bad.

We can close this MR and not merge it.

thom311:
reported issue #411 instead.

PR #405, opened 2021-07-20
Disable initscripts tests in rhel-9

liangwen12year:
Since network-scripts package will be dropped in rhel-9, disable all
the initscripts tests in rhel-9.

Signed-off-by: Wen Liang <liangwen12year@gmail.com>

PR #407, opened 2021-07-27
Skip tests on RHEL9 that use hostapd

richm:
The tests expect there is a hostapd package for doing wireless testing,
and that this package comes from EPEL on EL platforms.  Since there
is no EPEL9, skip tests that require hostapd.

### DIFF
--- a/tests/ensure_provider_tests.py
+++ b/tests/ensure_provider_tests.py
@@ -57,8 +57,14 @@ RUN_PLAYBOOK_WITH_NM = """# SPDX-License-Identifier: BSD-3-Clause
 MINIMUM_VERSION = "minimum_version"
 EXTRA_RUN_CONDITION = "extra_run_condition"
 NM_ONLY_TESTS = {
-    "playbooks/tests_802_1x_updated.yml": {},
-    "playbooks/tests_802_1x.yml": {},
+    "playbooks/tests_802_1x_updated.yml": {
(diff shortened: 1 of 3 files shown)


tests/ensure_provider_tests.py
```diff
@@ -57,8 +57,12 @@
...
 EXTRA_RUN_CONDITION = "extra_run_condition"
 NM_ONLY_TESTS = {
-    "playbooks/tests_802_1x_updated.yml": {},
-    "playbooks/tests_802_1x.yml": {},
+    "playbooks/tests_802_1x_updated.yml": {
+        EXTRA_RUN_CONDITION: "ansible_distribution != 'RedHat' or\n      ansible_distribution_major_version | int < 9",
+    },
+    "playbooks/tests_802_1x.yml": {
+        EXTRA_RUN_CONDITION: "ansible_distribution != 'RedHat' or\n      ansible_distribution_major_version | int < 9",
+    },
```
liangwen12year, line 65: Maybe it is better to wrap the the line at 88 characters, see 
https://github.com/linux-system-roles/network/pull/405/commits/5b5a7d73724bef16d97631bb785176ff5bd97b71#diff-ad4940be7c7c31554d78a7255dab1bb445177228ef8acedefc688c04ed461975R124

other than that, LGTM.

richm: ok - how about this - https://github.com/linux-system-roles/network/commit/6c3bfc296d74272da9e0c58de35f2039f3e3e4f2

liangwen12year (approved):

PR #422, opened 2021-09-14
replace ini_file with yum/dnf

richm:
For ansible-core 2.11 compatability, replace the use of `ini_file`
with the yum/dnf config-manager command


spetrosi (approved):
lgtm

PR #423, opened 2021-09-16
Set DNS search setting only for enabled IP protocols 

liangwen12year:
When either ipv4 or ipv6 method is disabled, silently ignore dns search
setting for ipv4 or ipv6.

When ipv6 method is ignore, silently ignore dns search setting for ipv6.


library/network_connections.py
```diff
@@ -1098,8 +1107,12 @@ def connection_create(self, connections, idx, connection_current=None):
             for nameserver in ip["dns"]:
                 if nameserver["family"] == socket.AF_INET6:
                     s_ip6.add_dns(nameserver["address"])
-            for search_domain in ip["dns_search"]:
-                s_ip6.add_dns_search(search_domain)
+            # NetworkManager disallows to config `ipv6.dns-search` setting when
+            # ipv6.method is disabled or ignore. The role should silently ignore
+            # such setting.
```
thom311, line 1112: if the role *always* is supposed to ignore dns-search with IP method disabled, then the right place to perform such normalization is `ArgValidator`. That one is supposed to normalize the user input in a form, that can be used directly (with little additional workarounds) by the provider implementation.

On the other hand, if this is only a special behavior of the `nm` backend, then ArgValidator must not normalize settings away, which might be used by other backends. In that case, the patch is right, but the comment is wrong ("The role should silently ignore such setting" is wrong, it is only the "nm" provider)

tyll:
Please provide the test cases first so it is clear what behavior you expect to implement and to show that it works as expected.

library/network_connections.py
liangwen12year: I looked at the files associated with "DNS search", "ipv6 disabled" in `network-scripts` ( `grep -r 'DOMAIN\|disable_ipv6\|IPV6INIT' $(rpm -ql network-scripts)`), I am convinced that it is only the special behavior of nm backend. I updated the comments, thanks.

liangwen12year:
> Please provide the test cases first so it is clear what behavior you expect to implement and to show that it works as expected.

Test playbook was provided, thanks.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: br-example
+            type: bridge
+            ip:
+              dhcp4: false
+              auto6: false
+              dns: 8.8.8.8
+              dns_search: google.com
```
tyll, line 14: Please use something like `example.com` here instead of a domain name that is not reserved for testing.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+              address:
+                - 10.2.2.77/26
+            state: up
+    - block:
+        - name: Verify nmcli connection DNS search entry for IPv4
```
tyll, line 19: There is no verification happening here. Please name it something like `Get DNS search entry for IPv4`

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+          shell: |
+            set -euxo pipefail
```
tyll, line 21: why is this set command needed?

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+            nmcli -f ipv4.dns-search connection show br-example
+          register: ipv4_dns_search
+          ignore_errors: yes
+          changed_when: false
+
+        - name: Verify nmcli connection DNS search entry for IPv6
+          shell: |
+            set -euxo pipefail
```
tyll, line 29: Why is this set command needed?

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+    - block:
+        - name: Verify nmcli connection DNS search entry for IPv4
+          shell: |
+            set -euxo pipefail
+            nmcli -f ipv4.dns-search connection show br-example
+          register: ipv4_dns_search
+          ignore_errors: yes
+          changed_when: false
+
+        - name: Verify nmcli connection DNS search entry for IPv6
```
tyll, line 27: This is not a verification.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+                type: bridge
+                ip:
+                  dhcp4: false
+                  auto6: false
+                  dns:
+                    - 8.8.8.8
+                    - 2001:db8::20
+                  dns_search: google.com
+                  address:
+                    - 10.2.2.77/26
```
tyll, line 62: Please also test the case that only static IPv6 is configured.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+            nmcli -f ipv6.dns-search connection show br-example
+          register: ipv6_dns_search
+          ignore_errors: yes
+          changed_when: false
+
+        - name: Assert that ipv4 DNS search entry is configured correctly
+          assert:
+            that:
+              - "'google.com' in ipv4_dns_search.stdout"
+            msg: "ipv4 DNS search entry is configured incorrectly"
```
tyll, line 39: Please state the actual error instead of only calling it "incorrectly". For example use something like: "IPv4 DNS search entry is missing" or "ipv4.dns-search setting is missing".

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,93 @@
...
+              dns: 8.8.8.8
+              dns_search: google.com
+              address:
+                - 10.2.2.77/26
+            state: up
+    - block:
+        - name: Verify nmcli connection DNS search entry for IPv4
+          shell: |
+            set -euxo pipefail
+            nmcli -f ipv4.dns-search connection show br-example
```
tyll, line 22: Since these are technically integration tests and not functional tests, it might be nicer to check if /etc/resolv.conf contains "search example.com" or similar. I am not sure if this is feasible on systems that use systemd-resolved. Maybe resolvectl can be used, then. @ffmancera what do you think about this?

tyll:
What will happen if there is an IPv6 nameserver but no IPv6 IP specified in the playbook? Is this tested? What if there is an IPv4 nameserver but no IPv4 IP? What if there is a nameserver but no IP at all? These questions assume that there is also a dns_search specified.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: https://github.com/linux-system-roles/network/pull/342#issuecomment-780137206

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: Updated.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: Changed, thanks.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: I realized that there is no need to use the `shell` module, let's use `command` module instead.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: Removed `set -euxo pipefail` and used `command` module instead.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: Updated into "Get DNS search entry for IPv6". Thanks.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: Thanks for suggestion, "IPv4 DNS search entry is missing" sounds good to me.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: The test case was added.

liangwen12year:
> What will happen if there is an IPv6 nameserver but no IPv6 IP specified in the playbook? Is this tested? What if there is an IPv4 nameserver but no IPv4 IP? What if there is a nameserver but no IP at all? These questions assume that there is also a dns_search specified.

if there is an IPv4 nameserver but no IPv4 IP specified in the playbook, the role would reject such setting ("IPv4 needs to be enabled to support IPv4 nameservers."),  

but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),  

I discussed this with @thom311 , maybe the role should also reject such setting in argvalidator just like what we did for IPv4.disabled and IPv6.disabled (https://github.com/linux-system-roles/network/commit/93e509b533fa2c3ca4ff63abcce490f12999a5b6#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1860).

tyll:
> > What will happen if there is an IPv6 nameserver but no IPv6 IP specified in the playbook? Is this tested? What if there is an IPv4 nameserver but no IPv4 IP? What if there is a nameserver but no IP at all? These questions assume that there is also a dns_search specified.
> 
> if there is an IPv4 nameserver but no IPv4 IP specified in the playbook, the role would reject such setting ("IPv4 needs to be enabled to support IPv4 nameservers."),
> 
> but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),
> 
> I discussed this with @thom311 , maybe the role should also reject such setting in argvalidator just like what we did for IPv4.disabled and IPv6.disabled ([93e509b#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1860](https://github.com/linux-system-roles/network/commit/93e509b533fa2c3ca4ff63abcce490f12999a5b6#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1860)).

According to
https://github.com/linux-system-roles/network/commit/93e509b533fa2c3ca4ff63abcce490f12999a5b6#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1868 there is an exception that describes this error. Do you mean something else? Does this not work? Please add test cases for all the scenarios that we are discussing here if they are missing.

liangwen12year:
> > > What will happen if there is an IPv6 nameserver but no IPv6 IP specified in the playbook? Is this tested? What if there is an IPv4 nameserver but no IPv4 IP? What if there is a nameserver but no IP at all? These questions assume that there is also a dns_search specified.
> > 
> > 
> > if there is an IPv4 nameserver but no IPv4 IP specified in the playbook, the role would reject such setting ("IPv4 needs to be enabled to support IPv4 nameservers."),
> > but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),
> > I discussed this with @thom311 , maybe the role should also reject such setting in argvalidator just like what we did for IPv4.disabled and IPv6.disabled ([93e509b#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1860](https://github.com/linux-system-roles/network/commit/93e509b533fa2c3ca4ff63abcce490f12999a5b6#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1860)).
> 
> According to
> [93e509b#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1868](https://github.com/linux-system-roles/network/commit/93e509b533fa2c3ca4ff63abcce490f12999a5b6#diff-5ad0dbcf8c19fec31ef0a300ad4a4eaba5bfbf67d40e7f2c6a9960f86c4ca608R1868) there is an exception that describes this error. Do you mean something else? Does this not work? Please add test cases for all the scenarios that we are discussing here if they are missing.

Added the commits (test cases included)
     - reject dns search setting if IPv4 is disabled and IPv6 method is ignore
     - reject dns setting if IPv6 method is ignore

tests/unit/test_network_connections.py
```diff
@@ -3570,6 +3570,32 @@ def test_ipv6_dns_without_ipv6_config(self):
...
+        ]
+
+        self.assertRaisesRegexp(
+            ValidationError,
+            "IPv6 needs to be enabled to support IPv6 nameservers.",
+            validator.validate_connection_one,
+            "nm",
+            validator.validate(ipv6_dns_with_ipv6_method_ignore),
+            0,
+        )
```
liangwen12year, line 3597: I should add another test case of which the IPv6 method is manual, test and see if the error is raised again.

thom311:
Currently there are 3 patches:

- 44f04414b9ca ('reject dns setting if IPv6 method is ignore')
- eca8a775d8f8 ('reject dns search setting if IPv4 is disabled and IPv6 method is ignore')
- f9068b1afcf2 ('ignore ipv[46] dns search setting if corresponding ip method is disabled or ignore')

You either need the first or the two last. Not all of them. Either `ArgValidator` rejects a setting as invalid altogether, then you don't need to handle the case that an invalid setting might be passed to `connection_create()`. Or, you silently ignore the case, but then `ArgValidator` must not reject it.

liangwen12year:
> Currently there are 3 patches:
> 
> * [44f0441](https://github.com/linux-system-roles/network/commit/44f04414b9caa08ff8dfa18dd78cac57bc7fdde5) ('reject dns setting if IPv6 method is ignore')
> * [eca8a77](https://github.com/linux-system-roles/network/commit/eca8a775d8f884bcceca1ad828955146678840ac) ('reject dns search setting if IPv4 is disabled and IPv6 method is ignore')
> * [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) ('ignore ipv[46] dns search setting if corresponding ip method is disabled or ignore')
> 
> You either need the first or the two last. Not all of them. Either `ArgValidator` rejects a setting as invalid altogether, then you don't need to handle the case that an invalid setting might be passed to `connection_create()`. Or, you silently ignore the case, but then `ArgValidator` must not reject it.

If I only kept the first patch, how should I handle the following situation properly:

but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook (IPv6 method is ignore) ?
Currently, it is handled by libnm, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),

tyll:
> If I only kept the first patch, how should I handle the following situation properly:
> 
> but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook (IPv6 method is ignore) ?
> Currently, it is handled by libnm, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),

Can you please rephrase this, I don't understand the situation.

thom311:
> Currently, it is handled by libnm

the role must not configure something in libnm, for which libnm says this is invalid. Errors must be avoided *before* that happens.

How to handle this with respect to `ipv6.method=ignore` is the same question as with `ipv[46].method=disabled`. The answers are either the role silently ignores the conflicting setting (f9068b1) or `ArgValidator` rejects it as invalid.

tyll:
> You either need the first or the two last. Not all of them. Either `ArgValidator` rejects a setting as invalid altogether, then you don't need to handle the case that an invalid setting might be passed to `connection_create()`. Or, you silently ignore the case, but then `ArgValidator` must not reject it.

As far as I understand, there is nothing to "silently ignore". I would describe this that the role needs to properly translate the playbook into profiles and there are bugs that need to be fixed. Also, there are some checks missing in arg validator to reject unsupported playbooks. This PR is now addressing multiple problems, which make this less clear.

liangwen12year:
> > If I only kept the first patch, how should I handle the following situation properly:
> > but if there is an IPv6 nameserver but no IPv6 IP specified in the playbook (IPv6 method is ignore) ?
> > Currently, it is handled by libnm, libnm would reject such setting ("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'"),
> 
> Can you please rephrase this, I don't understand the situation.

When IPv6 method is ignore and DNS is defined, it is considered as invalid, 
if I only kept this patch, f9068b1 ('ignore ipv[46] dns search setting if corresponding ip method is disabled or ignore'), 
then the invalid setting would be rejected by libnm (("nm-connection-error-quark: ipv6.dns: this property is not allowed for 'method=ignore'") rather than the role itself

thom311:
> As far as I understand, there is nothing to "silently ignore"

With "Silently ignore" I mean that the user's playbook contains conflicting values (which the role cannot configure), and the role will ignore such configuration and do something what makes sense (silently). That is what f9068b1 does.

Yes, I agree that accepting inconsistent/invalid/conflicting settings should be flagged as an error to begin with. f9068b1 should not be done (nor should it be necessary as ArgValidator would catch the issue a step earlier).

tyll:
> > As far as I understand, there is nothing to "silently ignore"
> 
> With "Silently ignore" I mean that the user's playbook contains conflicting values (which the role cannot configure), and the role will ignore such configuration and do something what makes sense (silently). That is what [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) does.
> 
> Yes, I agree that accepting inconsistent/invalid/conflicting settings should be flagged as an error to begin with. [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) should not be done (nor should it be necessary as ArgValidator would catch the issue a step earlier).

Unfortunately, the commit message does not explain the problem as I understand it. Commit [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) fixes a bug in the role where it translates a valid playbook into an invalid profile. The problem happens in the translation of the DNS settings in the playbook (which are not separated by IP family) into the profile (separated by IP family). From my understanding, this cannot/should not be fixed in the arg validator. I believe, that the commit message and code comments can be improved to make this more clear. Also, I believe that the current structure of the integration tests make it hard to see the tested scenarios (which is a problem that some existing integration tests have, too).

tests/unit/test_network_connections.py
```diff
@@ -3570,6 +3570,32 @@ def test_ipv6_dns_without_ipv6_config(self):
             0,
         )
 
+    def test_ipv6_dns_when_ipv6_method_is_ignore(self):
+        """
+        Test that configuring IPv6 DNS is not allowed when IPv6 method is ignore.
```
tyll, line 3575: What is the difference in the test scenario compared to the previous test? "test_ipv6_dns_without_ipv6_config". Please do not use "IPv6 method is ignore" in this context, I believe I explained this in another comment (you might not have read it, yet).

tyll: Ah, I see. So the possible scenarios are the combination of ipv6_disabled:true|false x auto6:true|false x static_ipv6_address:true|false

So is this table correct? The last column is the result of the previous columns:

<table>
<tr>
	<td>ipv6_disabled
	<td>auto6
	<td>static ipv6 address
	<td>IPv6 nameserver support
<tr>
	<td>true
	<td>*
	<td>*
	<td>no
<tr>
	<td>false
	<td>false
	<td>false
	<td>no
<tr>
	<td>false
	<td>true
	<td>*
	<td>yes
<tr>
	<td>false
	<td>*
	<td>true
	<td>yes
</table>

liangwen12year: Thanks for making it so clear. Yes, the table is correct.

tyll: ok, so you called row 2 "disabled" so it might make sense to call the previous test  "test_ipv6_dns_with_ipv6_disabled" and this one "test_ipv6_dns_without_ipv6_configuration" (just a suggestion). 

Btw. I did not look into this but it would also be good to add test cases for the other scenarios from the table that are not specific to DNS, so for example ipv6_disabled:true and auto6:true should probably also raise a validation error (not sure if they do or don't). But this is not related to the DNS feature.

liangwen12year: FYI,  `ipv6_disabled:true` and `auto6:true` are mutually exclusive, this has already been implemented, https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L655

liangwen12year: I created a separate PR to validate the other scenarios from the table,
https://github.com/linux-system-roles/network/pull/425

liangwen12year:
> > > As far as I understand, there is nothing to "silently ignore"
> > 
> > 
> > With "Silently ignore" I mean that the user's playbook contains conflicting values (which the role cannot configure), and the role will ignore such configuration and do something what makes sense (silently). That is what [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) does.
> > Yes, I agree that accepting inconsistent/invalid/conflicting settings should be flagged as an error to begin with. [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) should not be done (nor should it be necessary as ArgValidator would catch the issue a step earlier).
> 
> Unfortunately, the commit message does not explain the problem as I understand it. Commit [f9068b1](https://github.com/linux-system-roles/network/commit/f9068b1afcf246806bb00ca0c720a9cde82dbd6d) fixes a bug in the role where it translates a valid playbook into an invalid profile. The problem happens in the translation of the DNS settings in the playbook (which are not separated by IP family) into the profile (separated by IP family). From my understanding, this cannot/should not be fixed in the arg validator. I believe, that the commit message and code comments can be improved to make this more clear. Also, I believe that the current structure of the integration tests make it hard to see the tested scenarios (which is a problem that some existing integration tests have, too).

I updated the commit msg and comments, as well as rewording the integration test playbooks.

library/network_connections.py
```diff
@@ -1052,8 +1054,15 @@ def connection_create(self, connections, idx, connection_current=None):
...
                 if nameserver["family"] == socket.AF_INET:
                     s_ip4.add_dns(nameserver["address"])
-            for search_domain in ip["dns_search"]:
-                s_ip4.add_dns_search(search_domain)
+            # DNS (the protocol) is irrelevant to the IP family. NetworkManager allows
+            # you to confiugre `ipv[46].dns-*` options (separated by IP family), it is
+            # a misdesign in the API. The role implemented it correctly, but when these
+            # options need to be set, it should configure both IPv4 and IPv6 setting in
+            # NetworkManager. And NetworkManager disallows to configure
+            # `ipv4.dns-search` setting when `ipv4.method` is disabled.
```
tyll, line 1062: This comment does not seem to belong, here, IMHO. It is already mentioned in the commit message. Also, it is too generic for the specific code.

library/network_connections.py
```diff
@@ -1098,8 +1111,15 @@ def connection_create(self, connections, idx, connection_current=None):
...
                     s_ip6.add_dns(nameserver["address"])
-            for search_domain in ip["dns_search"]:
-                s_ip6.add_dns_search(search_domain)
+            # DNS (the protocol) is irrelevant to the IP family. NetworkManager allows
+            # you to confiugre `ipv[46].dns-*` options (separated by IP family), it is
+            # a misdesign in the API. The role implemented it correctly, but when these
+            # options need to be set, it should configure both IPv4 and IPv6 setting in
+            # NetworkManager. And NetworkManager disallows to configure
+            # `ipv6.dns-search` setting when `ipv6.method` is disabled or ignore.
+            if not is_ipv6_disabled and not is_ipv6_ignore:
```
tyll, line 1120: One boolean variable seems to be enough, you can call it for example ipv6_available if you want to avoid confusion with disabled or ignore.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1796,6 +1796,17 @@ def _ipv4_enabled(connection):
             )
             return connection["ip"]["dhcp4"] or has_addrs4
```
tyll, line 1798: Proposal for the commit message summary: "Reject dns_search setting if no IPv4/IPv6 is disabled" - there is no need to distinguish between disabled/not configured, here, IMHO.

library/network_connections.py
```diff
@@ -1029,6 +1029,7 @@ def connection_create(self, connections, idx, connection_current=None):
             addrs4 = list([a for a in ip["address"] if a["family"] == socket.AF_INET])
             addrs6 = list([a for a in ip["address"] if a["family"] == socket.AF_INET6])
 
+            is_ipv4_disabled = False
```
tyll, line 1032: regarding the commit message: It is not refactoring, if you change the behavior, IMHO. The change is something like "Set DNS search setting only for enabled IP protocols", then the details can explain the reasons for this.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,123 @@
...
+      import_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: br-example
+            type: bridge
+            ip:
+              dhcp4: false
+              auto6: false
+              dns: 8.8.8.8
```
tyll, line 14: Please don't use production IP addresses in tests.

module_utils/network_lsr/argument_validator.py
liangwen12year: Thanks for the proposal, I updated.

tests/unit/test_network_connections.py
```diff
@@ -3597,6 +3597,28 @@ def test_ipv6_dns_without_ipv6_configuration(self):
...
+                    "dhcp4": False,
+                    "auto6": False,
+                    "dns": ["2001:db8::20"],
+                    "address": ["2001:db8::2/32"],
+                },
+            }
+        ]
+
+        validator.validate_connection_one(
+            "nm", validator.validate(ipv6_dns_with_static_ipv6_configuration), 0
```
tyll, line 3619: what is the `0` here? It will be easier to understand if you assign it to a variable (then it can have a name) and use the variable here.
Also, a test should have some assertion.

tests/playbooks/tests_ipv6_dns_search.yml
liangwen12year: I used the address blocks reserved for documentation.

library/network_connections.py
liangwen12year: Thanks for the note. I updated.

library/network_connections.py
liangwen12year: How about 
```
# NetworkManager only allows to configure `ipv4.dns-search` setting when
# IPv4 is enabled
```

liangwen12year: I changed into:
```
# NetworkManager only allows to configure `ipv4.dns-search` setting when
# IPv4 is enabled
```

library/network_connections.py
liangwen12year: Nice hint ! Updated.

tests/unit/test_network_connections.py
liangwen12year: I rewrited as following:
```python
        # @idx is the index for the all explicitly defined connection profiles, since
        # there is only one connection profile defined here, then idx is 0
        idx = 0

        # Python unit testing framework does not define the function
        # `assertNotRaises()`. As a workaround, validate that if
        # `ArgValidator_ListConnections().validate_connection_one()` will return None.
        # Because the function only returns None when it executes successfully without
        # raising errors.
        self.assertEqual(
            validator.validate_connection_one(
                "nm", validator.validate(ipv6_dns_with_static_ipv6_configuration), idx
            ),
            None,
        )
```

liangwen12year: I rewrited as following:
```python
        # @idx is the index for the all explicitly defined connection profiles, since
        # there is only one connection profile defined here, then idx is 0
        idx = 0

        # Python unit testing framework does not define the function
        # `assertNotRaises()`. As a workaround, validate that if
        # `ArgValidator_ListConnections().validate_connection_one()` will return None.
        # Because the function only returns None when it executes successfully without
        # raising errors.
        self.assertEqual(
            validator.validate_connection_one(
                "nm", validator.validate(ipv6_dns_with_static_ipv6_configuration), idx
            ),
            None,
        )
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -1964,6 +1975,12 @@ def _ipv4_enabled(connection):
                     "Setting 'dns_search' or 'dns_options' is not allowed when "
                     "both IPv4 and IPv6 are disabled.",
                 )
+            elif not _ipv4_enabled(connection) and _ipv6_ignore(connection):
```
thom311, line 1977: for better or worse, NetworkManager does [this](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/6b3862e39ac4e2dee19d6b87c117ea3a1125c9dd/src/libnm-core-impl/nm-setting-ip6-config.c#L226). Meaning, it rejects setting "dns-search" for methods "ignore", "link-local" and "disabled".

The check `_ipv6_ignore()` only checks for the first of those 3 cases (as the name of the function indicates -- whereas the role currently never sets `ipv6.method=link-local`).

module_utils/network_lsr/argument_validator.py
```diff
@@ -1958,9 +1958,8 @@ def _ipv6_ignore(connection):
                     idx,
                     "IPv4 needs to be enabled to support IPv4 nameservers.",
                 )
-            if (
-                nameserver["family"] == socket.AF_INET6
-                and connection["ip"]["ipv6_disabled"]
+            if nameserver["family"] == socket.AF_INET6 and (
+                connection["ip"]["ipv6_disabled"] or _ipv6_ignore(connection)
```
thom311, line 1962: All these changes are in `ArgValidator.validate_connection_one()`. That function is not intended for such checks.

We have `ArgValidator*.validate()` -- which in particular calls `ArgValidator*._validate_post()`, which is the place for consistency checks that only depend on the values of one the playbook. In particular, `ArgValidator_DictConnection.validate()` verifies (and normalizes) one profile individually, and `ArgValidator_ListConnections.validate()` verifies (and normalizes) the entire list of profiles as a whole.

I guess, `ArgValidator_ListConnections.validate_connection_one()` is a bad name. But that function performs some validation *afterwards*, in particular depending on the provider (`mode`). For example, the `initscripts` provider does not support team. So `ArgValidator*.validate()` (which is independent of the provider/mode) cannot tell whether the playbook is valid for "initscripts". That validation can only happen in a second step: `validate_connection_one()`.

Here all these validation checks are independent of the mode/provider.


I don't mind. You can leave it like this. But in the future, this code should move.

richm:
[citest pending]

liangwen12year:
I will split the commits in this PR into several different PRs.

tests/playbooks/tests_ipv6_dns_search.yml
```diff
@@ -0,0 +1,149 @@
...
+        - name: Assert that configuring DNS search setting is not allowed when
+            both IPv4 and IPv6 are disabled
+          assert:
+            that:
+              - __network_connections_result.stderr is search("Setting
+                'dns_search' or 'dns_options' is not allowed when both IPv4 and
+                IPv6 are disabled.")
+            msg: Configuring DNS search setting is allowed when both IPv4 and
+              IPv6 are disabled
+
```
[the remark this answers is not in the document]
tyll: Please update the message to properly state the situation, then :-)

tests/ensure_provider_tests.py
```diff
@@ -72,6 +72,7 @@
     "playbooks/tests_ipv6_disabled.yml": {
         EXTRA_RUN_CONDITION: "ansible_distribution_major_version == '8'",
     },
+    "playbooks/tests_ipv6_dns_search.yml": {},
```
tyll, line 75: Why is this an NM-only test? Is setting DNS search not supported in initscripts?

liangwen12year: DNS search (`DOMAIN`) is supported in initscripts. But the situation is: "The role only set the DNS search setting for enabled IP family",  unlike NM, the `DOMAIN` is not distinguished on the IP  family, and `tests_ipv6_dns_search.yml` primarily test the specical behavior (The role only set the DNS search setting for enabled IP family) for NM backend.

But it is also true that we can test the DNS search setting for initscripts provider.

tests/playbooks/tests_ipv6_dns_search.yml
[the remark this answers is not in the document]
liangwen12year: I updated the error message into `Reconfiguring network connection is not failed with the error "Setting 'dns_search' or 'dns_options' is not allowed when both IPv4 and IPv6 are disabled."`

But I am afraid that the error message is not explicit.

library/network_connections.py
```diff
@@ -1029,14 +1029,17 @@ def connection_create(self, connections, idx, connection_current=None):
...
                 s_ip4.set_property(NM.SETTING_IP_CONFIG_METHOD, "auto")
                 s_ip4.set_property(
                     NM.SETTING_IP_CONFIG_DHCP_SEND_HOSTNAME,
                     ip["dhcp4_send_hostname"] is not False,
                 )
+                is_ipv4_available = True
             elif addrs4:
                 s_ip4.set_property(NM.SETTING_IP_CONFIG_METHOD, "manual")
+                is_ipv4_available = True
             else:
```
ffmancera, line 1043: What do you think about `is_ipv4_configured`? The current variable seems confusing to me.

library/network_connections.py
```diff
@@ -1062,12 +1068,15 @@ def connection_create(self, connections, idx, connection_current=None):
...
 
+            is_ipv6_available = False
             if ip["ipv6_disabled"]:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "disabled")
             elif ip["auto6"]:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "auto")
+                is_ipv6_available = True
             elif addrs6:
                 s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "manual")
+                is_ipv6_available = True
```
ffmancera, line 1079: Same here, what about  `is_ipv6_configured`?

library/network_connections.py
```diff
@@ -1062,12 +1068,15 @@ def connection_create(self, connections, idx, connection_current=None):
             for option in ip["dns_options"]:
                 s_ip4.add_dns_option(option)
```
ffmancera, line 1069: I think this is missing a check on `is_ipv4_available`.. right?

ffmancera (changes requested):

library/network_connections.py
liangwen12year: For dns option, the following are valid configuration:
``` 
ipv6.method:                            ignore
ipv6.dns:                               --
ipv6.dns-search:                        --
ipv6.dns-options:                       debug
```

liangwen12year: but when ipv4.method is disabled, then the ipv4.dns-options can not be configured.

library/network_connections.py
liangwen12year: Updated. thanks.

library/network_connections.py
liangwen12year: Updated.

tests/playbooks/tests_ipv6_dns_search.yml
ffmancera: Yes I agree. Ideally ´/etc/resolv.conf´ should show the values configured. Thanks and sorry for the delay.

library/network_connections.py
ffmancera: I think they can be configured.. maybe not with nmcli because of a bug on in ´make_ip4_config()´ in NM ifcfg plugin.. But when using libnm I think it is possible which if a problem..  Am I missing something?

ffmancera:
Please, rebase it, thanks!

liangwen12year:
> Please, rebase it, thanks!

rebased.

ffmancera (dismissed):
LGTM!

ffmancera (approved):
LGTM!

PR #425, opened 2021-09-27
validate that ipv6_disabled is conflicting with other settings

liangwen12year:
When IPv6 is disabled, it is conflicting with enabling `auto6`
or configuring static IPv6 or configuring `gateway6` or
configuring `route_metric6`. Add unit tests to check if
this is properly handled in Argvalidator.


tyll (approved):
Thank you, LGTM!

PR #428, opened 2021-10-06
[Testing CI]fix: python object has no attribute 'assertRaisesRegexp'

liangwen12year:

### DIFF
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -55,6 +55,19 @@ def pprint(msg, obj):
         obj.dump()
 
 
+class Python26CompatTestCase(unittest.TestCase):
+    # pylint: disable=no-member
+    def assertRaisesRegex(self, exception, regex, *args, **kwargs):
+        if sys.version[:3] == "2.6":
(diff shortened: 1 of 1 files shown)


tests/unit/test_network_connections.py
```diff
@@ -472,12 +472,20 @@ def test_validate_disallow_none_in_list(self):
...
+                v.validate,
+                ["pci-0001:00:00.0", None],
+            )
+        else:
+            self.assertRaisesRegexp(
+                ValidationError,
+                "must be a string but is 'None'",
+                v.validate,
+                ["pci-0001:00:00.0", None],
+            )
```
tyll, line 488: 1) This needs to be a helper that can be used instead of using the if condition in each affected test case and duplicating the assertion.

2) Instead of checking the python version, the code should use introspection to check which method is available

3) This seems to be addressed by Python Six, so maybe you can use that one: https://six.readthedocs.io/

ffmancera:
[citest bad]

ffmancera:
[citest bad]

PR #429, opened 2021-10-06
add support for ansible-core 2.11 ansible-lint, ansible-test

richm:
Add support for using latest ansible-lint and ansible-test with
ansible-core 2.11.  There are a few new warnings that need to
be addressed or suppressed.


.travis/custom.sh
```diff
@@ -3,6 +3,7 @@
 
 set -e
 
+# shellcheck disable=SC1091
```
tyll, line 6:
```suggestion
# https://github.com/koalaman/shellcheck/wiki/SC1091
# shellcheck disable=SC1091
```

PR #430, opened 2021-10-12
test IPv6 DNS when static IPv6 is configured

liangwen12year:


liangwen12year:
> Why is this PR waiting for other PRs but #423 is waiting for a review? It seems to me that it should be the other way.

@ffmancera , feel free to merge it if you agree with it

PR #433, opened 2021-10-14
fix ensure_provider_tests.py; fix generated tests; add check to CI

richm:
Fix ensure_provider_tests.py so that the tests are correctly generated.
Because the generated tests have not been checked in a while, there was
some drift between the actual tests and what should have been generated,
so the new generated tests have also been included.
When `ensure_provider_tests.py` detects differences, it will now show
the diffs using a unified diff format.
Added a new tox testenv `ensure_provider_tests` - use
`tox -e ensure_provider_tests` to run - to pass arguments, use
`tox -e ensure_provider_tests -- generate` or
`tox -e ensure_provider_tests -- --show-diff`
Added `ensure_provider_tests` as one of the tox testenvs to run for
github tox CI.
Added `tests/git-pre-commit.sh` and `tests/git-post-commit.sh` which can be copied to your local `.git/hooks/pre-commit` and `.git/hooks/post-commit` so that `ensure_provider_tests.py generate` is always run when you make a local commit, to catch any files which should have been generated, or generated files which were changed directly.


richm:
[citest]

richm:
@tyll The question about the use of the `NM` prefix for this change - apparently it is only the `tests/tests_*_nm.yml` tests that got out of sync with respect to their source playbooks and the use of `el_repo_setup` - so the most minimal change I could do is to identify those tests and add them to `ensure_provider_tests.py`, and add the `el_repo_setup` *only* to those NM related tests.  If this is incorrect - if `el_repo_setup` should be added to *all* tests - including all initscripts tests and all NM tests - then I can make that change to `ensure_provider_tests.py` - but it will cause a much, much larger change to all of the `tests/tests_*.yml` playbooks - is that what you want?

richm:
> > @tyll The question about the use of the `NM` prefix for this change - apparently it is only the `tests/tests_*_nm.yml` tests that got out of sync with respect to their source playbooks and the use of `el_repo_setup` - so the most minimal change I could do is to identify those tests and add them to `ensure_provider_tests.py`, and add the `el_repo_setup` _only_ to those NM related tests. If this is incorrect - if `el_repo_setup` should be added to _all_ tests - including all initscripts tests and all NM tests - then I can make that change to `ensure_provider_tests.py` - but it will cause a much, much larger change to all of the `tests/tests_*.yml` playbooks - is that what you want?
> 
> Not sure what I am missing but I see the `el_repo_setup` in this initscripts playbook, too:
> 
> https://github.com/linux-system-roles/network/blob/main/tests/tests_bridge_initscripts.yml#L7
> 
> Just checked, and it seems to be always present:
> 
> https://github.com/linux-system-roles/network/blob/main/tests/ensure_provider_tests.py#L134
> 
> I don't know right now, why this is the case. But at least from my understanding, whether or not the epel repo is needed, should only depend on the tests not on the provider. I might be missing something, though.

I didn't know if it was desirable for all tests to include el_repo_setup.  This latest commit does that - all tests now have the same consistent el_repo_setup.  I guess we'll see if we get the same test results.

PR #435, opened 2021-10-19
WIP: Use relative import and enable pytest unit test

cathay4t:
Since Ansible 2.9 is the minimum supported version, we can use relative
import instead of prefixing with `ansible.module_utils.`.

Included `conftest.py` file for pytest to setup the test environment
before invoking unit test cases.

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -9,9 +9,8 @@ import posixpath
 import socket
 import re
 
-# pylint: disable=import-error, no-name-in-module
-from ansible.module_utils.network_lsr.myerror import MyError  # noqa:E501
-from ansible.module_utils.network_lsr.utils import Util  # noqa:E501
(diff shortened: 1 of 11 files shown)


tyll:
> `test_network_connections.py` has many failures unrelated to this patch."

What failures do you mean?

tyll:
> @tyll The tox CI has detailed errors. One example would be:
> 
> ```
> 2021-10-19T06:19:10.8615183Z E ansible.module_utils.network_lsr.argument_validator.ValidationError: 
> connection[0]: 802.1x authentication is not supported by initscripts. Configure 802.1x in 
> /etc/wpa_supplicant.conf if you need to use initscripts.
> ```
> 
> It seems this patch revealed old hidden issue of old unit test.

What is the issue of the old unit tests? According to https://github.com/linux-system-roles/network/blob/main/tests/unit/test_network_connections.py#L288 the tests should not run with initscripts.

PR #436, opened 2021-11-02
support ansible-core 2.12; ansible-plugin-scan; fix pylint errors; py39

richm:
Use the new tox-lsr that uses ansible-core 2.12.
Use ansible-plugin-scan in CI.
Fix pylint errors found by new ansible-test pylinter.
CI test with py39


nhosoi (approved):
lgtm

spetrosi (approved):
lgtm

.github/workflows/tox.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        pyver: ['2.7', '3.6', '3.7', '3.8']
+        pyver: ['2.7', '3.6', '3.7', '3.8', '3.9']
```
tyll, line 18: Does Python 3.7 or 3.8 still need to be supported? Fedora 33 is already using Python 3.9 and so is RHEL 9 according to https://developers.redhat.com/articles/2021/11/03/red-hat-enterprise-linux-9-beta-here

tyll: Maybe one of those versions still needs to be used since it is the platform python version for Ubuntu, but I suspect that at least one version can be dropped when moving to 3.9.

tyll: And reading further into the diff, it seems that 3.7 is not used.

.sanity-ansible-ignore-2.12.txt
```diff
@@ -0,0 +1,65 @@
+plugins/module_utils/network_lsr/nm/__init__.py empty-init!skip
+plugins/module_utils/network_lsr/nm/__init__.py import-2.6!skip
+plugins/module_utils/network_lsr/nm/__init__.py import-2.7!skip
+plugins/module_utils/network_lsr/nm/__init__.py import-3.5!skip
```
tyll, line 4: Why are there skips for Python 3.5?

library/network_connections.py
```diff
@@ -256,7 +256,7 @@ def ValueEscape(cls, value):
         if r.match(value):
             return value
 
-        if any([ord(c) < ord(" ") for c in value]):
```
tyll, line 259: are these changes tested on Python 2.6?

tyll:
There are unexpected problems in the CI output: "load-failure [Errno 2] No such file or directory: 'tests/tasks/tasks/show_interfaces.yml'", not sure why this is not failing the CI. It is an annotation for the GitHub checks.

library/network_connections.py
richm: > are these changes tested on Python 2.6?

I tested them on a centos-6 image using python 2.6

.github/workflows/tox.yml
richm: > Maybe one of those versions still needs to be used since it is the platform python version for Ubuntu, but I suspect that at least one version can be dropped when moving to 3.9.

What is the harm of keeping 3.7 or 3.8?

richm: > And reading further into the diff, it seems that 3.7 is not used.

Not sure what you mean?
`          37) toxenvs="${toxenvs},coveralls" ;;`

.sanity-ansible-ignore-2.12.txt
richm: Because otherwise ansible-test complains about it

richm:
> There are unexpected problems in the CI output: "load-failure [Errno 2] No such file or directory: 'tests/tasks/tasks/show_interfaces.yml'", not sure why this is not failing the CI. It is an annotation for the GitHub checks.

Right.  We could not suppress those error messages, but we could make it so that they get classified as warnings not errors:
.ansible-lint
```
warn_list:
- load-failure  # allow include_tasks with tasks/ directory
```

.sanity-ansible-ignore-2.12.txt
richm: for example:
```
Running sanity test "import" on Python 3.5
ERROR: Found 6 import issue(s) on python 3.5 which need to be resolved:
ERROR: plugins/module_utils/network_lsr/nm/__init__.py:7:0: traceback: ImportError: No module named 'gi' (at plugins/module_utils/network_lsr/nm/client.py:13:0)
ERROR: plugins/module_utils/network_lsr/nm/active_connection.py:13:0: traceback: ImportError: No module named 'gi' (at plugins/module_utils/network_lsr/nm/client.py:13:0)
...
```

library/network_connections.py
richm: This is the error from ansible-test:
```
Running sanity test "pylint"
ERROR: Found 8 pylint issue(s) which need to be resolved:
ERROR: plugins/module_utils/network_lsr/argument_validator.py:667:29: use-a-generator: Use a generator instead 'any(a for a in result['address'] if a['family'] == socket.AF_INET6)'
ERROR: plugins/module_utils/network_lsr/argument_validator.py:698:79: use-a-generator: Use a generator instead 'any(a for a in result['address'] if a['family'] == socket.AF_INET)'
ERROR: plugins/module_utils/network_lsr/utils.py:223:11: use-a-generator: Use a generator instead 'any(c == '/' or c == ':' or c.isspace() for c in ifname)'
ERROR: plugins/modules/network_connections.py:259:11: use-a-generator: Use a generator instead 'any(ord(c) < ord(' ') for c in value)'
ERROR: plugins/modules/network_connections.py:473:21: consider-using-generator: Consider using a generator instead 'list(a for a in ip['address'] if a['family'] == socket.AF_INET)'
ERROR: plugins/modules/network_connections.py:474:21: consider-using-generator: Consider using a generator instead 'list(a for a in ip['address'] if a['family'] == socket.AF_INET6)'
ERROR: plugins/modules/network_connections.py:1029:21: consider-using-generator: Consider using a generator instead 'list(a for a in ip['address'] if a['family'] == socket.AF_INET)'
ERROR: plugins/modules/network_connections.py:1030:21: consider-using-generator: Consider using a generator instead 'list(a for a in ip['address'] if a['family'] == socket.AF_INET6)'
```

richm: Although for the `consider-using-generator` errors I think using a list comprehension is more pythonic than using `list(a generator expression)`

richm:
> > There are unexpected problems in the CI output: "load-failure [Errno 2] No such file or directory: 'tests/tasks/tasks/show_interfaces.yml'", not sure why this is not failing the CI. It is an annotation for the GitHub checks.
> 
> Right. We could not suppress those error messages, but we could make it so that they get classified as warnings not errors: .ansible-lint
> 
> ```
> warn_list:
> - load-failure  # allow include_tasks with tasks/ directory
> ```

. . . because `load-failure` has no effect if added to the `skip_list`

.github/workflows/tox.yml
tyll: > > Maybe one of those versions still needs to be used since it is the platform python version for Ubuntu, but I suspect that at least one version can be dropped when moving to 3.9.
> 
> What is the harm of keeping 3.7 or 3.8?

It uses unnecessary resources and adds more noise.

> > And reading further into the diff, it seems that 3.7 is not used.
> 
> Not sure what you mean? ` 37) toxenvs="${toxenvs},coveralls" ;;`

I meant not used for anything specific other than the unit tests. So if it was needed for ansible-lint, then it would make sense to keep it.

.sanity-ansible-ignore-2.12.txt
tyll: Is it possible to tell ansible-lint the target python versions instead?

richm: > Is it possible to tell ansible-lint the target python versions instead?

The problem is that we don't want to use the ansible-test import/compile checking, since we already have unit tests for that.  It would be quite a bit of work to make ansible-test import and compile checks pass, and it isn't really necessary.

richm: Note that ansible-test is really designed for checking *collections*, and primarily for checking/testing collections that provide *plugins* which are used directly by end users.  System roles does not provide plugins that are intended to be used directly by end users - our modules are intended to be used only within the context of the enclosing role, and the role ensures that the proper dependencies are in place in order to run the python modules, and we test that in unit tests and/or integration tests.  Therefore, supporting ansible-test import and compile checks are more trouble than they're worth, so we skip them.

.github/workflows/tox.yml
richm: fixed in https://github.com/linux-system-roles/network/pull/436/commits/59ed55cceb7b3d136d578506cdec4ad3978987e0

richm:
ok to merge?

richm:
If there are no further objections I will merge this.

tyll (approved):
LGTM

PR #438, opened 2021-11-18
update tox-lsr version to 2.8.0

spetrosi:
New version adds check for proper commenting of the ansible_managed var


richm (dismissed):
lgtm

PR #440, opened 2021-12-02
Reject DNS_search setting if no IP family is enabled

liangwen12year:
NetworkManager only allows to configure the DNS options for each IP
family, when they are enabled. Therefore, reject DNS settings in
ArgValidator if no IP family is enabled.

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -1974,6 +1974,12 @@ class ArgValidator_ListConnections(ArgValidatorList):
                     "Setting 'dns_search' or 'dns_options' is not allowed when "
                     "both IPv4 and IPv6 are disabled.",
                 )
+            elif not _ipv4_enabled(connection) and _ipv6_is_not_configured(connection):
+                raise ValidationError.from_connection(
+                    idx,
+                    "Setting 'dns_search' or 'dns_options' is not allowed when "
+                    "IPv4 is disabled and IPv6 is not configured.",
+                )
         # DNS options 'inet6', 'ip6-bytestring', 'ip6-dotint', 'no-ip6-dotint' are only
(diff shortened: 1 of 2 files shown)


module_utils/network_lsr/argument_validator.py
```diff
@@ -1974,6 +1974,12 @@ def _ipv6_is_not_configured(connection):
                     "Setting 'dns_search' or 'dns_options' is not allowed when "
                     "both IPv4 and IPv6 are disabled.",
                 )
+            elif not _ipv4_enabled(connection) and _ipv6_is_not_configured(connection):
```
ffmancera, line 1977: What is the difference between enabled and configured? Can be IPv6 enabled but not configured?

ffmancera (changes requested):

module_utils/network_lsr/argument_validator.py
liangwen12year: > What is the difference between enabled and configured?

In the current network role, if using nm provider as the backend, the role will configure ipv4.method as 'auto' or 'manual' or 'disabled'.  `not _ipv4_enabled` corresponds to the ipv4.method 'disabled'.

As a comparison, the role will configure ipv6.method as  'auto' or 'manual' or 'disabled' or 'ignore'. `_ipv6_is_not_configured` corresponds to the ipv6.method 'ignore'.
 



> Can be IPv6 enabled but not configured? 

No. You can take a look at the code here.  https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L703

ffmancera: This make sense, thanks for explaining!

ffmancera (approved):
LGTM! Thanks!

PR #443, opened 2021-12-21
Reload the connection before managing the device

liangwen12year:

### DIFF
--- a/tests/tasks/manage_test_interface.yml
+++ b/tests/tasks/manage_test_interface.yml
@@ -27,7 +27,9 @@
   when: "type == 'veth' and state == 'present' and
          interface not in current_interfaces"
 - name: Set up veth as managed by NetworkManager
-  command: nmcli d set {{ interface }} managed true
+  shell: |
+    nmcli c reload
+    nmcli d set {{ interface }} managed true
   # The varible for `network_provider` is not exists yet,
   # just ignore error for initscripts
   ignore_errors: yes  # noqa ignore-errors
--- a/tests/tasks/manage_test_interface.yml
+++ b/tests/tasks/manage_test_interface.yml
@@ -26,6 +26,23 @@
     - ip link set {{ interface }} up
   when: "type == 'veth' and state == 'present' and
          interface not in current_interfaces"
+
+- name: Install NetworkManager in CentOS7 when running with nm provider
+  package:
+    name: NetworkManager
+    state: present
+  when:
+    - ansible_distribution_major_version == '7' and
+      ansible_distribution == 'CentOS'
+    - network_provider == 'nm'
+
+- name: Start NetworkManager in CentOS7 when running with nm provider
+  command: systemctl start NetworkManager
+  when:
+    - ansible_distribution_major_version == '7' and
+      ansible_distribution == 'CentOS'
+    - network_provider == 'nm'
+
 - name: Set up veth as managed by NetworkManager
   shell: |
     nmcli c reload


liangwen12year:
I will add commit msg soon.

nhosoi (dismissed):
~~lgtm~~

tests/tasks/manage_test_interface.yml
```diff
@@ -27,7 +27,9 @@
   when: "type == 'veth' and state == 'present' and
          interface not in current_interfaces"
 - name: Set up veth as managed by NetworkManager
-  command: nmcli d set {{ interface }} managed true
+  command: |
+    nmcli c reload
+    nmcli d set {{ interface }} managed true
```
richm, line 32: Is this two separate commands?  If so, you'll have to use a `shell` instead of a `command`:
```yaml
shell: |
  set -euo pipefail
  nmcli c reload
  nmcli d set {{ interface }} managed true
```
Or, alternately, if `nmcli` can take a "compound" command - not sure about the syntax, something like
`nmcli "c reload" - "d set {{ interface }} managed true"` - then you could use a `command`

nhosoi: Sorry, I missed it. I ran this test to see it was successful. But it seems it did not execute the task `Set up veth as managed by NetworkManager`. Removing my "approval".
$ tox -e qemu -- --image-file /path/to/CentOS-7-x86_64-GenericCloud.qcow2c --become --become-user root -- tests_bridge_nm.yml

nhosoi: Yep, the task fails as follows, but it's ignored.
```
TASK [Set up veth as managed by NetworkManager] ********************************************************************************
task path: /path/to/linux-system-roles/network/tests/tasks/manage_test_interface.yml:29
fatal: [/path/to/CentOS-7-x86_64-GenericCloud.qcow2c]: FAILED! => {"changed": false, "cmd": "nmcli c reload nmcli d set testnic1 managed true", "msg": "[Errno 2] No such file or directory", "rc": 2}
...ignoring
```

liangwen12year: I updated the patch, thanks.

nhosoi: Thank you for the discussion, @liangwen12year. I ran these two scripts in a row as you suggested.
tests_ethernet_initscripts.yml tests_ethernet_nm.yml

It worked just fine with rhel-guest-image-7.9-30.x86_64.qcow2, which has the NetworkManager package installed by default. But it turned out the package is not included in CentOS-7-x86_64-GenericCloud.qcow2c which I used for my testing first. That's why my test failed.

I noticed these tests explicitly install NetworkManager package.
```
tests/tests_regression_nm.yml
tests/tests_eth_pci_address_match_nm.yml
tests/tests_ethtool_coalesce_nm.yml
tests/tests_ethtool_features_nm.yml
tests/tests_ethtool_ring_nm.yml
tests/tests_provider_nm.yml
```
In the same manner, can we add a task to install `NetworkManager` to manage_test_interface.yml to make sure it is available before calling `nmcli`?

liangwen12year: I will, thanks for testing.

liangwen12year: Hi @nhosoi , I updated the patch again.

tests/tasks/manage_test_interface.yml
```diff
@@ -26,8 +26,27 @@
...
+  when:
+    - ansible_distribution_major_version == '7' and
+      ansible_distribution == 'CentOS'
+    - network_provider == 'nm'
+
 - name: Set up veth as managed by NetworkManager
-  command: nmcli d set {{ interface }} managed true
+  shell: |
+    nmcli c reload
+    nmcli d set {{ interface }} managed true
```
tyll, line 49: IMHO, the previous code here and the new code are both wrong. The role should/needs to be fixed to handle this instead of having the tests prepare the system to work. This will allow admins to use the network role to manager the IP configuration etc on veth interfaces.

PR #444, opened 2022-01-06
Support routing tables in static routes

liangwen12year:


liangwen12year:
I will keep adding the documentation, adding the commit message, adding the examples and improving the patch tomorrow.

library/network_connections.py
```diff
@@ -1133,6 +1135,28 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                        for table in f.read().split("\n"):
+                            if (
+                                len(table.split()) >= 2
+                                and table.split()[1] == r["table"]
+                            ):
+                                tableid = int(table.split()[0])
+                    if tableid != None:
+                        NM.IPRoute.set_attribute(
+                            rr, "table", Util.GLib().Variant.new_uint32(tableid)
+                        )
```
tyll, line 1154: Parsing rt_tables should probably be a separate method and requires good unit test coverage.

library/network_connections.py
```diff
@@ -1133,6 +1135,28 @@ def connection_create(self, connections, idx, connection_current=None):
                 rr = NM.IPRoute.new(
                     r["family"], r["network"], r["prefix"], r["gateway"], r["metric"]
                 )
+                if isinstance(r["table"], int):
+                    NM.IPRoute.set_attribute(
+                        rr, "table", Util.GLib().Variant.new_uint32(r["table"])
+                    )
+                elif isinstance(r["table"], Util.STRING_TYPE):
+                    tableid = None
+                    with open("/etc/iproute2/rt_tables", "r") as f:
```
thom311, line 1144: the ArgValidator is supposed to validate and normalize the input configuration early on, before any changes are done. I think this parsing should happen there, it should not blindly pass on any strings. Yes, that would mean that ArgValidator needs a way to inject such "host configuration" (to access the name mapping). Add a singleton -- that can be stubbed for unit tests.

---

As I said on IRC, I find it a very bad idea to do this. It makes the configuration in the playbook dependent on what files are on the target host. Not to mention the added complexity for little benefit. Routing tables are numbers and not something to be hidden by the user.

library/network_connections.py
thom311: and, you wouldn't want to parse this more than once (e.g. for each tableid). Instead, parse it once and keep it around. The linux system role is not going to stick around for long, so it should handle that the file could change -- and if the file cannot change, there is no need to parse this repeatedly. Just cache it.

liangwen12year: Good idea, cache the mapping to avoid parsing again.

library/network_connections.py
liangwen12year: I updated the patch.

library/network_connections.py
liangwen12year: I created the function `validate_route_tables()` for parsing the rt_tables, and I added the unit tests for testing the function.

tests/unit/test_network_connections.py
```diff
@@ -267,17 +268,21 @@ def do_connections_validate_nm(self, input_connections, **kwargs):
                     s6.clear_routes()
                     for r in kwargs["nm_route_list_current"][idx]:
                         r = parser.validate(r)
-                        r = NM.IPRoute.new(
+                        rr = NM.IPRoute.new(
```
liangwen12year, line 271: I renamed the var from `r` into `rr` for the sake of consistency of the var name in `./library/network_connections.py`,  
https://github.com/linux-system-roles/network/blob/main/library/network_connections.py#L1133

module_utils/network_lsr/argument_validator.py
```diff
@@ -1858,6 +1904,31 @@ def _validate_post(self, value, name, result):
...
+        table_cache = {}
+        for r in connection["ip"]["route"]:
+            if isinstance(r["table"], Util.STRING_TYPE):
+                tableid_find = False
+                if table_cache != {}:
+                    if r["table"] in table_cache.keys():
+                        tableid_find = True
+                        r["table"] = int(table_cache[r["table"]])
+                else:
+                    with open("/etc/iproute2/rt_tables", "r") as f:
```
thom311, line 1917: the parsing of the mapping should go to a separate function (that can be unit-tested)... and to unit-test it, you might need two functions: one that operates only on the `bytes file_content` (so it can be tested), and one that reads the file and passes it to the other function.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +246,51 @@ def _validate_string_min_length(self, value):
...
+            if isinstance(value, bool):
+                # bool can (probably) be converted to self.numeric_type,
+                # but here we don't want to accept a boolean value.
+                pass
+            elif isinstance(value, int):
+                val_min = 1
+                val_max = UINT32_MAX
+                v = int(value)
+            elif isinstance(value, Util.STRING_TYPE):
+                v = value
```
thom311, line 272: I think you should try to convert the string to number. If it's a number, then use the number and don't interpret that as an alias.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2071,30 @@ def _ipv6_is_not_configured(connection):
...
+
+class RouteTableParsing:
+    @classmethod
+    def parse_route_tables_mapping(cls, file_content):
+        table_cache = {}
+        for table in file_content.split("\n"):
+            table_mapping = table.split()
+            # This is the original code about the route tables parsing in iproute2,
+            # https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/rt_names.c?id=ade99e208c1843ed3b6eb9d138aa15a6a5eb5219#n38
+            if len(table_mapping) == 2 and table[0] != "#" and table[0] != "0":
```
thom311, line 2085: iproute also support a `rt_tables.d`  directory. That seems rather useful, and I would think that all users should drop their aliases in their own files...

Anyway, the code here should make an effort to be exactly like what iproute2 does in the parsing.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2071,30 @@ def _ipv6_is_not_configured(connection):
...
+                table_cache[table_mapping[1]] = table_mapping[0]
```
thom311, line 2086: `table_mapping[0]` is supposed to be a valid uint32 (the table number). Don't store it as string here. Return the table number.

tests/unit/test_network_connections.py
```diff
@@ -4081,6 +4095,145 @@ def test_convert_passwd_flags_nm(self):
...
+        )
+
+    def test_parse_route_tables_mapping(self):
+        """
+        Test that `RouteTableParsing.parse_route_tables_mapping()` will parse the bytes
+        file content in `/etc/iproute2/rt_tables` and return a dictionary containing
+        the mapping between the route table name and route table id.
+        """
+        rt_file = "/etc/iproute2/rt_tables"
+        with open(rt_file, "r") as f:
```
thom311, line 4230: the unit test shouldn't read files from disk on the system where the unit test runs.

Here, you are not even testing `parse_route_tables_mapping_from_file()`, but `parse_route_tables_mapping()`. Just hard code a large set of input "file_contents" that you test in the test here.

If you would add a test for `parse_route_tables_mapping_from_file()`, then you would need to have a file on disk. But that file must be one shipped by the unit tests itself. But `parse_route_tables_mapping_from_file()` is rather trivial. It doesn't seem necessary to test the "can read file" part.

-- at which point you might ask, why does  `parse_route_tables_mapping_from_file()` even exist as separate function. You could drop it and inline all the logic into `get_route_tables_mapping()`. But both is fine.

tests/unit/test_network_connections.py
```diff
@@ -4081,6 +4095,145 @@ def test_convert_passwd_flags_nm(self):
...
+        Test that the validation error will be raised when the named route table is not
+        found in the `/etc/iproute2/rt_tables`
+        """
+
+        self.test_connections[0]["ip"]["route"][0]["table"] = "eth"
+
+        self.assertRaisesRegex(
+            ValidationError,
+            "can not find route table '%s' in `/etc/iproute2/rt_tables`"
+            % (self.test_connections[0]["ip"]["route"][0]["table"]),
```
thom311, line 4193: this unit test tests some files on the system where it's tested. You cannot do this. The unit test must work everywhere, also on a system where the user edited ther rt_tables file.

What you need to do, is actually stub/patch the code `RouteTableParsing.get_route_tables_mapping = lamda: return { ... }` -- which is the main reason to have the parsing code in a separate class, that you can stub for testing.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2071,30 @@ def _ipv6_is_not_configured(connection):
                             "match.path is not supported by the running version of "
                             "NetworkManger.",
                         )
+        self.validate_route_tables(connection, idx)
+
+
+class RouteTableParsing:
```
thom311, line 2077: `IPRouteUtils` seems a better name.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2071,30 @@ def _ipv6_is_not_configured(connection):
...
+
+    @classmethod
+    def parse_route_tables_mapping_from_file(cls, filename):
+        with open(filename, "r") as f:
+            return cls.parse_route_tables_mapping(f.read())
+
+    @classmethod
+    def get_route_tables_mapping(cls):
+        if not hasattr(cls, "_cached_rt_tables"):
+            cls._cached_rt_tables = cls.parse_route_tables_mapping_from_file(
```
thom311, line 2097: `parse_route_tables_mapping_from_file()` can raise an exception (if the file doesn't exist). The caching `get_route_tables_mapping()` should never fail. If the files doesn't exist, return an empty mapping (but still a usable dictionary, so you don't have to handle the failure specially).

module_utils/network_lsr/argument_validator.py
liangwen12year: Changed into `IPRouteUtils`

module_utils/network_lsr/argument_validator.py
liangwen12year: Updated.
```python
            elif isinstance(value, Util.STRING_TYPE):
                if value.isnumeric():
                    v = int(value)
                else:
                    v = value
```

thom311: `isnumeric()` seems not the right thing here... it checks whether all characters are (individually) numeric characters. OK, that might amount to in fact being the same, but still.

What I mean is that in my mind there is no guarantee that from `value.isnumeric()` always follows that a conversion `int(value)` is going to succeed. These are not the same things. At least, it wouldn't hold for `value="-1"` -- which you might say, is a case we don't want to handle here (but I think we do. We should also not allow "-1" as a valid alias).

I think the best way to check whether a string can be converted to an int is by just trying to do it.
```
    try:
        v = int(value)
    except:
        v = value
```

module_utils/network_lsr/argument_validator.py
liangwen12year: I updated.

module_utils/network_lsr/argument_validator.py
liangwen12year: I changed.

tests/unit/test_network_connections.py
liangwen12year: I dropped the unit tests here.

liangwen12year: I dropped the unit tests here.

liangwen12year: And I also dropped the `parse_route_tables_mapping_from_file()`.

tests/unit/test_network_connections.py
liangwen12year: I stubbed for the test cases.

tests/unit/test_network_connections.py
```diff
@@ -4081,6 +4095,119 @@ def test_convert_passwd_flags_nm(self):
...
+        """
+        Test that the `validate_route_tables()` will find the table id mapping from
+        `IPRouteUtils.get_route_tables_mapping()`.
+        """
+
+        self.test_connections[0]["ip"]["route"][0]["table"] = "eth"
+
+        old_getter = (
+            network_lsr.argument_validator.IPRouteUtils.get_route_tables_mapping
+        )
```
thom311, line 4192: move the storing and restoring of the old_getter to the `SetUp` and teardown function of the test.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +246,54 @@ def _validate_string_min_length(self, value):
...
+                pass
+            elif isinstance(value, int):
+                val_min = 1
+                val_max = UINT32_MAX
+                v = int(value)
+            elif isinstance(value, Util.STRING_TYPE):
+                try:
+                    v = int(value)
+                except Exception:
+                    v = value
```
thom311, line 275: I think you should check that the string contains only a benign set of characters and that it's not empty. An alias should be a plain word, no spaces, or newlines.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +246,54 @@ def _validate_string_min_length(self, value):
...
+        except Exception:
+            pass
+        if v is None:
+            raise ValidationError(
+                name,
+                "route table must be the named or numeric tables but is '%s'" % (value),
+            )
+        if val_min is not None and v < val_min:
```
thom311, line 283: let's drop val_min and val_max. Just:
```
if isinstance(v, int):
   if v < 1:
      raise...
   if v > 0xFFFFFFFF:
      raise
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -1858,6 +1907,19 @@ def _validate_post(self, value, name, result):
     VALIDATE_ONE_MODE_NM = "nm"
     VALIDATE_ONE_MODE_INITSCRIPTS = "initscripts"
 
+    def validate_route_tables(self, connection, idx):
+        for r in connection["ip"]["route"]:
+            if isinstance(r["table"], Util.STRING_TYPE):
+                table_cache = IPRouteUtils.get_route_tables_mapping()
+                if r["table"] in table_cache.keys():
```
thom311, line 1914: drop the `.keys()`.

liangwen12year:
@thom311 , I updated the PR, did I addressed all your comments properly ?

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,65 @@ def _validate_string_min_length(self, value):
...
+                    % (value),
+                )
+        if isinstance(v, Util.STRING_TYPE):
+            if v == "":
+                raise ValidationError(
+                    name,
+                    "route table name can not be empty string" % (value),
+                )
+            disallowed_char = [" ", "\t", "\n"]
+            if any(char in v for char in disallowed_char):
```
thom311, line 299: `if not re.match('^[a-zA-Z0-9_.-]+$', v):` ?

module_utils/network_lsr/argument_validator.py
```diff
@@ -1858,6 +1919,20 @@ def _validate_post(self, value, name, result):
     VALIDATE_ONE_MODE_NM = "nm"
     VALIDATE_ONE_MODE_INITSCRIPTS = "initscripts"
 
+    def validate_route_tables(self, connection, idx):
+        iproute = IPRouteUtils()
+        for r in connection["ip"]["route"]:
+            if isinstance(r["table"], Util.STRING_TYPE):
+                table_cache = iproute.get_route_tables_mapping()
```
thom311, line 1926: `table_cache` is not a good name. Which table? What significance has the "cache" here? Call it what is is. For example, `mapping`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2083,101 @@ def _ipv6_is_not_configured(connection):
...
+
+            # skip empty lines or comments
+            if not line:
+                continue
+            if line[0] == ord(b"#"):
+                continue
+
+            # Split at the first space. We accept here any whitespace (unlike iproute2).
+            # Also, this will strip any leading/trailing whitespaces (unlike iproute2)
+            ll = line.split(maxsplit=1)
```
ffmancera, line 2112: I think this is not compatible with python 2.7. The implementation in python 2.7 is not using keywords.

https://docs.python.org/2.7/library/stdtypes.html#str.split

module_utils/network_lsr/argument_validator.py
```diff
@@ -1858,6 +1916,19 @@ def _validate_post(self, value, name, result):
     VALIDATE_ONE_MODE_NM = "nm"
     VALIDATE_ONE_MODE_INITSCRIPTS = "initscripts"
 
+    def validate_route_tables(self, connection, idx):
+        for r in connection["ip"]["route"]:
+            if isinstance(r["table"], Util.STRING_TYPE):
+                mapping = IPRouteUtils.get_route_tables_mapping()
```
ffmancera, line 1922: I think this is incorrect. Looking at the implementation of the get_route_tables_mapping() it is a class method. Therefore you cannot call it directly. Normally, you need an IPRouteUtils AFAICS. Thanks!

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2083,101 @@ def _ipv6_is_not_configured(connection):
...
+    def _parse_route_tables_mapping_from_file(cls, filename, mapping):
+        try:
+            with open(filename, "rb") as f:
+                file_content = f.read()
+        except Exception:
+            return
+        cls._parse_route_tables_mapping(file_content, mapping)
+
+    @classmethod
+    def get_route_tables_mapping(cls):
```
ffmancera, line 2163: It is not clear to me why is this using the `@classmethod` decorator. It is a problem if you create an instance? I think it would simplify the code.

tests/unit/test_network_connections.py
ffmancera: Note for the future: We need to translate `r` and `rr` to meaningful names.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,62 @@ def _validate_string_min_length(self, value):
...
+    def __init__(
+        self,
+        name,
+        required=False,
+        default_value=None,
+    ):
+        ArgValidator.__init__(self, name, required, default_value)
+
+    def _validate_impl(self, value, name):
+        v = None
```
ffmancera, line 260: What about using a meaningful name for `v`? What does `v` mean? maybe value? But I can see that you are parsing `value` into `v`.. what is the difference between `value` and `v`? I would suggest something like: `parsed_value`. What do you think?

ffmancera: I know this is an extended issue on network-role.. not specific to this validator. But I would like to do not introduce more single char variable names.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2083,101 @@ def _ipv6_is_not_configured(connection):
...
+
+            # iproute2 only skips over leading ' ' and '\t'. We are more forgiving,
+            # and strip all leading+trailing whitespace.
+            line = line.strip()
+
+            # skip empty lines or comments
+            if not line:
+                continue
+            if line[0] == ord(b"#"):
+                continue
```
ffmancera, line 2108: You can simplify this on a single if:

```
if not line or line[0] == ord(b"#"):
    continue
```
The `or` operator is lazy, so if `line` is present it won't evaluate the next condition `line[0] == ord(b"#")`.

ffmancera (changes requested):
Thanks for all the efforts. Please check CI errors too. Thanks!

module_utils/network_lsr/argument_validator.py
liangwen12year: The reason that I separated the condition is because I want to make it more clear. (skip empty lines or comments during parsing) But either way is fine.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2084,106 @@ def _ipv6_is_not_configured(connection):
...
+                i = line2.find(b"\t")
+            if i >= 0:
+                # again, iproute2 only accepts one ' ', followed by '#'. We accept any number of
+                # space and any kind of whitespace (the split() above already partly took care
+                # of those).
+                line3 = line2[i + 1 :].strip()
+
+                # In Python 2.x, there is no new type called `bytes`. And Python 2.x
+                # `bytes` is just an alias to the str type
+                if sys.version[:1] == "2":
```
thom311, line 2128: there is `Util.PY3`

thom311: (the advantage of using that, is that you can grep for that thing and fine all places where we need different code depending on the python version... I guess, you still could grep for sys.version too, but this seems more explicit.

module_utils/network_lsr/argument_validator.py
liangwen12year: I removed the keyword.

thom311 (dismissed):

module_utils/network_lsr/argument_validator.py
liangwen12year: Notice that the method `get_route_tables_mapping` accesses other methods in the class,  so it is better to use classmethod here. And the benefits of using the `@classmethod` is to call the method without instantiate the class.

module_utils/network_lsr/argument_validator.py
liangwen12year: Compared to "value", the "v" is normalized value.

liangwen12year: I changed `v` into `norm_value`, thanks.

ffmancera (approved):
The current code looks good to me. Thank you for this awesome work!

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,62 @@ def _validate_string_min_length(self, value):
...
+    def __init__(
+        self,
+        name,
+        required=False,
+        default_value=None,
+    ):
+        ArgValidator.__init__(self, name, required, default_value)
+
+    def _validate_impl(self, value, name):
+        norm_value = None
```
tyll, line 260: IMHO, it would be great to avoid unnecessary abbreviations in variable such as "norm". Since this will be the routing table, using `table` here seems like a good choice.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,62 @@ def _validate_string_min_length(self, value):
...
+        try:
+            if isinstance(value, bool):
+                # bool can (probably) be converted to self.numeric_type,
```
tyll, line 263: `self.numeric_type` is not valid in this context.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,62 @@ def _validate_string_min_length(self, value):
...
+                try:
+                    norm_value = int(value)
+                except Exception:
+                    norm_value = value
+        except Exception:
+            pass
+        if norm_value is None:
+            raise ValidationError(
+                name,
+                "route table must be the named or numeric tables but is '%s'" % (value),
```
tyll, line 278: In this exception message, the `%s` is embedded inside single quotes, in the other messages it is not. It would be good to make this uniform. Also, for new code it seems to be best to use ".format()" which is already used in some areas.

module_utils/network_lsr/argument_validator.py
```diff
@@ -246,6 +247,62 @@ def _validate_string_min_length(self, value):
...
+                raise ValidationError(
+                    name,
+                    "route table value is %s but cannot be greater then 4294967295"
+                    % (value),
+                )
+        if isinstance(norm_value, Util.STRING_TYPE):
+            if norm_value == "":
+                raise ValidationError(
+                    name,
+                    "route table name can not be empty string" % (value),
```
tyll, line 296: This is using the % operator without a "%s" in the string, the % operator does not seem to be necessary here. Also, the other code is using "cannot" instead of "can not".

library/network_connections.py
tyll: Just for the record since we discussed this outside of the review: The table names exists because they make using tables more user friendly and I anticipate that users would like to use them. Also, the BZ for this feature mentioned named tables in their examples.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2083,114 @@ def _ipv6_is_not_configured(connection):
...
+
+    @classmethod
+    def _parse_route_tables_mapping(cls, file_content, mapping):
+        for line in file_content.split(b"\n"):
+
+            #
+            # We are a bit more forgiving, but the aliases must be nice UTF-8 names.
+
+            # iproute2 only skips over leading ' ' and '\t'. We are more forgiving,
+            # and strip all leading+trailing whitespace.
```
tyll, line 2101: What is the reason to deviate from iproute2, here?

tyll: Please note that `str.lstrip()` exists to skip over leading whitespace and it also accepts the characters to be stripped as a parameter.

README.md
```diff
@@ -473,8 +473,10 @@ The IP configuration supports the following options:
 
     Static route configuration can be specified via a list of routes given in the
     `route` option. The default value is an empty list. Each route is a dictionary with
-    the following entries: `network`, `prefix`, `gateway` and `metric`. `network` and
-    `prefix` specify the destination network.
+    the following entries: `network`, `prefix`, `gateway`, `metric` and `table`.
+    `network` and `prefix` specify the destination network. `table` supports both the
+    numeric table and named table. In order to specify the named table, the users have
+    to ensure the named table is properly defined in `/etc/iproute2/rt_tables`.
```
tyll, line 479: This is missing the support for the rt_tables.d directory.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1858,6 +1916,19 @@ def _validate_post(self, value, name, result):
...
+    def validate_route_tables(self, connection, idx):
+        for r in connection["ip"]["route"]:
+            if isinstance(r["table"], Util.STRING_TYPE):
+                mapping = IPRouteUtils.get_route_tables_mapping()
+                if r["table"] in mapping:
+                    r["table"] = mapping[r["table"]]
+                else:
+                    raise ValidationError.from_connection(
+                        idx,
+                        "can not find route table '%s' in `/etc/iproute2/rt_tables`"
```
tyll, line 1928: This is missing mentioning the rt_tables.d directory.

README.md
```diff
@@ -473,8 +473,11 @@ The IP configuration supports the following options:
 
     Static route configuration can be specified via a list of routes given in the
     `route` option. The default value is an empty list. Each route is a dictionary with
-    the following entries: `network`, `prefix`, `gateway` and `metric`. `network` and
-    `prefix` specify the destination network.
+    the following entries: `network`, `prefix`, `gateway`, `metric` and `table`.
+    `network` and `prefix` specify the destination network. `table` supports both the
+    numeric table and named table. In order to specify the named table, the users have
+    to ensure the named table is properly defined in `/etc/iproute2/rt_tables` and
+    `/etc/iproute2/rt_tables.d/*.conf`.
```
tyll, line 480:
```suggestion
    to ensure the named table is properly defined in `/etc/iproute2/rt_tables` or
    `/etc/iproute2/rt_tables.d/*.conf`.
```

tests/playbooks/tests_route_table.yml
```diff
@@ -0,0 +1,143 @@
...
+        that:
+          - route_table_30400.stdout is search("198.51.100.128/26 via
+            198.51.100.1 dev ethtest0 (proto static )?metric 2")
+        msg: "the route table 30400 does not exist or does not contain the
+          specified route"
+
+    - name: Add a new routing table
+      lineinfile:
+        path: /etc/iproute2/rt_tables
+        line: "200  custom"
```
tyll, line 82: it seems to me that this file is not cleaned after the test. It would probably be safer to use a dedicated file in rt_tables.d, ensure that it is created during the test and then removed at the end.

module_utils/network_lsr/argument_validator.py
liangwen12year: Originally I was tending to be more fault tolerant for the user when they added their custom route tables either through the Ansible lineinfile module or other methods. But after the second thought, I realized that this may inadvertently allow the user adding the table which can be parsed by the role but can not be parsed by iproute2.

liangwen12year:
The  iproute2 shipped in RHEL-6 or CentOS-6 will not parse the table from `/etc/iproute2/rt_tables.d/`. 

The iproute2 version from the CentOS-6 qcow2 image:
```bash
[root@ibm-p8-kvm-03-guest-02 ~]# ip -V
ip utility, iproute2-ss091226
```
And the  earliest tag for iproute2 that contained the keyword `rt_tables.d` is `v4.4.0`
(
tag name | v4.4.0 (7ca63aef7d1b0c808da0040c6b366ef7a61f38c1)
tag date | 2016-01-11 08:33:03 -0800
)

ffmancera:
> 

Hi Wen, it is not clear to me. Why should we skip it? IMO, we should adapt the code for RHEL 6/CentOS-6. For example if that is not supported on RHEL6/CentOS-6 it should fail.. what would happen if I try to use `rt_tables` on CentOS-6? Would the role fail or ignore the changes?

tyll:
1) Is there any request to support this for initscripts at all? IMHO it would be best to feature freeze the initscripts backend and only fix bugs there.
2) initscripts supports table names (it is just calling iproute2), so if we must support this feature in initscripts, the code does not need to (and maybe even should not) convert the names to numeric IDs (my preference is not to add more features to initscritps without strong requests).

liangwen12year:
> > 
> 
> Hi Wen, it is not clear to me. Why should we skip it? IMO, we should adapt the code for RHEL 6/CentOS-6. For example if that is not supported on RHEL6/CentOS-6 it should fail.. what would happen if I try to use `rt_tables` on CentOS-6? Would the role fail or ignore the changes?

When appending the custom table names to `/etc/iproute2/rt_tables` in RHEL-6 and CentOS-6, the test passed as successfully.

liangwen12year:
> 1. Is there any request to support this for initscripts at all? IMHO it would be best to feature freeze the initscripts backend and only fix bugs there.
> 2. initscripts supports table names (it is just calling iproute2), so if we must support this feature in initscripts, the code does not need to (and maybe even should not) convert the names to numeric IDs (my preference is not to add more features to initscritps without strong requests).

I checked the bz bug about the routing table feature request , the user did not request the support on the initscripts side.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2012,3 +2083,102 @@ def _ipv6_is_not_configured(connection):
...
+class IPRouteUtils(object):
+
+    ROUTE_TABLE_ALIAS_RE = re.compile("^[a-zA-Z0-9_.-]+$")
+
+    @classmethod
+    def _parse_route_tables_mapping(cls, file_content, mapping):
+        for line in file_content.split(b"\n"):
+
+            #
+            # the aliases must be nice UTF-8 names.
```
tyll, line 2098: This comment does not seem to belong here, utf-8 is handled later

liangwen12year:
@tyll , I resolved all the comments and I checked the PR again meticulously.

tyll:
> @tyll , I resolved all the comments and I checked the PR again meticulously.

Thanks. My concerns seem to be addressed.

ffmancera (approved):
LGTM!

PR #446, opened 2022-01-07
Fix problem when switch provider from initscript to nm

cathay4t:
Problem:

After `tests_bridge_initscripts.yml` passed, the `tests_bridge_nm.yml`
will fail NetworkManager 1.18.

Root cause:

 1. The `absent` and `down` action of initscript provider will not
    remove the bridge interface which fail the assertion in
    `tests_bridge_nm.yml`.
 2. In initscript mode, network role will create ifcfg file with
    `NM_CONTROLLED=no` instructing NetworkManager to mark the bridge as
    unmanaged. The follow up `down` and `absent` action of initscript
    provider will not change the NetworkManager's understanding on
    unmanaged state of this interface.

Fixes:
 1. We cannot change existing behaviour of initscript on not deleting
    interface in `down` and `absent` action. So we change the test
    function `tests/playbooks/down_profile.yml` to delete the interface
    manually via `ip link del <ifname>` command.

 2. Use command `nmcli connection reload` in `tasks/main.yml` before
    applying network configurations for nm provider.

Previous test infrastructure is running each test file in a brand new VM
or container which cause this problem not been found before.

Enabled the github CI to test the same scenario.


tasks/main.yml
```diff
@@ -50,6 +50,18 @@
     - network_provider == "nm"
   no_log: true
 
+# If user switch from initscripts provider where NM_CONTROLLED=no defined in
+# ifcfg-ethX file, NetworkManager daemon will treat that interface as
+# strickly unmanaged, even the follow up deletion of ifcfg-ethX file cannot
+# change the NetworManager's unmanaged state of this interface. This will
+# prevent any follow up "nm" provider action on this interface.
+# To solve that, we instruct NetworkManager to reload the configuration.
+- name: Reload NetworkManager configration
```
richm, line 59: configuration

tasks/main.yml
```diff
@@ -50,6 +50,18 @@
...
+  ansible.builtin.command: nmcli connection reload
```
richm, line 60: please just use `command` - no need to use the FQCN here

tests/playbooks/down_profile.yml
```diff
@@ -8,3 +8,11 @@
         state: down
   roles:
     - linux-system-roles.network
+
+- name: Delete the interface when initscripts
+  hosts: all
+  tasks:
+    - ansible.builtin.command: ip link del "{{ profile }}"
```
richm, line 15: please use `command` here instead of FQCN

richm, line 15: please add a task name e.g. `name: Delete the interface when initscripts`

tests/playbooks/down_profile.yml
```diff
@@ -8,3 +8,11 @@
...
+      ignore_errors: true
```
richm, line 16: add an ansible-lint suppression comment e.g. `# noqa ignore-errors`
https://ansible-lint.readthedocs.io/en/latest/rules.html#false-positives-skipping-rules

tests/playbooks/down_profile.yml
```diff
@@ -8,3 +8,11 @@
         state: down
   roles:
     - linux-system-roles.network
+
+- name: Delete the interface when initscripts
```
liangwen12year, line 12: how about "Delete the interface when the network provider is initscripts" ?

cathay4t:
> 



> Hi Gris, as suggested by Till here ([#443 (comment)](https://github.com/linux-system-roles/network/pull/443#discussion_r777471723)) , can you remove the following part of code snippets in `tests/tasks/manage_test_interface.yml` and the tests still pass as expected ?
> 
> ```yaml
> - name: Set up veth as managed by NetworkManager
>   command: nmcli d set {{ interface }} managed true
>   # The varible for `network_provider` is not exists yet,
>   # just ignore error for initscripts
>   ignore_errors: yes  # noqa ignore-errors
>   when: "type == 'veth' and state == 'present'"
> ```

In theory, NM should automatically mark interface as  managed when its profile been activated. But let's do that in another PR.

cathay4t:
New patch uploaded. Please review when you have time.

tests/playbooks/down_profile.yml
```diff
@@ -8,3 +8,12 @@
...
     - linux-system-roles.network
+
+- name: Delete the interface when the network provider is initscripts
+  hosts: all
+  tasks:
+    - name: Delete the interface when the network provider is initscripts
+      command: ip link del "{{ profile }}"
+      ignore_errors: true   # noqa ignore-errors
+      when:
+        - network_provider == "initscripts"
```
tyll, line 19: The other change implies that the goal is to support switching from initscripts to NM (which seems to be correct to me). This means, that the tests should not be altered and that we should have an actual test that tests that switching from initscripts to NM actually works.

tasks/main.yml
```diff
@@ -50,6 +50,18 @@
...
   no_log: true
 
+# If user switch from initscripts provider where NM_CONTROLLED=no defined in
+# ifcfg-ethX file, NetworkManager daemon will treat that interface as
+# strickly unmanaged, even the follow up deletion of ifcfg-ethX file cannot
+# change the NetworManager's unmanaged state of this interface. This will
+# prevent any follow up "nm" provider action on this interface.
+# To solve that, we instruct NetworkManager to reload the configuration.
+- name: Reload NetworkManager configuration
+  command: nmcli connection reload
```
tyll, line 60: What is the reason to do this with nmcli instead of doing it via libnm in the python? Also, this specific command might have side effects beyond the interfaces that are specified in network_interfaces, so it might not be a good idea. I believe it has been considered in the past. @thom311 what's your opinion?

tests/playbooks/down_profile.yml
cathay4t: I have explained this in the reason of this change in the first comment: The absent and down action of initscript provider will not remove the bridge interface which fail the assertion in `tests_bridge_nm.yml`.

i will include a dedicate test case for this.

cathay4t: Dedicate test case included and reproduced on RHEL/CentOS 7.

cathay4t:
This problem only exists in RHEL/CentOS 7 where NM is 1.18.
So this PR only reload the config on NM 1.18 to minimize the impact.

Also include the patch which enable the CentOS 7 CI.

tasks/main.yml
cathay4t: Changed to use `libnm` to reload the configuration when on NM 1.18.

.github/Dockerfile.c7-network-role
```diff
@@ -0,0 +1,11 @@
+FROM quay.io/centos/centos:centos7
+
+RUN yum -y install epel-release && \
+    yum -y upgrade && \
+    yum -y install NetworkManager NetworkManager-wifi \
+    procps-ng iproute ansible openssh-server openssh-clients \
+    dnsmasq hostapd wpa_supplicant openssl ethtool iputils
```
tyll, line 7: AFAIK, it is also recommended to run "yum clean all" at the end to avoid problems with outdated yum caches in case any of the tests install something, too.

tests/playbooks/tests_switch_provider.yml
```diff
@@ -0,0 +1,66 @@
...
+      include_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: "{{ interface }}"
+            state: up
+            type: bridge
+            ip:
+              dhcp4: false
+              auto6: true
```
tyll, line 25: just wondering, why set `auto6` to `true`?

cathay4t: Just copy from other yml. Changed to false.

.github/Dockerfile.c7-network-role
cathay4t: Added. Thanks!

tests/playbooks/tests_switch_provider.yml
```diff
@@ -0,0 +1,66 @@
...
+      vars:
+        network_connections:
+          - name: "{{ interface }}"
+            state: up
+            type: bridge
+            ip:
+              dhcp4: false
+              auto6: false
+    - include_tasks: tasks/assert_device_present.yml
+    - name: "Remove bridge {{ interface }} via initscripts provider"
```
tyll, line 27: The name says "remove" and the later comment mentions initscripts does not remove the interface. So this should be something like "Take the profile {{ interface }} down via initscripts provider".

tests/playbooks/down_profile.yml
tyll: In the new integration test you use a task in a file `delete_interface.yml` which removed interfaces and is slightly different (using the interface variable and also setting `changed_when`):

https://github.com/linux-system-roles/network/blob/main/tests/tasks/delete_interface.yml

Also, in the new integration test, you are writing down the content of the original `down_profile.yml` instead of using it, since it is now not down_profile but down_profile+delete_interface. To make it easier to understand what is going on, it seems to me that a file that is named so that it explains what happens, would be better.

tests/playbooks/tests_switch_provider.yml
cathay4t: Fixed. Thanks.

tests/playbooks/down_profile.yml
tyll: this does not seem to be resolved. At least the `changed_when` should be added and the other remarks should be considered.

ffmancera:
I have rebased the PR and added a patch. Let's see if it fixes all the versions as I expect or not.

tests/playbooks/down_profile.yml
ffmancera: I have introduced a new patch addressing your comments. Could you check if it looks good to you know? Thanks!

tests/playbooks/tests_eth_dns_support.yml
```diff
@@ -108,7 +108,7 @@
           - "'timeout:1' in ipv6_dns.stdout"
         msg: "DNS options are configured incorrectly"
 
-- import_playbook: down_profile.yml
+- import_playbook: down_profile+delete_interface.yml
```
tyll, line 111: The test is removing the interface already at the end, so it seems this is now done twice.

ffmancera: This is happening in multiple tests. Fixing it.

ffmancera: Fixed! Thanks!

.github/run_test.sh
```diff
@@ -21,7 +23,7 @@ tests_ethtool_ring_nm.yml
 tests_ipv6_disabled_nm.yml
 tests_ipv6_nm.yml
 tests_vlan_mtu_nm.yml
-tests_wireless_nm.yml
+tests_switch_provider.yml
```
tyll, line 26: Not sure why this change is here, I missed it before. I assume this is because the wireless test does not work on EL7. Would be good to track that this test needs to be added back to be running for EL8. This will probably need some additional support for different test lists for different environments.

tests/playbooks/tests_switch_provider.yml
```diff
@@ -0,0 +1,66 @@
...
+    - include_tasks: tasks/assert_device_present.yml
+    - name: "Take the profile {{ interface }} down via initscripts provider"
+      include_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
+          - name: "{{ interface }}"
+            persistent_state: absent
+            state: down
+            type: bridge
```
tyll, line 35: This is the duplication of the content of `down_profile.yml`. Since the tests need to be cleaned up anyhow, it might be fine to add more tech debt here. IMHO, it would be better to move more towards a cleaner test format like https://github.com/linux-system-roles/network/blob/main/tests/playbooks/tests_states.yml or something better than that.
By checking that file I noticed, that it contains an exception for the case that this PR is supposed to fix, so that should be removed, too:
https://github.com/linux-system-roles/network/blob/main/tests/playbooks/tests_states.yml#L129

.github/run_test.sh
ffmancera: Nice, catch. I didn't notice it before. Let me check if it works on EL7. If it does not work on EL7 I will add a comment explaining it. I think it should be simple to add a support for different test lists depending on the environment.

ffmancera: I have added it back. I will check the CI.

tests/playbooks/tests_switch_provider.yml
ffmancera: I have done several changes. down_profile keeps doing the same thing than before. down_profile+delete_interface is a new file that is importing down_profile and using delete_interface task. This is avoiding duplicate code.

ffmancera: Also removed the exception in `tests_states.yml`

.github/run_test.sh
ffmancera: test_wireless does not work on el7. I have added a FIXME comment.

tests/playbooks/tests_switch_provider.yml
```diff
@@ -0,0 +1,66 @@
...
+            ip:
+              dhcp4: false
+              auto6: false
+    - include_tasks: tasks/assert_device_present.yml
+    - name: "Remove bridge {{ interface }} via nm provider"
+      include_role:
+        name: linux-system-roles.network
+      vars:
+        network_connections:
```
liangwen12year, line 66: how about importing `tasks/remove+down_profile.yml`  here?

ffmancera: Please, notice the comment: `# NetworkManager should not remove pre-exist interface for down/absent`. So this is why we are asserting that the device is present and then deleting it with `delete_interface.yml` instead of doing it in one step.

Thanks!

ffmancera: Oh, my bad. I didn't read this correctly. Yes, let's use remove+down_profile.yml, thanks!

liangwen12year:
Looks good to me now, thank!

liangwen12year (approved):

PR #450, opened 2022-02-02
CHANGELOG: update changelog for 1.6.0 release

liangwen12year:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
 Changelog
 =========
 
+[1.6.0] - 2022-02-02
+--------------------
+
+### New features
+
+- Support routing tables in static routes
+
+### Bug fixes
+
+- Fix setting DNS search settings when only one IP family is enabled
+- Fix switching from initscripts to NetworkManager 1.18
+
 [1.5.0] - 2021-12-14
 --------------------
 


CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
...
+--------------------
+
+### New features
+
+- Support routing tables in static routes
+
+### Bug fixes
+
+- Fix setting DNS search settings only for enabled IP families
+- Fix reloading the configuration for nm provider on NetworkManager 1.18
```
ffmancera, line 14: I think the right description would be "Fix switching providers on NetworkManager 1.18". What do you think?

ffmancera (changes requested):

CHANGELOG.md
tyll: Technically, it is not only about switching the provider but also using the role the first time with the NM provider on a system that used network-scripts in the past and NM_CONTROLLED=no.

CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
...
+[1.6.0] - 2022-02-02
+--------------------
+
+### New features
+
+- Support routing tables in static routes
+
+### Bug fixes
+
+- Fix setting DNS search settings only for enabled IP families
```
tyll, line 13:
```suggestion
- Fix setting DNS search settings if only one IP family is enabled
```

CHANGELOG.md
tyll:
```suggestion
- Fix switching from initscripts to NetworkManager 1.18
```

liangwen12year: I updated, thanks.

ffmancera (approved):
LGTM!

PR #462, opened 2022-02-11
Disable all initscripts provider tests on Fedora

liangwen12year:
The initscripts provider tests are unstable on Fedora and users are not
using the initscripts on Fedora, therefore, disable all initscripts
provider tests on Fedora.


tyll (approved):
Thank you, :+1:

PR #464, opened 2022-02-14
System Roles should consistently use ansible_managed in configuration files it manages

nhosoi:
bz#2044640

The network role create an ifcfg file for initscripts. The file
used to append a comment "# this file was created by ansible".
This patch replaces the proprietary string with the ansible
standard {{ ansible_managed }} to adjust to the other system
roles.

For the implementation, it borrowed the method from kernel_settings,
getting the ansible managed comment using the get_ansible_managed.j2
template and pass the comment to network_connections which is added
to the ifcfg file.

In case network_provider is nm, the comment is not added to the
ifcfg file.

Signed-off-by: Noriko Hosoi <nhosoi@redhat.com>


tests/playbooks/tests_reapply.yml
```diff
@@ -46,7 +46,11 @@
                     - 192.0.2.72/31
                   dhcp4: false
                   auto6: false
-          ignore_errors: true
+            ignore_errors: true
+            ansible_managed_new: "{{ __network_new_header }}"
```
tyll, line 50: These indentation changes look wrong.
Also, changing the test is surprising. I guess the change is intended for the tasks file.

tasks/main.yml
```diff
@@ -85,6 +85,9 @@
     force_state_change: "{{ network_force_state_change | default(omit) }}"
     connections: "{{ network_connections | default([]) }}"
     __debug_flags: "{{ __network_debug_flags | default(omit) }}"
+    ansible_managed_new: "{{ __network_new_header }}"
```
tyll, line 88: 1) What's the reason to have the `_new` suffix?
2) It seems that this should be an internal variable instead of part of the API (so it should use a `__` prefix)
3) What is the reason it needs a template instead of doing `__ansible_managed: "{{ ansible_managed | comment }}"`

tests/tasks/get_profile_stat.yml
```diff
@@ -24,6 +26,30 @@
...
   changed_when: false
 
-- name: Set NM profile exist flag based on the nmcli output
-  set_fact: lsr_net_profile_exists=true
+# lsr_net_profile_ansible_managed:
+# under NetworkManager's control, the comment is not added by design.
+# Thus, set it always to true.
+- name: >-
+    Set NM profile exist and true ansible_managed comment flag
+    based on the nmcli output
```
tyll, line 34: There is also a possibility to add some kind of comments to keyfiles via the NM API. If it is considered to be useful, then the role could be updated to do this, too.

tyll (commented):
Please see my comments. Specificially the public name seems to be troublesome.

tasks/main.yml
richm: > 1. What's the reason to have the `_new` suffix?

No reason.  What would you suggest?

> 
>     2. It seems that this should be an internal variable instead of part of the API (so it should use a `__` prefix)

+1

> 
>     3. What is the reason it needs a template instead of doing `__ansible_managed: "{{ ansible_managed | comment }}"`

Because it is the `template` plugin that evaluates `ansible_managed` - it isn't a "simple" variable that can be expanded.  And, in order to use the `template` plugin, a template file is required - one cannot apparently pass in a template string - it must be a template file to evaluate - https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_lookup.html#ansible-collections-ansible-builtin-template-lookup

tests/tasks/get_profile_stat.yml
richm: > There is also a possibility to add some kind of comments to keyfiles via the NM API. If it is considered to be useful, then the role could be updated to do this, too.

Maybe, but the intention is to have the `ansible_managed` header in files that are directly managed by the role.  Managing the keyfiles via the NM API sounds like some level of indirection, so I'm not sure.

tests/playbooks/tests_reapply.yml
richm: Right - `ansible_managed_new` should be passed as a key/value pair to the `network_connections` module, but `ignore_errors` should be outdented to the same level as `vars` e.g.
```yaml
          network_connections:
            ansible_managed_new: "{{ __network_new_header }}"
            ...
          ignore_errors: true
          vars:
            ...
```

tasks/main.yml
tyll: > > 1. What's the reason to have the `_new` suffix?
> 
> No reason. What would you suggest?

`__network_header: "{{__ansible_managed}}"`

tests/playbooks/tests_reapply.yml
tyll: I believe that it is not correct to pass `ansible_managed_new` here at all since it is supposed to happen in `tasks/main.yml`.

nhosoi: Thank you, @tyll and @richm. I'm fixing the indentation as @richm suggested.

> I believe that it is not correct to pass ansible_managed_new here at all since it is supposed to happen in tasks/main.yml.

Are you suggesting to remove `ansible_managed_new` at line 50 (now it is at line 49 before ignore_errors, though). Without `ansible_managed_new` here, tests_reapply_nm.yml fails as follows. I'd like to keep it to make the test pass if ok...
```
TASK [Show test_module_run] ***********************************************************************************
task path: /.../linux-system-roles/network/tests/playbooks/tests_reapply.yml:54
ok: [/home/nhosoi/.cache/linux-system-roles/rhel-8.qcow2] => {
    "test_module_run": {
        "changed": false,
        "failed": true,
        "msg": "missing required arguments: ansible_managed_new"
    }
}

TASK [Assert that reapply is found in log output] *************************************************************
task path: /.../linux-system-roles/network/tests/playbooks/tests_reapply.yml:57
fatal: [/.../.cache/linux-system-roles/rhel-8.qcow2]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'stderr'.
The error appears to be in '/.../linux-system-roles/network/tests/playbooks/tests_reapply.yml': line 57, column 11, but may
be elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
            var: test_module_run
        - name: Assert that reapply is found in log output
          ^ here
"}

TASK [network_connections] ************************************************************************************
task path: /.../linux-system-roles/network/tests/playbooks/tests_reapply.yml:64
fatal: [/.../.cache/linux-system-roles/rhel-8.qcow2]: FAILED! => {"changed": false, "msg": "missing required arguments: ansible_managed_new"}
...ignoring
```

tasks/main.yml
nhosoi: We already have `ansible_managed_new` variable in multiple roles. Do we want to replace it with `__ansible_managed` in other roles, too?

richm: > We already have `ansible_managed_new` variable in multiple roles. Do we want to replace it with `__ansible_managed` in other roles, too?

Eventually, yes, but we do not have to do it right away

tests/playbooks/tests_reapply.yml
richm: Ah, I see - the test `tests_reapply.yml` calls the module directly - it does not go through the role interface - so it must add the parameter - alternately, we could make the parameter optional, instead of required, and add some logic to the module code to skip ansible_managed handling - @tyll which would you prefer?

tests/tasks/get_profile_stat.yml
nhosoi: So, is it ok to leave the ifcfg file generated by NetworkManager without the "this file is managed by NM" type of comment? I searched around the ifcfg-rh plugin, but I could not find a way to add it...

tasks/main.yml
nhosoi: I replaced `ansible_managed_new` with `__ansible_managed` in https://github.com/linux-system-roles/network/pull/464/commits/270b141b2c1d7c8a81824196299e3d4b8cd24496.

tests/playbooks/tests_reapply.yml
nhosoi: Is it ok to leave this choice -- keeping `__ansible_managed` required and `__ansible_managed: "{{ __network_new_header }}"` in the test script?

nhosoi:
If these two items are ok as they are, I'd like to squash the 2 commits and remove WIP.
https://github.com/linux-system-roles/network/pull/464#discussion_r808584807
https://github.com/linux-system-roles/network/pull/464#discussion_r808583034
Thanks!

tests/tasks/get_profile_stat.yml
richm: If there is no way to do it, then I think we should not do it

tasks/main.yml
```diff
@@ -85,6 +85,9 @@
     force_state_change: "{{ network_force_state_change | default(omit) }}"
     connections: "{{ network_connections | default([]) }}"
     __debug_flags: "{{ __network_debug_flags | default(omit) }}"
+    __ansible_managed: "{{ __network_new_header }}"
+  vars:
+    __network_new_header: "{{ lookup('template', 'get_ansible_managed.j2') }}"
```
tyll, line 90: This was my proposal (I also removed the `network` prefix here since it is scoped by being a parameter for the `network_connections` module):

```suggestion
    __header: "{{ __ansible_managed }}"
  vars:
    __ansible_managed: "{{ lookup('template', 'get_ansible_managed.j2') }}"
```

tests/tasks/get_profile_stat.yml
tyll: ok

tests/playbooks/tests_reapply.yml
tyll: > Ah, I see - the test `tests_reapply.yml` calls the module directly - it does not go through the role interface - so it must add the parameter - alternately, we could make the parameter optional, instead of required, and add some logic to the module code to skip ansible_managed handling - @tyll which would you prefer?

Thank you for the clarification.

> Is it ok to leave this choice -- keeping `__ansible_managed` required and `__ansible_managed: "{{ __network_new_header }}"` in the test script?

it is correct to pass the parameter there, so they can be kept, just please with a different name (see my other comment).

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -145,6 +145,10 @@
                       features:
                         tx_tcp_segmentation: "no"
                         tx-tcp-segmentation: "no"
+                ansible_managed_new: "{{ __network_new_header }}"
+              vars:
+                __network_new_header: "{{
+                  lookup('template', 'get_ansible_managed.j2') }}"
```
liangwen12year, line 151: What is the  purpose of only adding this to `tests/playbooks/tests_ethtool_features.yml` and `tests/playbooks/tests_reapply.yml` ? For passing the integration test ?

tasks/main.yml
nhosoi: Thanks for the clarification, @tyll. So, you prefer to use `__header` for the network_connection parameter.

@richm, what do you think about sharing the parameter name among the roles?
For instance, I was planning to replace `ansible_managed_new` with `__ansible_managed` in the certificate role some time later (introduced by https://github.com/linux-system-roles/certificate/pull/112). Do we want to use `__header` for all the applicable parameters or we do not care using the different names?

tests/playbooks/tests_ethtool_features.yml
nhosoi: > What is the purpose of only adding this to `tests/playbooks/tests_ethtool_features.yml` and `tests/playbooks/tests_reapply.yml` ? For passing the integration test ?

Yes. Without adding them, tests including the playbooks fail. You wrote "only adding to ...". Do you think I missed adding the ansible managed parameter at any other places where I should have? Thanks!

tasks/main.yml
tyll: > Thanks for the clarification, @tyll. So, you prefer to use `__header` for the network_connection parameter.

Yes.

> @richm, what do you think about sharing the parameter name among the roles? For instance, I was planning to replace `ansible_managed_new` with `__ansible_managed` in the certificate role some time later (introduced by [linux-system-roles/certificate#112](https://github.com/linux-system-roles/certificate/pull/112)). Do we want to use `__header` for all the applicable parameters or we do not care using the different names?

I realized that instead of `__ansible_managed` it should pobably be `__lsr_ansible_managed`, since it is an linux system roles internal variable to avoid potential conflicts with something else using `__ansible_managed`.

tests/playbooks/tests_ethtool_features.yml
liangwen12year: No, you are not missing adding it somewhere else.

library/network_connections.py
```diff
@@ -2617,6 +2620,7 @@ def main():
             is_check_mode=run_env_ansible.module.check_mode,
             ignore_errors=params["ignore_errors"],
             force_state_change=params["force_state_change"],
+            ansible_managed=params["__header"],
```
tyll, line 2623: This does not seem to be used anymore.

library/network_connections.py
```diff
@@ -1679,6 +1680,7 @@ def __init__(
         is_check_mode=False,
         ignore_errors=False,
         force_state_change=False,
+        ansible_managed="",
```
tyll, line 1683: This does not seem to be used anymore.

nhosoi: Thank you so much, @tyll. (my mistakes... ): I removed the unused variables and pushed a new commit.

tests/tasks/assert_profile_present.yml
```diff
@@ -5,3 +5,8 @@
   assert:
     that: lsr_net_profile_exists
     msg: "profile {{ profile }} does not exist"
+
+- name: "assert that ansible managed comment in '{{ profile }}' is present"
+  assert:
+    that: lsr_net_profile_ansible_managed
```
richm, line 11: use `lsr_net_profile_ansible_managed | bool` to avoid bare variable problem

richm: hmm - but I see the `that` condition above does not use `| bool` either . . . odd that this is working with no warnings . . .

tests/tasks/get_profile_stat.yml
```diff
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
-- name: Initialize NM profile exist flag
-  set_fact: lsr_net_profile_exists=false
+- name: Initialize NM profile exist and ansible_managed comment flag
+  set_fact:
+    lsr_net_profile_exists=false
+    lsr_net_profile_ansible_managed=false
```
richm, line 6:
```yaml
set_fact:
  lsr_net_profile_exists: false
  lsr_net_profile_ansible_managed: false
```

tests/tasks/get_profile_stat.yml
```diff
@@ -24,6 +26,30 @@
...
 
-- name: Set NM profile exist flag based on the nmcli output
-  set_fact: lsr_net_profile_exists=true
+# lsr_net_profile_ansible_managed:
+# under NetworkManager's control, the comment is not added by design.
+# Thus, set it always to true.
+- name: >-
+    Set NM profile exist flag and ansible_managed flag true
+    based on the nmcli output
+  set_fact:
```
richm, line 35: see above

tests/tasks/get_profile_stat.yml
```diff
@@ -24,6 +26,30 @@
...
+- name: Check ansible_managed comment for the initscripts case
+  block:
+    - name: Get the ansible_managed comment in ifcfg-{{ profile }}
+      command: >-
+        grep "^# Ansible managed"
+        /etc/sysconfig/network-scripts/ifcfg-{{ profile }}
+      register: _result
+
+    - name: Verify the ansible_managed comment in ifcfg-{{ profile }}
+      set_fact:
```
richm, line 49: see above

tests/templates/get_ansible_managed.j2
```diff
@@ -0,0 +1 @@
+{{ ansible_managed | comment }}
```
richm, line 1: can this use the template from the role?

tests/tasks/assert_profile_present.yml
nhosoi: Is it ok to add `| bool` to `lsr_net_profile_exists`, too?
```
  assert:
    that: lsr_net_profile_exists
```

tests/tasks/get_profile_stat.yml
nhosoi: @richm, This is from set_fact doc.
```
The set_fact module takes ``key=value`` pairs or ``key: value``(YAML notation) as variables to set in the playbook scope. The 'key' is the resulting variable name and the value is, of course, the value of said variable.
You can create multiple variables at once, by supplying multiple pairs, but do NOT mix notations.
```
What'd be the preferable notation in `set_fact`? If there is just one line as `set_fact: lsr_net_profile_exists=true`, choose `=`; otherwise, the YAML notation?

tests/tasks/assert_profile_present.yml
richm: I would say - do not add `| bool` to either of them, if they are working - I would assume there are more instances of bare variables in `that` conditions - we should investigate why they are working

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -145,6 +145,10 @@
                       features:
                         tx_tcp_segmentation: "no"
                         tx-tcp-segmentation: "no"
+                __header: "{{ __lsr_ansible_managed }}"
+              vars:
+                __lsr_ansible_managed: "{{
+                  lookup('template', 'get_ansible_managed.j2') }}"
```
tyll, line 151:
```suggestion
                __header: "# Anisble managed test header"
```

Please consider to just using some placeholder header here since this test call of the internal module does not need to use the ansible header - this can simplify the test a little bit.

tests/tasks/get_profile_stat.yml
richm: If the surrounding network role code uses the `=` style, then use it, to be consistent - otherwise, use the yaml style

tests/templates/get_ansible_managed.j2
nhosoi: Instead, updated to use a hardcoded string in the tests following the suggestion from @tyll.

tests/playbooks/tests_ethtool_features.yml
nhosoi: Done. Thanks.

tests/tasks/get_profile_stat.yml
nhosoi: Both styles are found in the tests.
E.g., (in the main branch)
```
tests/tests_vlan_mtu_nm.yml
      set_fact:
        network_provider: nm

tests/playbooks/tasks/get_profile_stat.yml:  set_fact: lsr_net_profile_exists=false
tests/playbooks/tasks/get_profile_stat.yml:  set_fact: lsr_net_profile_exists=true
tests/playbooks/tasks/get_profile_stat.yml:  set_fact: lsr_net_profile_exists=true
```
I'm adding a new variable `lsr_net_profile_ansible_managed` with `set_fact` to tests/playbooks/tasks/get_profile_stat.yml, where `=` style was used. But I'd think it's ok to replace it with the YAML style.

tyll (approved):
Thank you, :+1:

richm:
[citest bad]

nhosoi:
Thank you for your reviews and discussions, @tyll, @richm, and @liangwen12year.

If you think it's safe to merge, may I ask you to do so?

I noticed there are some issues left in this pr.

Sorry about the 6 "Unresolved conversations" left. I should have cleaned them up before squashing. Now all of them are outdated and not sure how to close them...

There are also some CI test failures. They don't look introduced by this pr, though... 

E.g., RHEL-8.5.0-20211013.2/ansible-2.9/(citool) has 4 failed test cases; 3 are due to the remote host unreachable and one failed to download epel metadata for repo. It looks there was some network issue at the moment?

rhel-7/ansible-2.12 failed in tests_wireless_nm.yml as follows. Is it something we should worry about?
```
TASK [fedora.linux_system_roles.network : Configure networking connection profiles] ***
[003] <info>  #0, state:up persistent_state:present, 'wlan0': add connection wlan0, 1acdf0b7-e20e-45c9-a0f1-a6b5c8f36914
[004] <info>  #0, state:up persistent_state:present, 'wlan0': up connection wlan0, 1acdf0b7-e20e-45c9-a0f1-a6b5c8f36914 (is-modified)
[005] <error> #0, state:up persistent_state:present, 'wlan0': up connection failed while waiting: connection not activated: ssid-not-found
```

And 3 pending citool tests {CentOS-7-latest/ansible-2.9/(citool), CentOS-8-latest/ansible-2.9/(citool), RHEL-7.9-20200917.0/ansible-2.9/(citool)} failed in setting up the guest VM with
```
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```
It is not an issue of the network role, but of the baseos ci system. The status of the test themselves is `CI pipeline failed to finish` and although `Tests did not run correctly`, they don't show failed on this pr, but stay in `Pending`.

cathay4t (approved):

PR #466, opened 2022-02-21
Fix hostapd package installation error

liangwen12year:
The hostapd package cannot be installed via RHEL Repo until RHEL-8.6,
as a result, any RHEL-8 of which the version released before REHL-8.6
has to install the hostapd via CentOS Stream in order to run managed
host testing.

### DIFF
--- a/tests/tasks/setup_802_1x_server.yml
+++ b/tests/tasks/setup_802_1x_server.yml
@@ -1,5 +1,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
+# This task can be removed once the RHEL-8.5 is not tested anymore
+- name: Install hostapd via CentOS Stream
+  command: dnf -y install http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/Packages/hostapd-2.10-1.el8.x86_64.rpm
+  when:
+    - ansible_distribution_version | float < 8.6
+    - ansible_distribution_major_version == '8'
+    - ansible_distribution == 'RedHat'
+
 - name: Install hostapd
   package:
     name: hostapd
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -33,3 +33,4 @@ rules:
       /tests/tests_wireless_wpa3_sae_nm.yml
       /tests/tests_eth_pci_address_match_nm.yml
       /tests/playbooks/tests_eth_pci_address_match.yml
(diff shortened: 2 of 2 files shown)


nhosoi (dismissed):
lgtm

tests/tasks/setup_802_1x_server.yml
```diff
@@ -4,6 +4,17 @@
...
+    - ansible_distribution_major_version != '8' or
+      ansible_distribution != 'RedHat' or
+      ansible_distribution_version | float >= 8.6
+
+- name: Install hostapd via CentOS Stream
+  command: dnf -y install http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/Packages/hostapd-2.10-1.el8.x86_64.rpm
+  when:
+    - ansible_distribution_version | float < 8.6 and
+      ansible_distribution_major_version == '8'
+    - ansible_distribution == 'RedHat'
```
tyll, line 17: Moving this before the package command will allow to remove the when statements for the package command. Also, a comment that explains that this can be removed once RHEL 8.5 is not tested anymore, would be great. The commit message should mention that hostapd was moved from EPEL to CentOS Stream and is therefore not available anymore on EPEL, which made the installation fail.

liangwen12year: I updated, thanks.

tests/tasks/setup_802_1x_server.yml
```diff
@@ -1,5 +1,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 ---
+# This task can be removed once the RHEL-8.5 is not tested anymore
+- name: Install hostapd via CentOS Stream
+  command: dnf -y install http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/Packages/hostapd-2.10-1.el8.x86_64.rpm
+  when:
+    - ansible_distribution_version | float < 8.6 and
```
richm, line 7: for the sake of consistency, you should not mix using an explicit `and`, and using the list format - so
```yaml
  when:
   - ansible_distribution_version | float < 8.6
   - ansible_distribution_major_version == '8'
   - ansible_distribution == 'RedHat'
```

liangwen12year: Thanks for notifying me on this. Updated.

tyll (approved):
Thank you

PR #469, opened 2022-02-28
use pgrep instead of pidof; ensure packages exist

richm:
The tests are giving failures that are unfortunately being ignored.
One of which is that pidof is not available by default on all platforms,
and on some platforms is not available at all (el7).  Same with sysctl.

There are a couple of fixes:
* run the large embedded test `shell` with `set -euo pipefail` to
trigger script errors to be handled.
* install the procps or procps-ng package which provides the tools
such as pgrep, pkill, sysctl, etc.

### DIFF
--- a/tests/tasks/create_test_interfaces_with_dhcp.yml
+++ b/tests/tasks/create_test_interfaces_with_dhcp.yml
@@ -5,12 +5,28 @@
     name: dnsmasq
     state: present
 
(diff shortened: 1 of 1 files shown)


richm:
@nhosoi fyi

cathay4t (approved):

cathay4t:
Thanks!

PR #470, opened 2022-03-02
Bond: Clarify and tests the order of destroying bond

liangwen12year:


README.md
```diff
@@ -264,6 +264,14 @@ specified first. Also, `--check` ignores the value of the `controller` and assum
 will be present during a real run. That means, in presence of an invalid `controller`,
 `--check` may signal success but the actual play run fails.
 
+When destroying and removing the bonding profiles, the order of the `connections` list
+also matters. If only specifying the `controller` profile in the `connections` list and
```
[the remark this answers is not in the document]
tyll: you mean to test if bringing down the controller will also bring down the ports?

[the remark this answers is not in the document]
liangwen12year: yes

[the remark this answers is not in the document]
tyll: yes, would be good to test this to ensure that the documentation matches the actual behavior.

cathay4t:
@liangwen12year Please fix the github CI. Both failures are about missing `bonding` kernel module.

README.md
```diff
@@ -264,6 +264,14 @@ specified first. Also, `--check` ignores the value of the `controller` and assum
...
+bringing it down, the port profiles will be brought down automatically. However, if not
+all the port profiles are brought down, the connection state of the `controller`
+profile will not be affected (stay `up`). Therefore, when it comes to destroy and
+remove the bonding profiles utterly, it is recommended to bring down and remove all the
+port profiles first, then bring down and remove the controller profile.
```
tyll, line 273: Why do the port profiles need to be removed first? From the text before that it seems it would be enough to destroy all profiles at the same time.

.github/run_test.sh
```diff
@@ -16,6 +16,7 @@ EXCLUDE_TESTS='
 -e tests/tests_bond_deprecated_initscripts.yml
 -e tests/tests_bond_deprecated_nm.yml
 -e tests/tests_bond_initscripts.yml
+-e tests/tests_bond_removal_initscripts.yml
```
tyll, line 19: This should be squashed into the other commit that is introducing this test.

README.md
liangwen12year: Right, it is true that bringing down the controller profile only will necessarily and automatically bringing down the port profile. In terms of destroying the bonding connection,  the port network connection dictionary and the controller network connection dictionary can be defined in the same network connections list, the "down" action will be executed before "absent" action, as a result,  all profiles in the bonding connection can be and destroyed and removed at the same time.

liangwen12year:
> > Meanwhile, why this activation/deactivation order is only about bond, why linux bridge, vlan not mentioned?
> 
> Notice that the clarification about the activation/deactivation is adding to the section `type: bridge, type: bond, type: team`, and we also explicitly indicate that `The bridge, bond, team device types work similar`. And here the tests using the bond as an example to confirm the clarification, I also tested destroying the bridges, it works the same as destroying bond (If only specifying the `controller` profile in the `connections` list and bringing it down, the port profiles will be brought down automatically. ) .

And for vlan, when running with nm provider and bringing down the parent, the vlan will stay up ( the connection is still activated on vlan). If running with initscripts provider and bringing down the parent, since initscripts admistratively call `ip link set $iface down`, the vlan will lose the carrier and no traffic passing through the vlan thereafter.

liangwen12year:
> > > Meanwhile, why this activation/deactivation order is only about bond, why linux bridge, vlan not mentioned?
> > 
> > 
> > Notice that the clarification about the activation/deactivation is adding to the section `type: bridge, type: bond, type: team`, and we also explicitly indicate that `The bridge, bond, team device types work similar`. And here the tests using the bond as an example to confirm the clarification, I also tested destroying the bridges, it works the same as destroying bond (If only specifying the `controller` profile in the `connections` list and bringing it down, the port profiles will be brought down automatically. ) .
> 
> And for vlan, when running with nm provider and bringing down the parent, the vlan will stay up ( the connection is still activated on vlan). If running with initscripts provider and bringing down the parent, since initscripts admistratively call `ip link set $iface down`, the vlan will lose the carrier and no traffic passing through the vlan thereafter. But the behavior above has nothing to do with the network role, it is purely determined by the nature of NM and intiscripts.

README.md
```diff
@@ -264,6 +264,15 @@ specified first. Also, `--check` ignores the value of the `controller` and assum
 will be present during a real run. That means, in presence of an invalid `controller`,
 `--check` may signal success but the actual play run fails.
 
+When destroying and removing the bonding profiles, if only specifying the `controller`
+profile in the `connections` list and bringing it down, the port profiles will be
+brought down automatically. However, if not all the port profiles are brought down,
+the connection state of the `controller` profile will not be affected (stay `up`).
```
ffmancera, line 270: This part is quite confusing for me. Could you separate the 3 cases?
1. Removing only the controller.
2. Removing all the ports.
3. Removing some ports.

ffmancera (changes requested):
Thanks! I've made a comment :-)

README.md
tyll: Also, this only needs to be documented, if there is something special that users need to consider/could be surprising. If it basically works in any order, then it might not need to be documented in such detail.

liangwen12year: In terms of removing all the ports, there is some difference between the nm provider and initscripts provider. For nm provider, if removing all the ports, we can still have the active controller connection. But for initscripts, if removing all the ports, the controller will also be brought down.

liangwen12year: The README and the tests have been updated. Thanks.

ffmancera (approved):
LGTM!

README.md
```diff
@@ -262,6 +262,10 @@ specified first. Also, `--check` ignores the value of the `controller` and assum
 will be present during a real run. That means, in presence of an invalid `controller`,
 `--check` may signal success but the actual play run fails.
 
+When destroying and removing the bonding profiles, if only bringing down the
```
tyll, line 265: Having both a "when..." and a "if ..." clause seems to be complicated and hard to understand. I don't understand this part, since it first mentions destroying and removing, then "bringing down". It would be best to simplify this as simple if, then sentences.

ffmancera (approved):

PR #472, opened 2022-03-04
add support to Rocky Linux

codeRhein:
Add support to Rocky Linux

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,7 +9,8 @@ network_provider_os_default: "{{
     'initscripts' if ansible_distribution in [
       'RedHat',
       'CentOS',
-      'OracleLinux'
+      'OracleLinux',
+      'Rocky'
     ] and ansible_distribution_major_version is version('7', '<')
     else 'nm' }}"
 # If NetworkManager.service is running, assume that 'nm' is currently in-use,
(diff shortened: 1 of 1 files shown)


richm:
thanks!
The DCO test failed - this repo only accepts signed commits (e.g. git commit -s)
note that we don't have CI testing with Rocky, so we can't be sure if everything will work on Rocky

liangwen12year:
If I understand it correctly, you only supported the Rocky Linux for initscripts provider, is there any user cases or user requests around this, if so, can you explain in the commit message?

codeRhein:
Thanks again ! This is also similar to the previous PR I opened, there is no strong request from our side, we can close this as well, sorry for the noise, thanks !

tyll:
@codeRhein if you sign off your commit as described in https://github.com/linux-system-roles/network/blob/main/contributing.md#sign-off-your-commit, then the PR is probably good to go.

codeRhein:
> @codeRhein if you sign off your commit as described in https://github.com/linux-system-roles/network/blob/main/contributing.md#sign-off-your-commit, then the PR is probably good to go.

thanks so much! I should read more carefully about the contribution section, now it is signed. Thanks again!

tyll (approved):
:+1:

PR #473, opened 2022-03-09
Remove connection profile before running pytest

liangwen12year:
Running the pytest with nm provider failed in the downstream testing
because the "NM_CONTROLLED=no" appeared in
`/etc/sysconfig/network-scripts/ifcfg-testeth` which caused the veth
`testeth` strictly unmanaged by NetworkManager. To fix it, remove such
a connection profile when it is existed before running the pytest.

### DIFF
--- a/tests/integration/test_ethernet.py
+++ b/tests/integration/test_ethernet.py
@@ -78,6 +78,15 @@ def testnic1():
         yield veth_name
     finally:
         subprocess.call(["ip", "link", "delete", veth_name])
+        if os.path.isfile("/etc/sysconfig/network-scripts/ifcfg-" + veth_name):
+            os.unlink("/etc/sysconfig/network-scripts/ifcfg-" + veth_name)
+            subprocess.call(
+                [
+                    "nmcli",
+                    "con",
+                    "reload",
+                ]
+            )
 
 
 def _get_ip_addresses(interface):


tests/integration/test_ethernet.py
```diff
@@ -61,6 +61,28 @@ def provider(request):
 def testnic1():
     veth_name = "testeth"
     try:
+        return_code = subprocess.call(
```
richm, line 64: you don't need to do a subprocess call for this - you can do
```python
if os.path.isfile("/etc/sysconfig/network-scripts/ifcfg-" + veth_name):
    os.unlink("/etc/sysconfig/network-scripts/ifcfg-" + veth_name)
```
unless you can't use `os` or `os.path` in this context?

liangwen12year: ok,right

liangwen12year: Updated. Thanks. It is wasteful to spawn another process indeed.

tests/integration/test_ethernet.py
```diff
@@ -61,6 +61,15 @@ def provider(request):
...
     try:
+        if os.path.isfile("/etc/sysconfig/network-scripts/ifcfg-" + veth_name):
+            os.unlink("/etc/sysconfig/network-scripts/ifcfg-" + veth_name)
+            subprocess.call(
+                [
+                    "nmcli",
+                    "con",
+                    "reload",
+                ]
+            )
```
richm, line 72: does `nmcli con reload` return immediately, or does it wait for the connection(s) to be reloaded?

liangwen12year: AFAIK, it returns immediately.

richm: Then is there a chance that the `ip link add testeth` call might fail if `nmcli con reload` has not completed?

tyll: If at all, the cleanup needs to happen in the `finally` block.

liangwen12year: > Then is there a chance that the `ip link add testeth` call might fail if `nmcli con reload` has not completed?

I honestly did not see the racing problem between the `ip link add testeth` and `nmcli con reload`.

richm:
lgtm

ffmancera (approved):
LGTM! AFAICS, `/etc/sysconfig/network-scripts/ifcfg-..` file should be present only when initscripts is the provider.. but yes, it seems redundant to me to check `nm_provider == initscripts` too.. Thanks!

ffmancera:
> LGTM! AFAICS, `/etc/sysconfig/network-scripts/ifcfg-..` file should be present only when initscripts is the provider.. but yes, it seems redundant to me to check `nm_provider == initscripts` too.. Thanks!

Oh according to @liangwen12year the ifcfg will be there also when the provider is NM.. well, it looks good to me anyway :-)

PR #481, opened 2022-03-11
fix: class Python26CompatTestCase broken by minor python versions

liangwen12year:

### DIFF
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -58,9 +58,9 @@ def pprint(msg, obj):
 class Python26CompatTestCase(unittest.TestCase):
     # pylint: disable=no-member
     def assertRaisesRegex(self, exception, regex, *args, **kwargs):
-        if sys.version[:3] == "2.6":
+        if sys.version_info[:2] == (2, 6):
             self.assertRaises(exception, *args, **kwargs)
-        elif sys.version[:3] < "3.2":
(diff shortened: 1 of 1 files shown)


tests/unit/test_network_connections.py
```diff
@@ -60,7 +60,7 @@ class Python26CompatTestCase(unittest.TestCase):
     def assertRaisesRegex(self, exception, regex, *args, **kwargs):
         if sys.version[:3] == "2.6":
             self.assertRaises(exception, *args, **kwargs)
-        elif sys.version[:3] < "3.2":
+        elif sys.version[:3] < "3.2" and sys.version[3] == ".":
```
thom311, line 63: why not use `sys.version_info`?

`if sys.version_info[:2] < (3,2):`

Even the doc says: "Do not extract version information out of it, rather, use version_info and the functions provided by the platform module." (https://docs.python.org/3/library/sys.html).

ffmancera: +1

liangwen12year: I updated. Thanks

tyll:
Please explain the problem that the patch is solving in the commit message. The code looks good, thanks @thom311 for the suggestion.

liangwen12year:
> Please explain the problem that the patch is solving in the commit message. The code looks good, thanks @thom311 for the suggestion.

Explained.

ffmancera (approved):

PR #482, opened 2022-03-11
[Draft] Reject configuring `ipv6 disabled` if not supported in NM

liangwen12year:
In NM, the property `NM_SETTING_IP6_CONFIG_METHOD_DISABLED` is only
supported since NM 1.20, therefore, the role should reject configuring
`ipv6 disabled` if `NM_SETTING_IP6_CONFIG_METHOD_DISABLED` is not
supported by the running version of NM.


module_utils/network_lsr/argument_validator.py
```diff
@@ -2216,6 +2217,15 @@ def _ipv6_is_not_configured(connection):
...
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+
+            if not self.TOX_UNIT_TEST and not hasattr(
+                Util.NM(), "SETTING_IP6_CONFIG_METHOD_DISABLED"
+            ):
+                raise ValidationError.from_connection(
+                    idx,
+                    "ip.ipv6_disabled is not supported by the running version of "
+                    "NetworkManger.",
```
tyll, line 2227: Please consider enhancing the message to mention what users can do instead, IIRC there is a way to make NM ignore Ipv6 in this case.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2216,6 +2217,15 @@ def _ipv6_is_not_configured(connection):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+
+            if not self.TOX_UNIT_TEST and not hasattr(
```
tyll, line 2221: What's the problem that `self.TOX_UNIT_TEST` is supposed to fix? You need to find another solution, maybe adjusting the unit test to expect something else or to mock or remove the `"SETTING_IP6_CONFIG_METHOD_DISABLED"`

module_utils/network_lsr/argument_validator.py
liangwen12year: How about: 
```python
                raise ValidationError.from_connection(
                    idx,
                    "ip.ipv6_disabled is not supported by the running version of "
                    "NetworkManger. However, if not explicitly enabling "
                    "ip.ipv6_disabled, ip.auto6, or specifying any IPv6 address, the "
                    "role will instruct NetworkManager not to configure IPv6 for this "
                    "connection",
                )
```

.github/workflows/tox.yml
```diff
@@ -9,6 +9,7 @@ env:
   LSR_ANSIBLES: 'ansible==2.9.*'
   LSR_MSCENARIOS: default
   # LSR_EXTRA_PACKAGES: "libdbus-1-dev libgirepository1.0-dev python3-dev"
+  LSR_EXTRA_PACKAGES: libgirepository1.0-dev gir1.2-nm-1.0
```
richm, line 12: i also had to add `pkg-config libcairo2-dev libnm-dev`

.github/workflows/tox.yml
```diff
@@ -9,6 +9,7 @@ env:
   LSR_ANSIBLES: 'ansible==2.9.*'
   LSR_MSCENARIOS: default
   # LSR_EXTRA_PACKAGES: "libdbus-1-dev libgirepository1.0-dev python3-dev"
+  LSR_EXTRA_PACKAGES: pkg-config libcairo2-dev libnm-dev libgirepository1.0-dev gir1.2-nm-1.0 python-gobject
```
richm, line 12: and `python-gi`

module_utils/network_lsr/argument_validator.py
liangwen12year: `Util.NM()` called the function here, 
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/utils.py#L65. 
Under the tox Python2 test virtual env, we can not import the `gi` module or install  pygobject in proper manner. So I propose to test differently regarding to the test env:

```
    def test_ipv6_dns_with_ipv6_disabled(self):
        """
        Test that configuring IPv6 DNS is not allowed when IPv6 is disabled.
        """
        validator = network_lsr.argument_validator.ArgValidator_ListConnections()
        ipv6_dns_with_ipv6_disabled = [
            {
                "name": "test_ipv6_dns",
                "type": "ethernet",
                "ip": {
                    "ipv6_disabled": True,
                    "dhcp4": True,
                    "dns": ["2001:db8::20"],
                },
            }
        ]

        if sys.version[:3] >= "3.6" and sys.version[3] == ".":
            self.assertRaisesRegex(
                ValidationError,
                "IPv6 needs to be enabled to support IPv6 nameservers.",
                validator.validate_connection_one,
                "nm",
                validator.validate(ipv6_dns_with_ipv6_disabled),
                0,
            )
        else:
            self.assertEqual(
                validator.validate(ipv6_dns_with_ipv6_disabled)[0]["ip"][
                    "ipv6_disabled"
                ],
                True,
            )
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -2216,6 +2216,15 @@ def _ipv6_is_not_configured(connection):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+            if not hasattr(Util.NM(), "SETTING_IP6_CONFIG_METHOD_DISABLED"):
+                raise ValidationError.from_connection(
+                    idx,
+                    "ip.ipv6_disabled is not supported by the running version of "
+                    "NetworkManger. However, if not explicitly enabling "
```
tyll, line 2223: After reading this, it is not clear to me why someone would need to use ipv6_disabled at all. What do you expect a user to do if setting ipv6_disabled is not supported.

tests/unit/test_network_connections.py
```diff
@@ -3723,14 +3723,23 @@ def test_ipv6_dns_with_ipv6_disabled(self):
...
-        self.assertRaisesRegex(
-            ValidationError,
-            "IPv6 needs to be enabled to support IPv6 nameservers.",
-            validator.validate_connection_one,
-            "nm",
-            validator.validate(ipv6_dns_with_ipv6_disabled),
-            0,
-        )
+
+        if sys.version[:3] >= "3.6" and sys.version[3] == ".":
```
tyll, line 3727: What does this if statement solve? It seems to replicate a different solution that you added in the past.

module_utils/network_lsr/argument_validator.py
liangwen12year: The users need to use `ipv6_disabled` when they want to disable IPv6 on a device, in the situation that `ipv6_disabled` is not supported, user can use `ipv6_ignore` instead (IPv6 is not required or IPv6 is handled by some other mechanism, and NetworkManager should not configure IPv6 for this connection), but this feature has not been implemented by the role. To simulate the `ipv6_ignore` setting, the user can set `ipv6_disabled` to False  (if not specifying, `ipv6_disabled` is set to False by default) and set `auto6` to False (if not specifying, `auto6` is set to True by default) and not specifying any IPv6 address. I got hint from:

```python
            if ip["ipv6_disabled"]:
                s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "disabled")
            elif ip["auto6"]:
                is_ipv6_configured = True
            elif addrs6:
                is_ipv6_configured = True
            else:
                # we should not set "ipv6.method=ignore". "ignore" is a legacy mode
                # and not really useful. Instead, we should set "link-local" here.
                #
                # But that fix is a change in behavior for the role, so it needs special
                # care.
                s_ip6.set_property(NM.SETTING_IP_CONFIG_METHOD, "ignore")
```

liangwen12year: I updated. Thanks.

tests/unit/test_network_connections.py
tyll: Please see https://github.com/linux-system-roles/network/blob/main/tests/unit/test_network_connections.py#L3981 how to skip a test.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2216,6 +2216,14 @@ def _ipv6_is_not_configured(connection):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+            if not hasattr(Util.NM(), "SETTING_IP6_CONFIG_METHOD_DISABLED"):
+                raise ValidationError.from_connection(
+                    idx,
+                    "ip.ipv6_disabled is not supported by the running version of "
+                    "NetworkManger. But you can make the role ignore IPv6 for this "
+                    "connection by setting ip.auto6 to False and not specifying any "
+                    "IPv6 address.",
```
tyll, line 2225: It would be good to mention the version that is required to get the support.  Also, the role will not ignore IPv6 but NetworkMananger and it seems to me valid to assume that user do not specify IPv6 addresses if they want to disable this, so the main fact is that auto6 needs to be disabled (please check/adjust my proposal to describe the effects properly). Also, it would be good to mention what will still be different on the system compared to setting ipv6_disabled to True (not in the NM config but the effect on the actual system/the kernel IP configuration). Maybe this actually belongs to the README and this should mainly be a pointer to that one.

```suggestion
                    "ip.ipv6_disabled is not supported by the running version of "
                    "NetworkManger, it requires at least version ... . To disable IPv6 auto configuration, set ip.auto6 to False. Then NetworkManager will ignore IPv6 for this profile. This will still ...",
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -2216,6 +2216,17 @@ def _ipv6_is_not_configured(connection):
                     idx,
                     "ip.ipv6_disabled is not supported by initscripts.",
                 )
+            if not hasattr(Util.NM(), "SETTING_IP6_CONFIG_METHOD_DISABLED"):
```
thom311, line 2219: if you run on a system with initscripts backend, then you cannot assume to have `Util.NM()`. Check with
`if mode == self.VALIDATE_ONE_MODE_NM:`.

liangwen12year: If you looked at the code holistically, the `Util.NM()`  will only be called when the mode is `self.VALIDATE_ONE_MODE_NM`, therefore, I think that the additional check is not needed. 
```python
        # initscripts does not support ip.ipv6_disabled, so raise errors when network
        # provider is initscripts
        if connection["ip"]["ipv6_disabled"]:
            if mode == self.VALIDATE_ONE_MODE_INITSCRIPTS:
                raise ValidationError.from_connection(
                    idx,
                    "ip.ipv6_disabled is not supported by initscripts.",
                )
            if not hasattr(Util.NM(), "SETTING_IP6_CONFIG_METHOD_DISABLED"):
                raise ValidationError.from_connection(
                    idx,
                    "ip.ipv6_disabled is not supported by the running version of "
                    "NetworkManager, it requires at least version 1.20. But you can "
                    "disable IPv6 auto configuration by setting ip.auto6 to False. "
                    "Then NetworkManager will ignore IPv6 for this connection. This "
                    "will still leave the sysctl value 'disable_ipv6' unchanged, but "
                    "setting ip.ipv6_disabled to True in the role will set the sysctl "
                    "value 'disable_ipv6' to True ",
                )
```

thom311: you are right. I didn't see!!

ffmancera (changes requested):
Hello! I noticed that the unit test `test_route_v6` is failing in multiple distributionse e.g Fedora 35. Could you check why? I suspect it could be related to the new mocks added but I am not sure. Other than that, it looks good to me :-)

Thanks!

liangwen12year:
Notice that I did not mock the property in the setup function and restore the mock in teardown function as it will cause the testing failures in `test_route_v6()` function:
```
FAIL: test_route_v6 (__main__.TestValidator)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/lsrtest_a4920m5h/test_network_connections.py", line 2391, in test_route_v6
    self.do_connections_validate(
  File "/tmp/lsrtest_a4920m5h/test_network_connections.py", line 335, in do_connections_validate
    self.do_connections_validate_nm(input_connections, **kwargs)
  File "/tmp/lsrtest_a4920m5h/test_network_connections.py", line 299, in do_connections_validate_nm
    self.assert_nm_connection_routes_expected(
  File "/tmp/lsrtest_a4920m5h/test_network_connections.py", line 244, in assert_nm_connection_routes_expected
    self.assertEqual(route_list_exp, route_list_new)
AssertionError: Lists differ: [{'network': '192.168.40.0', 'prefix': 24, '[586 chars]10>}] != [{'family': <MagicMock name='mock().SimpleCo[975 chars]6'>}]

First differing element 0:
{'network': '192.168.40.0', 'prefix': 24, '[78 chars]: 2>}
{'family': <MagicMock name='mock().SimpleCo[462 chars]76'>}

First list contains 3 additional elements.
First extra element 2:
{'network': '192.168.45.0', 'prefix': 24, 'metric': 545, 'gateway': None, 'table': None, 'family': <AddressFamily.AF_INET: 2>}

+ [{'family': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_family()' id='139734115795344'>,
+   'gateway': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_next_hop()' id='139734120456304'>,
- [{'family': <AddressFamily.AF_INET: 2>,
-   'gateway': None,
-   'metric': 545,
-   'network': '192.168.40.0',
-   'prefix': 24,
-   'table': None},
-  {'family': <AddressFamily.AF_INET: 2>,
-   'gateway': None,
-   'metric': -1,
?             -

+   'metric': 1,
-   'network': '192.168.46.0',
+   'network': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_dest()' id='139734115749552'>,
-   'prefix': 30,
?             ^^

+   'prefix': 1,
?             ^

+   'table': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_attribute()' id='139734120370576'>},
+  {'family': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_family()' id='139734115795344'>,
+   'gateway': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_next_hop()' id='139734120456304'>,
-   'table': None},
-  {'family': <AddressFamily.AF_INET: 2>,
-   'gateway': None,
-   'metric': 545,
-   'network': '192.168.45.0',
-   'prefix': 24,
-   'table': None},
-  {'family': <AddressFamily.AF_INET6: 10>,
-   'gateway': None,
-   'metric': -1,
?             -

+   'metric': 1,
-   'network': 'a:b:c:f::',
+   'network': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_dest()' id='139734115749552'>,
-   'prefix': 64,
?             ^^

+   'prefix': 1,
?             ^

+   'table': <MagicMock name='mock().SimpleConnection.new().get_setting().get_route().get_attribute()' id='139734120370576'>}]
-   'table': None},
-  {'family': <AddressFamily.AF_INET6: 10>,
-   'gateway': None,
-   'metric': -1,
-   'network': 'a:b:c:d::',
-   'prefix': 64,
-   'table': None}]

----------------------------------------------------------------------
Ran 132 tests in 1.570s

FAILED (failures=1)
```

ffmancera (approved):

PR #484, opened 2022-03-15
[citest skip]Enable more integration tests in GitHub CI

liangwen12year:

### DIFF
--- a/.github/run_test.sh
+++ b/.github/run_test.sh
@@ -8,47 +8,46 @@ C7_CONTAINER_IMAGE="quay.io/linux-system-roles/c7-network-role"
 C9S_CONTAINER_IMAGE="quay.io/linux-system-roles/c9s-network-role"
 PODMAN_OPTS="--systemd=true --privileged"
 
-# FIXME: test_wireless has been removed because is not supported on CI
-# container EL7. We need to add this back for EL8 or greater.
-EXCLUDE_TESTS='
--e tests/tests_802_1x_updated_nm.yml
+# exclude bond tests since missing the bonding kernel module
+# exclude tests/tests_wireless_nm.yml since failing to load mac80211_hwsim kernel
+# module to mock a wifi network
+EXCLUDE_TESTS_C7='
 -e tests/tests_auto_gateway_initscripts.yml
 -e tests/tests_bond_deprecated_initscripts.yml
--e tests/tests_bond_deprecated_nm.yml
 -e tests/tests_bond_initscripts.yml
 -e tests/tests_bond_removal_initscripts.yml
--e tests/tests_default_initscripts.yml
--e tests/tests_default_nm.yml
--e tests/tests_default.yml
--e tests/tests_ethernet_initscripts.yml
--e tests/tests_ethernet_nm.yml
--e tests/tests_eth_pci_address_match_nm.yml
--e tests/tests_ethtool_coalesce_initscripts.yml
--e tests/tests_ethtool_features_initscripts.yml
--e tests/tests_ethtool_ring_initscripts.yml
--e tests/tests_helpers_and_asserts.yml
--e tests/tests_integration_pytest.yml
--e tests/tests_ipv6_dns_search_nm.yml
--e tests/tests_ipv6_initscripts.yml
--e tests/tests_provider_nm.yml
--e tests/tests_reapply_nm.yml
--e tests/tests_regression_nm.yml
--e tests/tests_route_table_nm.yml
--e tests/tests_states_initscripts.yml
--e tests/tests_states_nm.yml
 -e tests/tests_team_nm.yml
--e tests/tests_team_plugin_installation_nm.yml
 -e tests/tests_unit.yml
--e tests/tests_vlan_mtu_initscripts.yml
 -e tests/tests_wireless_nm.yml
--e tests/tests_wireless_plugin_installation_nm.yml
+'
+
+# exclude bond tests since missing the bonding kernel module
+# exclude tests/tests_wireless_wpa3_owe_nm.yml and tests/tests_wireless_wpa3_sae_nm.yml
+# since failing to install mac80211_hwsim kernel module
+EXCLUDE_TESTS_C8S='
+-e tests/tests_auto_gateway_initscripts.yml
+-e tests/tests_bond_deprecated_initscripts.yml
+-e tests/tests_bond_initscripts.yml
+-e tests/tests_bond_removal_initscripts.yml
+-e tests/tests_integration_pytest.yml
+-e tests/tests_team_nm.yml
+-e tests/tests_unit.yml
 -e tests/tests_wireless_wpa3_owe_nm.yml
 -e tests/tests_wireless_wpa3_sae_nm.yml
 '
 
-read -r -d '' TEST_FILES << EOF || :
-$(find tests/tests_*.yml | egrep -v ${EXCLUDE_TESTS})
-EOF
+# exclude tests_provider_nm.yml and tests_regression_nm.yml since no package
+# network-scripts available
+# exclude tests/tests_wireless_wpa3_owe_nm.yml and tests/tests_wireless_wpa3_sae_nm.yml
+# since failing to install mac80211_hwsim kernel module
+EXCLUDE_TESTS_C9S='
+-e tests/tests_provider_nm.yml
+-e tests/tests_regression_nm.yml
+-e tests/tests_team_nm.yml
+-e tests/tests_unit.yml
+-e tests/tests_wireless_wpa3_owe_nm.yml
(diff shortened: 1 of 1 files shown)


liangwen12year:
On c8s, most of the tests failed on this or similar, 
TASK [remove the NetworkManager-team package] **********************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'epel': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=epel-8&arch=x86_64&infra=stock&content=centos [Could not resolve host: mirrors.fedoraproject.org]", "rc": 1, "results": []}
TASK [Install iproute] **********************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'epel': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=epel-8&arch=x86_64&infra=stock&content=centos [Could not resolve host: mirrors.fedoraproject.org]", "rc": 1, "results": []}

investigating now ...

.github/run_test.sh
```diff
@@ -72,19 +87,28 @@ while [[ $# -gt 0 ]]; do
...
+"c9s")
+    CONTAINER_IMAGE=$C9S_CONTAINER_IMAGE
+    read -r -d '' TEST_FILES <<EOF || :
+    $(find tests/tests_*.yml | egrep -v ${EXCLUDE_TESTS_C9S})
+EOF
+    ;;
+*)
+    echo "Unsupported OS type $OS_TYPE"
+    exit 1
+    ;;
```
ffmancera, line 111: Why is the indentation changing here?

ffmancera (changes requested):
I see the CI is failing now more than before..

.github/run_test.sh
```diff
@@ -10,44 +10,59 @@ PODMAN_OPTS="--systemd=true --privileged"
...
 -e tests/tests_ethernet_nm.yml
--e tests/tests_eth_pci_address_match_nm.yml
 -e tests/tests_ethtool_coalesce_initscripts.yml
+-e tests/tests_ethtool_coalesce_nm.yml
 -e tests/tests_ethtool_features_initscripts.yml
+-e tests/tests_ethtool_features_nm.yml
 -e tests/tests_ethtool_ring_initscripts.yml
+-e tests/tests_ethtool_ring_nm.yml
+-e tests/tests_eth_dns_support_nm.yml
+-e tests/tests_eth_pci_address_match_nm.yml
```
tyll, line 39: What's the reason to exclude more tests?

loveinmemory: The reason is that I wrongly enabled a test, and such a test left the system in a very bad state which caused a series of failure like this https://github.com/linux-system-roles/network/pull/484#issuecomment-1073833334. (Which test it is, I will verify it in the VM)

liangwen12year: The reason is that I wrongly enabled a test, and such a test left the system in a very bad state which caused a series of failure like this #484 (comment). (Which test it is, I will verify it in the VM)

.github/run_test.sh
```diff
@@ -10,7 +10,17 @@ PODMAN_OPTS="--systemd=true --privileged"
 
 # FIXME: test_wireless has been removed because is not supported on CI
 # container EL7. We need to add this back for EL8 or greater.
```
ffmancera, line 12: I think this comment is not true after your changes, could you remove it? In addition it would be nice to have comments on why some tests are excluded.

.github/run_test.sh
ffmancera: Thanks!

ffmancera (changes requested):
It looks good, just a nit. Thank you!

.github/run_test.sh
```diff
@@ -8,48 +8,49 @@ C7_CONTAINER_IMAGE="quay.io/linux-system-roles/c7-network-role"
...
+-e tests/tests_team_nm.yml
+-e tests/tests_unit.yml
+-e tests/tests_wireless_wpa3_owe_nm.yml
+-e tests/tests_wireless_wpa3_sae_nm.yml
+'
+
+# exclude tests_provider_nm.yml and tests_regression_nm.yml since no package
+# network-scripts available
+# exclude tests_wireless_wpa3_owe_nm.yml and tests_wireless_wpa3_sae_nm.yml since
+# destination directory /etc/hostapd not found in the container
```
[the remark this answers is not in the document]
liangwen12year: hostapd package is not installed for C9s.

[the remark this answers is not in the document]
liangwen12year: But even if we install the hostapd in C9s, the tests still fail in installiing mac80211_hwsim kernel module from here https://koji.mbox.centos.org/pkgs/packages/kernel/, because the container shares the kernel with ubuntu host.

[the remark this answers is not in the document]
liangwen12year: The following extracted from `tests/tasks/setup_mock_wifi_wpa3_owe.yml`, maybe we should not change into ansible_distribution_major_version != '7', because we can not install the mac80211_hwsim kernel module in C9s

```yaml
# Even though hostapd can be installed via EPEL 8, Opportunistic Wireless Encryption
# (OWE) has not been enabled by default. To warrant the test support on RHEL(dist-tag),
# we setup hostapd copr repo to enable OWE option.
- name: install hostapd and mac80211_hwsim kernel module in CentOS 8
  shell: |
    dnf -y copr enable liangwen12year/hostapd-owe
    dnf -y install hostapd
    release=$(uname -r)
    tmp="${release/-//}"
    tmp="${tmp/.x//x}"
    dnf -y install https://koji.mbox.centos.org/pkgs/packages/kernel/$tmp/kernel-core-$release.rpm
    dnf -y install https://koji.mbox.centos.org/pkgs/packages/kernel/$tmp/kernel-modules-$release.rpm
    dnf -y install https://koji.mbox.centos.org/pkgs/packages/kernel/$tmp/kernel-modules-internal-$release.rpm
  when:
    - ansible_distribution_major_version == '8'
    - ansible_distribution == 'CentOS'
```

.github/run_test.sh
```diff
@@ -8,48 +8,49 @@ C7_CONTAINER_IMAGE="quay.io/linux-system-roles/c7-network-role"
 C9S_CONTAINER_IMAGE="quay.io/linux-system-roles/c9s-network-role"
 PODMAN_OPTS="--systemd=true --privileged"
 
-# FIXME: test_wireless has been removed because is not supported on CI
-# container EL7. We need to add this back for EL8 or greater.
-EXCLUDE_TESTS='
+# exclude bond tests since missing the bonding kernel module
```
ffmancera, line 11: I am not sure if this is the real reason. Isn't c9s using the same ubuntu kernel? Why is it failing on c7/c8s but not on c9s? Thanks!

ffmancera (changes requested):

.github/run_test.sh
liangwen12year:
```
RUN_PLAYBOOK_WITH_INITSCRIPTS = """# SPDX-License-Identifier: BSD-3-Clause
# This file was generated by ensure_provider_tests.py
---
- hosts: all
  name: Run playbook '{test_playbook}' with initscripts as provider
  tasks:
    - include_tasks: tasks/el_repo_setup.yml
    - name: Set network provider to 'initscripts'
      set_fact:
        network_provider: initscripts
      tags:
        - always

- import_playbook: {test_playbook}
  when: (ansible_distribution in ['CentOS','RedHat'] and\n    \
ansible_distribution_major_version | int < 9)
"""
```
Initscripts provider tests are skipped for rhel-9 and centos-9.

liangwen12year:
https://github.com/linux-system-roles/network/pull/482, this PR breaks the  `tests_ipv6_dns_search_nm`.

.github/run_test.sh
```diff
@@ -8,48 +8,49 @@ C7_CONTAINER_IMAGE="quay.io/linux-system-roles/c7-network-role"
...
+'
+
+# exclude bond tests since missing the bonding kernel module
+# exclude tests/tests_wireless_wpa3_owe_nm.yml and tests/tests_wireless_wpa3_sae_nm.yml
+# since failing to install mac80211_hwsim kernel module
+EXCLUDE_TESTS_C8S='
 -e tests/tests_802_1x_updated_nm.yml
 -e tests/tests_auto_gateway_initscripts.yml
 -e tests/tests_bond_deprecated_initscripts.yml
 -e tests/tests_bond_deprecated_nm.yml
```
tyll, line 31: What's the reason this test does not work on CS 8 but it does on CS 9?

ffmancera (approved):
LGTM!

PR #485, opened 2022-03-18
README: Remove outdated information regarding the ifcfg-rh plugin.

tyll:

cc @briansmith0


liangwen12year:
LGTM, but get confused with `storage plugin` initially, maybe better to replace it with `ifcfg-rh plugin`, not sure.

liangwen12year (approved):

PR #486, opened 2022-03-20
Add support for routing rules

liangwen12year:
In order to enable the user using the policy routing (e.g. source
routing) or instructing the system which table to look up to determine
the correct route, add support for routing rules.


liangwen12year:
Will add the unit tests soon.

library/network_connections.py
```diff
@@ -1148,6 +1148,18 @@ def connection_create(self, connections, idx, connection_current=None):
                     s_ip4.add_route(rr)
                 else:
                     s_ip6.add_route(rr)
+            for r in ip["routing_rules"]:
+                rr = NM.IPRoutingRule.new(r["family"])
```
tyll, line 1152:
```suggestion
                routing_rule = NM.IPRoutingRule.new(r["family"])
```

library/network_connections.py
```diff
@@ -1148,6 +1148,53 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                            "unreachable",
+                        ]
+                    ):
+                        if action == r["action"]:
+                            # check the link below for the enum value of supported action
+                            # https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fib_rules.h#n88
+                            NM.IPRoutingRule.set_action(routing_rule, (1, 6, 7, 8)[i])
+                if r["family"] == socket.AF_INET:
+                    s_ip4.add_routing_rule(routing_rule)
+                else:
```
ffmancera, line 1196: The variables here are confusing. AFAICU, 'r' is the routing rule provided by the user and 'routing_rule' is the NM object. Could you improve this? maybe we can have something like 'routing_rule' and 'nm_routing_rule'.

ffmancera (changes requested):

library/network_connections.py
```diff
@@ -1148,6 +1148,61 @@ def connection_create(self, connections, idx, connection_current=None):
                     s_ip4.add_route(rr)
                 else:
                     s_ip6.add_route(rr)
+            for r in ip["routing_rules"]:
+                routing_rule = NM.IPRoutingRule.new(r["family"])
+                if r["priority"] is not None:
```
thom311, line 1153: the priority is mandatory. `ArgValidator` should always ensure that it's present.

library/network_connections.py
```diff
@@ -1148,6 +1148,61 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                        [
+                            "to-table",
+                            "blackhole",
+                            "prohibit",
+                            "unreachable",
+                        ]
+                    ):
+                        if action == r["action"]:
+                            # check the link below for the enum value of supported action
+                            # https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fib_rules.h#n88
```
thom311, line 1166: use the URL that contains the commit ID to which you point at. Otherwise, `#n88` (or the entire file) is not guaranteed to be there in the future.

library/network_connections.py
```diff
@@ -1148,6 +1148,61 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                            NM.IPRoutingRule.set_action(routing_rule, (1, 6, 7, 8)[i])
```
thom311, line 1167:
```python
    action_ids = {
        "to-table": 1,
        "blackhole": 6,
        ...
    }
    NM.IPRoutingRule.set_action(routing_rule, action_ids[r["action"]])
```

library/network_connections.py
```diff
@@ -1148,6 +1148,61 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                if r["dport_start"]:
```
thom311, line 1168: it seems ugly in the YAML to have fields `*_start` and `*_end`.

Have just a `dport|sport|uid` field, which can be a single number (integer or as string). Additionally, it can be a string in the form of `r'^ *([0-9]+) *- *([0-9]+) *$'`.

library/network_connections.py
```diff
@@ -1148,6 +1148,61 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                    # https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=14c0b97ddfc29
+                    action_ids = {
+                        "to-table": 1,
+                        "blackhole": 6,
+                        "prohibit": 7,
+                        "unreachable": 8,
+                    }
+                    for action in action_ids:
+                        if action == r["action"]:
+                            NM.IPRoutingRule.set_action(
```
thom311, line 1165: Drop the `for: if:` part, it serves no purpose.

library/network_connections.py
thom311: maybe it could also a two element list:
```
    - dport: 
      - 55
      - 66
```
but I think that
```
     - dport: "55-66"
```
is better API in the playbook.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +670,179 @@ def _validate_post(self, value, name, result):
...
+                    "fwmark", default_value=None, val_min=0, val_max=0xFFFFFFFF
+                ),
+                ArgValidatorNum(
+                    "fwmask", default_value=None, val_min=0, val_max=0xFFFFFFFF
+                ),
+                ArgValidatorStr("iif", default_value=None),
+                ArgValidatorBool("invert", default_value=None),
+                ArgValidatorNum("ipproto", default_value=None, val_min=0, val_max=255),
+                ArgValidatorStr("oif", default_value=None),
+                ArgValidatorRange("sport", default_value=0, val_min=0, val_max=65534),
```
thom311, line 714: usually, the `default_value` should be a value that is also detected by the validator... or a special sentinel like `None`.
For example, `ArgValidatorNum` should only return numbers of the requested `numeric_type` *plus*, as special case, `None` or sometimes `ArgValidator.DEFAULT_VALUE` or whatever makes sense.

I mean, the ArgValidator implementation allows you to return any default value. It seems ugly that this `ArgValidatorRange` (currently) either returns a dictionary `{ "range_start":..., "range_end":...}` or numeric `0`.

Also, `ArgValidatorRange.__init__()` has `default_value=ArgValidator.DEFAULT`. If all uses of that constructor need to choose a *different* default value, it's probably not a good default value.

module_utils/network_lsr/argument_validator.py
```diff
@@ -353,6 +353,43 @@ def _validate_impl(self, value, name):
...
+        ArgValidatorNum.__init__(self, name, required, default_value)
+        self.val_min = val_min
+        self.val_max = val_max
+        self.default_value = default_value
+        self.numeric_type = numeric_type
+
+    def _validate_impl(self, value, name):
+        if re.match(r"^ *([0-9]+) *- *([0-9]+) *$", value):
+
+            val_list = list(map(int, list(map(str.strip, value.split("-")))))
```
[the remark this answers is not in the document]
thom311: also, the previous implementation wouldn't check that the resulting list only contains two elements... What would it do about `1-2-3` ?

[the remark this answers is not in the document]
thom311: > also, the previous implementation wouldn't check that the resulting list only contains two elements... What would it do about `1-2-3` ?

ah ok. The regex above ensure that...

library/network_connections.py
ffmancera: This comment is not resolved yet

module_utils/network_lsr/argument_validator.py
```diff
@@ -353,6 +353,77 @@ def _validate_impl(self, value, name):
...
+                name,
+                "value is %s but cannot be greater then %s" % (value, self.val_max),
+            )
+        return v
+
+    def _validate_impl(self, value, name):
+        match_group = re.match(r"^ *([0-9]+) *- *([0-9]+) *$", value)
+        if match_group:
+            try:
+                val_list = list(map(int, [match_group.group(1), match_group.group(2)]))
```
thom311, line 407: seems hard to read, unless you fully bought into functional programming. And then, why not `list(map(int, map(match_group.group, [1, 2])))` or `list(map(lambda x: int(match_group.group(x)), [1, 2]))` ?

How about `val_list = (int(match_group.group(1)), int(match_group.group(2)))`

liangwen12year: Updated.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
+                ArgValidatorNum("to_len", default_value=0, val_min=0),
+                ArgValidatorNum("tos", default_value=0, val_max=255),
+                ArgValidatorRange("uid", val_min=0, val_max=0xFFFFFFFF),
+            ],
+            default_value=None,
+        )
+        self.family = family
+
+    def _validate_post(self, value, name, result):
+        result["family"] = (socket.AF_INET, socket.AF_INET6)[result["family"] == "ipv6"]
```
thom311, line 765: `(socket.AF_INET if (result["family"] == "ipv4") else socket.AF_INET6)`

also, it would seem convenient to detect the family based on to/from (if given). And, to/from must agree to have the same address family.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
+        if result["from"]:
+            network_from = result["from"]
+            family = network_from["family"]
+            result["from"] = network_from["address"]
+            result["family"] = family
```
thom311, line 770: this just overwrites the family (silently). That seems wrong.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
+
+        if result["from_len"] == 0:
+            if result["from"] is not None:
+                raise ValidationError(
+                    name,
+                    "the prefix length for from/src cannot be zero when from/src is specified",
+                )
+        elif result["from"] is None:
+            raise ValidationError(name, "missing from/src for a non zero prefix-length")
+        elif result["from_len"] > (32, 128)[result["family"] == socket.AF_INET6]:
```
thom311, line 792: is this `[a, b](boolean_expression)` really a desirable pattern? Seems not good to me. It's hard to read, and does it actually require python to create a list? Maybe python can optimize this away, but it's not clear that it can.

Instead, add a function `Util.addr_family_prefix_length(prefix)`, and also use it from `Util.addr_family_valid_prefix()`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
+
+        if result["fwmark"] is not None:
+            if result["fwmask"] is None:
+                raise ValidationError(
+                    name,
+                    "the fwmark requires fwmask to be set",
+                )
+
+        if result["fwmask"] is not None:
+            if result["fwmark"] is None:
```
thom311, line 814:
```python
if (result["fwmask"] is None) != (result["fwmark"] is None):
    raise ValidationError(name, "fwmask and fwmark must be set together")
```

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
+                raise ValidationError(
+                    name,
+                    "the outgoing interface '{0}' specified in the routing rule is invalid interface_name.".format(
+                        result["oif"]
+                    ),
+                )
+        if result["sport"]:
+            if (
+                result["sport"]["range_start"] == 0
+                and result["sport"]["range_end"] != 0
```
thom311, line 839: what is the meaning of this check? `ArgValidatorRange` should already sufficiently validate the range.

liangwen12year: NM validated this way, https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/libnm-core-impl/nm-setting-ip-config.c#L2622

module_utils/network_lsr/argument_validator.py
liangwen12year: Your proposal works, but I think that it more like the personal taste.

For the `suppress_prefixlength` setting, we need to know the IP family, that is the primary reason that I introduced `family` as an attribute.

module_utils/network_lsr/argument_validator.py
```diff
@@ -633,6 +704,177 @@ def _validate_post(self, value, name, result):
...
 
+class ArgValidatorIPRoutingRules(ArgValidatorDict):
+    def __init__(self, name, family=None, required=False):
+        ArgValidatorDict.__init__(
+            self,
+            name,
+            required,
+            nested=[
+                # a valid priority ranges from 0 to 0xFFFFFFFF, "-1" is also allowed
+                # to reset the priority
```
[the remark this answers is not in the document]
liangwen12year: I have not fully tested resetting the priority, I got hint from here, https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/libnm-core-impl/nm-setting-ip-config.c#L1812.

module_utils/network_lsr/argument_validator.py
thom311: > Your proposal works, but I think that it more like the personal taste.

Where did you see the pattern `[a, b](bool_val)`? While it works, it does not seem common to me (never saw it) and in my opinion, it's not a good pattern.

> For the suppress_prefixlength setting, we need to know the IP family, that is the primary reason that I introduced family as an attribute.

Of course you need to know the family. That means, at least one of `family`, `to`, `from` must be specified. If several of them are specified,then their family must agree.

This does not mean that `family` should be mandatory, if you can deduce it from `from`/`to`. Requiring the user to provide redundant information is cumbersome (for the user).

module_utils/network_lsr/argument_validator.py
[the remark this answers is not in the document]
thom311: `nm_ip_routing_rule_set_priority()` is a function of libnm. In libnm, `NMIPRoutingRule` is a relatively simple stucture that contains information about rules. It has setters and getters. Initially, when you create a new `NMIPRoutingRule` instance, it has no priority set. So there is a setter to set a priority, but the setter also allows to reset the unset priority.

Of course, a `NMIPRoutingRule` instance without a priority is not valid. The ability to reset the priority is merely for completeness, so that the setters allow you to create a `NMIPRoutingRule` in any configuration -- even an invalid one.

Having that part of the playbook API where users could create an invalid rule (without a priority) is not right. In fact, the priority is a mandatory parameter in the playbook, they an already create an invalid rule by omitting it.

module_utils/network_lsr/argument_validator.py
thom311: I mean that the constancy rules (validate) should be handled by where the error happens. That is, `ArgValidatorRange` needs to validate that the range is valid, and not `ArgValidatorIPRoutingRules`.

The result of `ArgValidatorRange` should be either "None" or a valid range (that is,one that satisfies the min/max values and the min<max rule).

`_rr_xport_range_valid()` does this check, because `NMIPRoutingRule` also needs to express to specify no range (`None`) or a valid range. Since `0` is not a valid port anyway (outside of the valid range), it uses that value to encode a missing range. Of course, setting `dport_start=0` && `dport_end!=0` would be meaningless/inconsistent, so that is rejected as invalid.

`ArgValidatorRange` encodes a missing range entirely differently (`None`).

thom311: See also `ArgValidatorRange("sport", val_min=0, val_max=65534)`.

The `val_min` of a port must be 1, not 0.
A missing port is encoded by not specifying the range (and `ArgValidatorRange` will represent that as `None`).

module_utils/network_lsr/argument_validator.py
thom311: > Where did you see the pattern [a, b](bool_val)? While it works, it does not seem common to me (never saw it) and in my opinion, it's not a good pattern.

i don't like python's `val1 if condition else val2` either. I think it's often hard to read, but this is the ternary operator in Python.

I often instead write
```
   if condition:
        val = val1
   else:
        val = val2
```
but that's not great too (verbose).

thom311 (approved):
very nice!!

README.md
```diff
@@ -501,6 +501,55 @@ The IP configuration supports the following options:
...
+      number means lower priority.
+   - `action` -
+      The action of the rule. The possible values are `to-table` (default),
+      `blackhole`, `prohibit`, `unreachable`.
+   - `dport`-
+      The range of the destination port. The range consists of start and end, which can
+      be specified in the form of `$start - $end`. The valid range value has to falls
+      within the range `0 - 65534`.
+   - `family` -
+      The IP family of the rule. The possible values are `ipv4`, `ipv6`.
```
tyll, line 521:
```suggestion
      The IP family of the rule. The possible values are `ipv4` and `ipv6`.
```

README.md
```diff
@@ -501,6 +501,55 @@ The IP configuration supports the following options:
...
+   - `priority` -
+      The priority of the rule. A valid priority ranges from 0 to 4294967295. Higher
+      number means lower priority.
+   - `action` -
+      The action of the rule. The possible values are `to-table` (default),
+      `blackhole`, `prohibit`, `unreachable`.
+   - `dport`-
+      The range of the destination port. The range consists of start and end, which can
+      be specified in the form of `$start - $end`. The valid range value has to falls
+      within the range `0 - 65534`.
```
tyll, line 519: The priority uses a different format for the range using "to" which looks better IMHO. Also using $start and $end does not seem to be the style that is used for the documentation here. Does the destination port always need to be a range?

Additionally, I mentioned already on the BZ regarding the acceptance criteria that it would be better IMHO not to use these abbreviations like `dport` to make it easier to read.

README.md
```diff
@@ -501,6 +501,55 @@ The IP configuration supports the following options:
...
+   - `from` -
+      The source address of the packet to match (e.g. `192.168.100.58/24`).
+   - `fwmark` -
+      The fwmark value of the packet to match.
+   - `fwmask` -
+      The fwmask value of the packet to match.
+   - `iif` -
+      Select the incoming interface name to match.
+   - `invert` -
+      Invert the selected match of the rule.
```
tyll, line 531: Please explain this in more detail. What is the value of this setting and what does it mean.

README.md
```diff
@@ -501,6 +501,55 @@ The IP configuration supports the following options:
...
+   - `oif` -
+      Select the outgoing interface name to match.
+   - `sport` -
+      The range of the source port. The range consists of start and end, which can
+      be specified in the form of `$start - $end`. The valid range value has to falls
+      within the range `0 - 65534`.
+   - `suppress_prefixlength` -
+      Reject routing decisions that have a prefix length of the specified or less.
+   - `table` -
+      The route table to look up for the `to-table` action.
```
tyll, line 543: What is the to-table action?

liangwen12year: The role currently supports the action `to-table`(default), `blackhole`, `prohibit`, `unreachable`.  

About the `to-table` action, I refered from kernel and NM, which basically stands for table lookup:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/fib_rules.h?id=f443e374ae131c168a065ea1748feac6b2e76613#n77
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/libnm-core-impl/nm-setting-ip-config.c#L2828

README.md
liangwen12year: > The priority uses a different format for the range using "to" which looks better IMHO. 
Do you mean to use `to` instead of `-` in between the range start and range end ?

liangwen12year: I changed into the following, is this better ?
```
   - `dport`-
      The range of the destination port. The range consists of start, hyphen and end.
      The valid range value has to falls within the range `0 - 65534`.
```

README.md
```diff
@@ -501,6 +501,56 @@ The IP configuration supports the following options:
...
+      `True` and `False` (default). If the value is `True`, this is equivalent to match
+      any packet that not satisfying selected match of the rule.
+   - `ipproto` -
+      Select the IP protocol value to match, the valid value ranges from 1 to 255.
+   - `oif` -
+      Select the outgoing interface name to match.
+   - `sport` -
+      The range of the source port. The range consists of start and end, which can
+      be specified in the form of `$start - $end`. The valid range value has to falls
+      within the range `0 - 65534`.
```
tyll, line 540: This is also using a different style to describe the range (not using "to").

liangwen12year:
The failure in rhel-7 is an IT problem, which affects the "download.devel.redhat.com"

README.md
```diff
@@ -501,6 +501,57 @@ The IP configuration supports the following options:
...
+   - `priority` -
+      The priority of the rule. A valid priority ranges from 0 to 4294967295. Higher
+      number means lower priority.
+   - `action` -
+      The action of the rule. The possible values are `to-table` (default),
+      `blackhole`, `prohibit`, `unreachable`.
+   - `dport`-
+      The range of the destination port. The range consists of start, hyphen and end. A
+      valid dport value for both start and end ranges from 0 to 65534. And the start
+      cannot be greater than the end.
```
tyll, line 519: It seems that I was not clear initially. Using the dollar sign seemed odd to me. For IP addresses, you provide an explizit example, so that would be better here as well instead of the "start, hyphen, and end" description, IMHO. Also, does this always need to be a range or is a single value ok, too?

liangwen12year: Single value (int/string ) is also ok, but it will be normalized into the range internally, I am hesitating to mention the single value support in the readme because it might confuse the user IMHO.

liangwen12year: Is this better ?

ffmancera (approved):
LGTM!

PR #493, opened 2022-04-11
infiniband: Add the integration tests

liangwen12year:


liangwen12year:
This PR is not ready for reviewing.

liangwen12year:
Note: the interface name for the virtual infiniband device is different between nm and initscripts provider. 

nm provider:
```
"fatal error: created connection failed to normalize: nm-connection-error-quark: connection.interface-name: interface name of software infiniband device must be 'ib0.000a' or unset (instead it is 'ib0.800a')
```

initsripts provider: 
```
++	# Normalize our PKEY_ID to have the high bit set
++	NEW_PKEY_ID=`printf "0x%04x" $(( 0x8000 | ${PKEY_ID} ))`
++	NEW_PKEY_NAME=`printf "%04x" ${NEW_PKEY_ID}`
++	[ "${DEVICE}" != "${PHYSDEV}.${NEW_PKEY_NAME}" ] && {
++                net_log $"Configured DEVICE name does not match what new device name would be.  This
++is most likely because once the PKEY_ID was normalized, it no longer
++resulted in the expected device naming, and so the DEVICE entry in the
++config file needs to be updated to match. This can also be caused by
++giving PKEY_ID as a hex number but without using the mandatory 0x prefix.
++	Configured DEVICE=$DEVICE
++        Configured PHYSDEV=$PHYSDEV
++        Configured PKEY_ID=$PKEY_ID
++        Calculated PKEY_ID=$NEW_PKEY_ID
++        Calculated name=${PHYSDEV}.${NEW_PKEY_NAME}"
++                exit 1
++        }
```

tests/playbooks/tests_infiniband.yml
```diff
@@ -0,0 +1,160 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- hosts: all
```
richm, line 3: add
```yaml
  tags:
    - tests::infiniband
```
so that we can skip this test in CI

tests/playbooks/tests_infiniband.yml
```diff
@@ -0,0 +1,160 @@
...
+            - command: ip link del ib0.000a
+              ignore_errors: true
+              when: network_provider == 'nm'
+            - command: ip link del ib0.800a
+              ignore_errors: true
+              when: network_provider == 'initscripts'
+          tags:
+            - "tests::cleanup"
+  tags:
+    - "tests::expfail"
```
richm, line 160: On second thought, I think the tag should be explicit - so use `tests::infiniband` instead so that we know the purpose of the tag

liangwen12year: Updated. Thanks.

tests/playbooks/tests_infiniband.yml
```diff
@@ -0,0 +1,160 @@
...
+        msg: "##################################################"
+    - block:
+        # For initscripts provider, the script
+        # `/etc/sysconfig/network-scripts/ifup-ib` is required in order to
+        # configure the infiniband profiles
+        - name: Validate the script 'ifup-ib' existed
+          command: test -x /etc/sysconfig/network-scripts/ifup-ib
+          register: ifup_ib_exist
+          failed_when: ifup_ib_exist.rc == 1
+          when: network_provider == "initscripts"
```
tyll, line 16: Where does this script come from?

tests/playbooks/tests_infiniband.yml
```diff
@@ -0,0 +1,160 @@
...
+                  transport_mode: datagram
+                parent: ib0
+                state: up
+                ip:
+                  dhcp4: no
+                  auto6: no
+                  address:
+                    - 198.51.100.133/30
+                    - 2001:db8::2/32
+          when: network_provider == 'nm'
```
tyll, line 44: What is the reason to have different network_connections depending on the provider?

tests/playbooks/tests_infiniband.yml
liangwen12year: https://raw.githubusercontent.com/Mellanox/rdma-utils/master/rdma.ifup-ib

Actually it was included in rhel-7.9 and dropped in rhel-8.0. 
https://pkgs.devel.redhat.com/cgit/rpms/rdma-core/commit/?h=rhel-7.9&id=697c706324f2c866d7b855682a8a3c80371a6651
https://pkgs.devel.redhat.com/cgit/rpms/rdma-core/commit/?h=rhel-8.0&id=1fbba1120742c5446ffeed9cb8bb91542b8ea317

tests/playbooks/tests_infiniband.yml
liangwen12year: I simplied the code to only have one network_connections variable. And you may notice that the ipoib interface name changed and pkey value changed when using different network provider, it is determined by how initscripts and NM backend would activate the connection. 

Reference;
1. https://github.com/linux-system-roles/network/pull/493#issuecomment-1101554399 
2. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ipoib

tyll:
> Note: the interface name for the virtual infiniband device is different between nm and initscripts provider.

Can you please find out, why this is the case and why this only accepts a certain value for each provider (what is the point in making this a setting/variable in this case) and also document this in the README? Infiniband does not seem to be documented at all.

liangwen12year:
> > Note: the interface name for the virtual infiniband device is different between nm and initscripts provider.
> 
> Can you please find out, why this is the case and why this only accepts a certain value for each provider (what is the point in making this a setting/variable in this case) and also document this in the README? Infiniband does not seem to be documented at all.

I did further investigation, the only difference is that Initscripts will always normalize the pkey internally to set up the full membership bit (e.g. 0x800a, the number 8 here is the full membership bit), 
```extracted from rdma.ifup-ib
	# Normalize our PKEY_ID to have the high bit set
	NEW_PKEY_ID=`printf "0x%04x" $(( 0x8000 | ${PKEY_ID} ))`
```
but NM will not normalize the pkey value to set up the full membership bit. So Ideally, pkey value defined via the role can be the same for both provider, Initscripts provider will normalize it internally though. 

And in the next step, both provider will use the pkey value to calculate the interface name, the way that the interface name is calculated is the same for both providers, but since the initscripts provider will use the normalized the pkey value (full membership bit set), as a result, interface name will vary for two providers.  

Based on the discussion with @thom311 , nobody is aware of what initscripts does when implementing setting the pkey value and interface name in NM.

tests/playbooks/tests_infiniband.yml
liangwen12year: I will address this.

liangwen12year: Yes. I can limit the test for initscripts provider in only RHEL 7. But there is a possibility that copying the `/etc/sysconfig/network-scripts/ifup-ib` from the RHEL-7 machine to other RHEL os version machine prior to the testing.

tests/playbooks/tests_infiniband.yml
```diff
@@ -0,0 +1,104 @@
...
+  tasks:
+    - name: "INIT Prepare setup"
+      debug:
+        msg: "##################################################"
+    - block:
+        # For initscripts provider, the script
+        # `/etc/sysconfig/network-scripts/ifup-ib` is required in order to
+        # configure the infiniband profiles, and only the RHEL 7 rdma-core
+        # package contains such script
+        - name: Install the rdma-core package
```
liangwen12year, line 16: I will install it at the role level.

README.md
```diff
@@ -279,6 +279,31 @@ role.
 Similar to `controller` and `vlan`, the `parent` references the connection profile in
 the ansible role.
 
+#### `type: infiniband`
+
+For the infiniband connection, the following options are supported:
+
+- `p_key`: The infiniband P_Key to use for the device. The default value of `-1` means
```
tyll, line 286: In the other PR it seems that `-1` is not intended to be set by the user but just an internal default value. What is correct? Also, it seemed to me that it is mandatory to set this, is it not? This is confusing.

README.md
```diff
@@ -279,6 +279,31 @@ role.
...
+**Note:** For the `p_key` setting, initscripts provider will normalize the `p_key`
+  internally to always set up the full membership bit (e.g. 0x000a will be normalized
+  into 0x800a) while nm provider does not. And both providers calculate the interface
+  name exactly the same way - based on the `parent` interface name and the P_Key, as a
+  result, `interface_name` will vary for two providers. (e.g. if the parent
+  `interface_name` is `ib0` and user defined `p_key` is 0x000a, the `interface_name`
+  for nm provider should be `ib0.000a`and the `interface_name` for initscripts provider
+  should be `ib0.800a`)
+
+**Note:** When using the initscripts provider, the script
```
tyll, line 301: Just wondering, since the other code is fixing infiniband support. Does it mean that it never worked with the network role before? In that case, it seems feasible to only support it with the nm provider from now on.

liangwen12year: Correct,  the ipoib never worked before

README.md
liangwen12year: https://www.kernel.org/doc/Documentation/infiniband/ipoib.txt, the kernel did not mention `-1` as the internal default value

But NM uses `-1` means to use the  default P_Key (aka "the P_Key at index 0")
```
=== [p-key] ===
[NM property description]
The InfiniBand P_Key to use for this device. A value of -1 means to use the default P_Key (aka "the P_Key at index 0"). Otherwise, it is a 16-bit unsigned integer, whose high bit is set if it is a "full membership" P_Key.
```

liangwen12year: Yes, it is mandatory to set, if user does not set the pkey explicitly, the default value `-1` will be choose for the role which  corresponds to the default P_Key in kernel (aka "the P_Key at index 0")

liangwen12year: maybe I should change the `unset` into `not explicitly set`

README.md
liangwen12year: > Just wondering, since the other code is fixing infiniband support. Does it mean that it never worked with the network role before? In that case, it seems feasible to only support it with the nm provider from now on.

Did you mean to drop the infiniband tests for the initscripts provider?

tyll:
> Based on the discussion with @thom311 , nobody is aware of what initscripts does when implementing setting the pkey value and interface name in NM.

What do you mean with nobody? At least the network-scripts developers or some users should know what initscripts is doing.

liangwen12year:
> > Based on the discussion with @thom311 , nobody is aware of what initscripts does when implementing setting the pkey value and interface name in NM.
> 
> What do you mean with nobody? At least the network-scripts developers or some users should know what initscripts is doing.

I meant that NM and initscripts implemented setting the pkey and interface name differently because the time NM implemented it, nobody in NM team awared how initscripts implement it.

README.md
liangwen12year: For configuring the network connection on the physical infiniband interface, it worked before. The ipoib (IP OVER INFINIBAND) is creating the connection on the virtual infiniband interface, it did not work before.

README.md
liangwen12year: Well, then pkey is not mandatory, if the user does not set the pkey explicitly (to a value different than -1), then the user creating the connection on the physical infiniband interface (not ipoib mode), in such situation,  https://github.com/linux-system-roles/network/pull/495/files#diff-6b3f5f6149ac1330c5e2c832286b63b3597158a998f2244ddbec2045ace1f794R2022 will raise an error when missing the physical infiniband interface.

README.md
tyll: Is the rdma package required for physical infiniband support, too?

liangwen12year: Yes, as long as `TYPE=Infiniband` specified in the ifcfg file, then ifup-ib (rdma package) is needed to support setting up the connection. Then I will drop the support for initscripts provider. Thanks.

liangwen12year:
The following is just for the record which might be proper to put in this PR. 

**Note:** For the `p_key` setting, initscripts provider will normalize the `p_key`
  internally to always set up the full membership bit (e.g. 0x000a will be normalized
  into 0x800a) while nm provider does not. And both providers calculate the interface
  name exactly the same way - based on the `parent` interface name and the P_Key, as a
  result, `interface_name` will vary for two providers. (e.g. if the parent
  `interface_name` is `ib0` and user defined `p_key` is 0x000a, the `interface_name`
  for nm provider should be `ib0.000a`and the `interface_name` for initscripts provider
  should be `ib0.800a`)

**Note:** When using the initscripts provider, the script
  `/etc/sysconfig/network-scripts/ifup-ib` is required to configure the infiniband
  connection, only the RHEL 6 `rdma` package (shipped in RHEL6 by default) and the
  RHEL 7 `rdma-core` package contain such script, therefore, setting up the infiniband
  connection will only work in RHEL 6 and RHEL 7.

README.md
tyll: > Well, then pkey is not mandatory, if the user does not set the pkey explicitly (to a value different than -1), then the user creating the connection on the physical infiniband interface (not ipoib mode)

Ok, so if the `p_key` setting also controls the infiniband mode (physical vs. ipoib), then this should be mentioned in the documentation. Also, the documentation mentions that it would use the P_Key at index 0, so what happens when  p_key is set to 0 by the user? Based on the descriptions so far it seems that this also implies physical mode, since p_key -1 means 0 means physical mode? The code does not check for p_key == 0, though.

README.md
```diff
@@ -279,6 +279,31 @@ role.
...
+  possible modes are `datagram` and `connected`.
+
+**Note:** For the `p_key` setting, initscripts provider will normalize the `p_key`
+  internally to always set up the full membership bit (e.g. 0x000a will be normalized
+  into 0x800a) while nm provider does not. And both providers calculate the interface
+  name exactly the same way - based on the `parent` interface name and the P_Key, as a
+  result, `interface_name` will vary for two providers. (e.g. if the parent
+  `interface_name` is `ib0` and user defined `p_key` is 0x000a, the `interface_name`
+  for nm provider should be `ib0.000a`and the `interface_name` for initscripts provider
+  should be `ib0.800a`)
```
tyll, line 299: If the interface name is determined by the p_key, is there any reason for the user to specify both the p_key and the interface_name for infiniband interfaces if the mode is ipoib? Would it make sense for the role to calculate the interface_name instead?

README.md
tyll: If phyiscal infiniband interfaces work with RHEL 7, then we cannnot drop support for that IMHO.

liangwen12year: But rdma-core package is not installed by default for initscripts provider in RHEL-7 though, it would not work for REHL 7 unless the users manually install the rdma-core package in the managed host themselves.

README.md
liangwen12year: That is a very good point, we can omit the interface name as a setting and calculate the interface name internally.

README.md
liangwen12year: Well, if the user set the pkey to 0, then ipoib will be created. I referred from "There are special requirements for the names of IPoIB P_Key interfaces. All IPoIB P_Keys range from 0x0000 to 0x7fff, and the high bit, 0x8000, denotes that membership in a P_Key is full membership instead of partial membership." [1] 

For "use the P_Key at index 0", I would interprete as using the base infiniband fabric, and for the pkey value greater or equal to zero,  I would interprete as creating the partition or virtual subnet.  

1. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ipoib

liangwen12year: README updated.

README.md
```diff
@@ -279,6 +279,31 @@ role.
...
 
+#### `type: infiniband`
+
+For the infiniband connection, the following options are supported:
+
+- `p_key`: The infiniband P_Key to use for the device. The default value of `-1` means
+  to use the P_Key at the index 0 (unset the p_key). Otherwise, it is a 16-bit unsigned
+  integer and the high bit should be set if it is a "full membership" P_Key.
+- `transport_mode`: The ipoib (IP OVER INFINIBAND) connection operation mode. The
+  possible modes are `datagram` and `connected`.
```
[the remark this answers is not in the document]
liangwen12year: There is a default value in the role, which is `datagram`. 

I got the clue from:
1. https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L2091
2. https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1570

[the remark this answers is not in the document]
liangwen12year: There is a default value in the role, which is `datagram`. 

I got the clue from:
1. https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L2091
2. https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/argument_validator.py#L1570

README.md
liangwen12year: A new commit was added.

README.md
```diff
@@ -279,6 +279,24 @@ role.
...
+#### `type: infiniband`
+
+For the infiniband connection, the following options are supported:
+
+- `p_key`: The infiniband P_Key to use for the device. The default value of `-1` means
+  to use the P_Key at the index 0 and the connection will be created on the physical
+  infiniband fabrics. Otherwise, it is a 16-bit unsigned integer and the ipoib (IP OVER
+  INFINIBAND) connection will be created, the high bit should be set if it is a
+  "full membership" P_Key.
+- `transport_mode`: The ipoib (IP OVER INFINIBAND) connection operation mode. The
```
tyll, line 291: Pleas don't use all caps but "IP over InfiniBand" (or however this is written)

module_utils/network_lsr/argument_validator.py
```diff
@@ -2097,6 +2097,18 @@ def _validate_post(self, value, name, result):
                             "a infiniband device with 'infiniband.p_key' "
                             "property also needs 'mac' or 'parent' property",
                         )
+                    if "parent" in result:
+                        if "interface_name" in result:
+                            raise ValidationError(
+                                name + ".interface_name",
+                                "the 'interface_name' does not need to be specified when "
+                                "'infiniband.p_key' is not '-1' and the 'parent' exists",
+                            )
```
tyll, line 2106: This error is only ok, if InfiniBand support was broken in the past and there are no playbooks out there that would fail now.

README.md
```diff
@@ -279,6 +279,24 @@ role.
 Similar to `controller` and `vlan`, the `parent` references the connection profile in
 the ansible role.
 
+#### `type: infiniband`
```
tyll, line 282: Mention, that this is only supported for the NM provider.

module_utils/network_lsr/argument_validator.py
liangwen12year: Yes, the ipoib support was broken before, so it is safe to raise error here.

README.md
liangwen12year: Updated.

README.md
liangwen12year: Sure.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2097,6 +2097,18 @@ def _validate_post(self, value, name, result):
...
+                        result["interface_name"] = "%s.%04x" % (
```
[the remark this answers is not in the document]
liangwen12year: In NM, users are not required to specify the interface name for ipoib connection, but user can specify it, but the specified interface name will  exactly match "%s.%04x". I would suggest doing the same for the role.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2292,14 +2296,34 @@ def _validate_post(self, value, name, result):
...
+                    if (
+                        connection["type"] == "infiniband"
+                        and connection["infiniband"]["p_key"] != -1
+                        and connection["interface_name"]
+                    ):
+                        virtual_interface_name = "{0}.{1:0>4x}".format(
+                            parent_connection["interface_name"],
+                            connection["infiniband"]["p_key"],
+                        )
+                        if connection["interface_name"] != virtual_interface_name:
```
thom311, line 2337: I tighten up the check in https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1223, so an interface-name is only allowed, if also a parent is allowed (mac-address alone is not enough).

In other words, if you didn't set parent (for a virtual infiniband profile), then the interface-name is also rejected.

Maybe this kind of "normalization" and checking is just pointless. The user is in fact not free to set an arbitary interface-name, it would seem easier to just reject any.

module_utils/network_lsr/argument_validator.py
```diff
@@ -1553,6 +1553,8 @@ def get_default_bond(self):
 
 
 class ArgValidator_DictInfiniband(ArgValidatorDict):
+    DEFAULT_PKEY = -1
```
thom311, line 1556: as said before, when talking about the playbook/YAML syntax, the physical interface should primarily be represented by not specifying the `p_key` property (only accept `-1` as alternative).

ArgValidator takes the playbook, and normalizes it to something that is most useful for internal use. It does not necessarily need to use the same representation as in yaml. Rather, what is useful. As such, if the value is missing in YAML, ArgValidator might normalize that to `-1`. Or it might normalize it to None. I don't see any advantage of using -1 here. In particular, if you now feel the urge to add a DEFAULT_PKEY alias only to represent an unset value. We already have a well-known, commonly used value to represent missing values: `None`.

One argument could be, tha if you use -1, then you could always call
```
                 s_infiniband.set_property(
                     NM.SETTING_INFINIBAND_P_KEY, connection["infiniband"]["p_key"]
                 )
```
(which is true, as far as libnm is concerned). But the code doesn't even make use of that. It explicitly checks:
```
            if (
                connection["infiniband"]["p_key"]
                != ArgValidator_DictInfiniband.DEFAULT_PKEY
            ):
                s_infiniband.set_property(
                    NM.SETTING_INFINIBAND_P_KEY, connection["infiniband"]["p_key"]
                )
```


Only because libnm does something, it does not mean that this is also the best representation of the setting -- neither in playbook/YAML nor for ArgValidator.   Usually yes (because what libnm does often makes sense already, and because we might be familiar with that), but not always.

thom311: if you were to leave it at `-1`, I wouldn't say anything. After all, I added that (without thinking properly).
But now building on top of that wrong choice, by adding `ArgValidator_DictInfiniband.DEFAULT_PKEY`, seems not good.

thom311: also, this commit fixes one actual bug (the `!=` vs. `==`). It's a one-line fix (even a one-character fix).

Instead, this commit is +25/-5 lines of unrelated change. The bugfix is important and it's almost impossible to see, please create separate commits.

thom311 (changes requested):

module_utils/network_lsr/argument_validator.py
```diff
@@ -2292,14 +2296,34 @@ def _validate_post(self, value, name, result):
...
                         raise ValidationError(
                             name + "[" + str(idx) + "].parent",
                             "references non-existing 'parent' connection '%s'"
                             % (connection["parent"]),
                         )
+
+                    if (
+                        connection["type"] == "infiniband"
+                        and connection["infiniband"]["p_key"] != -1
+                        and connection["interface_name"]
```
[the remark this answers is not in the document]
liangwen12year: After calling `ArgValidator_DictConnection._validate_post()`, the `interface_name` will be assigned to `None` when not specfied.

module_utils/network_lsr/argument_validator.py
liangwen12year: As you mentioned, the normalization and checking in the NM are useless,  maybe the NM could have rejected the interface name set by user initially for the ipoib connection. Yes, currently, the simpliest approach in the role would be always rejecting the interface name as invalid for the ipoib connection.

module_utils/network_lsr/argument_validator.py
liangwen12year: > as said before, when talking about the playbook/YAML syntax, the physical interface should primarily be represented by not specifying the `p_key` property (only accept `-1` as alternative).
> 
> ArgValidator takes the playbook, and normalizes it to something that is most useful for internal use. It does not necessarily need to use the same representation as in yaml. Rather, what is useful. As such, if the value is missing in YAML, ArgValidator might normalize that to `-1`. Or it might normalize it to None. I don't see any advantage of using -1 here. In particular, if you now feel the urge to add a DEFAULT_PKEY alias only to represent an unset value. We already have a well-known, commonly used value to represent missing values: `None`.
> 
> One argument could be, tha if you use -1, then you could always call
> 
> ```
>                  s_infiniband.set_property(
>                      NM.SETTING_INFINIBAND_P_KEY, connection["infiniband"]["p_key"]
>                  )
> ```
> 
> (which is true, as far as libnm is concerned). But the code doesn't even make use of that. It explicitly checks:
> 
> ```
>             if (
>                 connection["infiniband"]["p_key"]
>                 != ArgValidator_DictInfiniband.DEFAULT_PKEY
>             ):
>                 s_infiniband.set_property(
>                     NM.SETTING_INFINIBAND_P_KEY, connection["infiniband"]["p_key"]
>                 )
> ```
> 
> Only because libnm does something, it does not mean that this is also the best representation of the setting -- neither in playbook/YAML nor for ArgValidator. Usually yes (because what libnm does often makes sense already, and because we might be familiar with that), but not always.

Thanks for the explanation, I will change the default value of `p_key` into  `None`.

richm:
[citest bad]

module_utils/network_lsr/argument_validator.py
thom311: > Yes, currently, the simpliest approach in the role would be always rejecting the interface name as invalid for the ipoib connection.

The last version does not longer reject it (the validate() checks), or does it? Where does it happen? Despite the doc saying `**Note:** If the `p_key` is specified , then the `interface_name` must be unset.` and that libnm would really not like it.

thom311: ah ok, `"the 'interface_name' must be unset for the ipoib "` . Never mind.

module_utils/network_lsr/argument_validator.py
thom311: > also, this commit fixes one actual bug (the != vs. ==). It's a one-line fix (even a one-character fix).

Is this fix done? Please make it a separate commit.

liangwen12year: I changed the order of the commits.

thom311 (approved):

ffmancera (approved):

PR #494, opened 2022-04-12
no gather facts setup snapshot

richm:
- support playbooks which use gather_facts: false
- add setup-snapshot.yml used to prepare testing snapshots

### DIFF
--- a/tests/setup-snapshot.yml
+++ b/tests/setup-snapshot.yml
@@ -1,0 +1,12 @@
+- hosts: all
+  tasks:
+    - name: Set facts used by role
+      include_role:
+        name: linux-system-roles.network
+        tasks_from: set_facts.yml
+        public: true
+
+    - name: Install test packages
+      package:
+        name: "{{ network_packages }}"
+        state: present
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -121,3 +121,11 @@ network_packages: "{{
     __network_provider_setup[network_provider]['packages'] }}"
 network_service_name: "{{
     __network_provider_setup[network_provider]['service_name'] }}"
+
+# ansible_facts required by the role
+__network_required_facts:
+  - distribution
+  - distribution_major_version
+  - distribution_version
+  - os_family
+  - python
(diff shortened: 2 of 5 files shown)


richm:
See https://github.com/redhat-cop/automation-good-practices/pull/62 for the rationale behind these changes.

richm:
[citest pending]

tests/setup-snapshot.yml
```diff
@@ -0,0 +1,12 @@
+- hosts: all
```
tyll, line 1: What is the purpose of this file? It does not seem to be used in this PR.

richm: > What is the purpose of this file? It does not seem to be used in this PR.

It is used by CI for test performance purposes.  The first time CI runs a test for the role for a particular platform, it will create a snapshot image which will be pre-populated with the latest repo metadata and packages, and will have the epel repo installed (but disabled - the role tests will need to enable epel when needed).  If the role provides the file tests/setup-snapshot.yml, then this will also be run to prepare the snapshot.  One of the primary things this file does is to install packages which will be required by CI, because package installation is usually one of the most time consuming tasks.  And because we don't want to duplicate the list of packages in this file *and* in the role somewhere, the role has been somewhat refactored so that we can use the list of packages defined by the role in setup-snapshot.yml by moving the code that establishes the list of packages to `tasks/set_facts.yml`, then calling this file from setup-snapshot.yml by using include_role with `tasks_from: set_facts.yml`.  CI will run the test by adding the arguments `--use-snapshot` and `--setup-yml tests/setup-snapshot.yml`

https://github.com/linux-system-roles/tox-lsr#qemu-testing

liangwen12year:
LGTM.

liangwen12year (dismissed):

nhosoi (approved):
lgtm

tyll (approved):
:+1:

richm:
centos-7/ansible-2.9 failure due to transient network issue - will retest

richm:
[citest bad]
still waiting for the ansible-2.12 results - the storage PR is ahead of it in the queue

richm:
[citest bad]

richm:
[citest bad]

PR #495, opened 2022-04-29
infiniband: Fix the bug of wrongly checking whether the device exists

liangwen12year:
When activating the connection on the virtual infiniband device, the
device does not need to be present upfront. The previous implementation
is wrong which validating the presence of virtual infiniband device
upfront, as a result, the users have to create the virtual infiniband
device manually in the target system in order to activating the
connection on the virtual infiniband device through the role. This is
an actual bug, the exception raised when the users do not
pre-configured such a virtual infiniband device. To address the
problem, fix the bug of wrongly checking whether the device exists.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -44,6 +44,7 @@ from ansible.module_utils.network_lsr.myerror import MyError  # noqa:E501
 
 from ansible.module_utils.network_lsr.argument_validator import (  # noqa:E501
     ArgUtil,
+    ArgValidator_DictInfiniband,
     ArgValidator_ListConnections,
     ValidationError,
 )
@@ -348,7 +349,10 @@ class IfcfgUtil:
                 if (connection["infiniband"]["transport_mode"] == "connected")
                 else "no"
             )
-            if connection["infiniband"]["p_key"] != -1:
+            if (
+                connection["infiniband"]["p_key"]
+                != ArgValidator_DictInfiniband.DEFAULT_PKEY
+            ):
                 ifcfg["PKEY"] = "yes"
                 ifcfg["PKEY_ID"] = str(connection["infiniband"]["p_key"])
                 if connection["parent"]:
@@ -846,7 +850,10 @@ class NMUtil:
                 NM.SETTING_INFINIBAND_TRANSPORT_MODE,
                 connection["infiniband"]["transport_mode"],
             )
-            if connection["infiniband"]["p_key"] != -1:
+            if (
+                connection["infiniband"]["p_key"]
+                != ArgValidator_DictInfiniband.DEFAULT_PKEY
+            ):
                 s_infiniband.set_property(
                     NM.SETTING_INFINIBAND_P_KEY, connection["infiniband"]["p_key"]
                 )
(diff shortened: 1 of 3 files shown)


library/network_connections.py
```diff
@@ -2012,7 +2012,7 @@ def run_prepare(self):
                                 "interface exists" % (connection["interface_name"]),
                             )
                         elif connection["type"] == "infiniband":
-                            if connection["infiniband"]["p_key"] != -1:
+                            if connection["infiniband"]["p_key"] == -1:
```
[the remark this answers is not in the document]
liangwen12year: Well. `-1` basically means that `p_key` is unset and the physical infiniband device is required for to be present for creating the connection on it. I will update the code. 

```
ArgValidatorNum("p_key", val_min=-1, val_max=0xFFFF, default_value=-1)
```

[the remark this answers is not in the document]
liangwen12year: And the reason that I separate this commit into a PR is because it may break the existing test playbooks.

library/network_connections.py
```diff
@@ -2012,7 +2012,12 @@ def run_prepare(self):
                                 "interface exists" % (connection["interface_name"]),
                             )
                         elif connection["type"] == "infiniband":
-                            if connection["infiniband"]["p_key"] != -1:
+                            # The default `p_key` value is -1, which means that the
+                            # `p_key` is unset. In this case, the physical infiniband
+                            # device is required to be present for creating the
+                            # connection on it
+                            default_pkey_val = -1
+                            if connection["infiniband"]["p_key"] == default_pkey_val:
```
tyll, line 2020: IMHO the constant should be used in all place where `-1` is used for `p_key`, so also in the place where the default value is assigned. This is to make sure that it is consistently used everywhere.

liangwen12year: I updated, thanks.

library/network_connections.py
```diff
@@ -348,7 +349,10 @@ def ifcfg_create(
                 if (connection["infiniband"]["transport_mode"] == "connected")
                 else "no"
             )
-            if connection["infiniband"]["p_key"] != -1:
+            if (
+                connection["infiniband"]["p_key"]
+                != ArgValidator_DictInfiniband.DEFAULT_PKEY_VAL
```
tyll, line 354:
```suggestion
                != ArgValidator_DictInfiniband.DEFAULT_PKEY
```

The `_VAL` suffix is not needed. IMHO it is generally clear that a constant represents a value (assuming that VAL is the abbreviation of value).

module_utils/network_lsr/argument_validator.py
```diff
@@ -1553,6 +1553,9 @@ def get_default_bond(self):
 
 
 class ArgValidator_DictInfiniband(ArgValidatorDict):
+    # The default `p_key` value is -1, which means that the `p_key` is unset
```
tyll, line 1556: IMHO the comment is not necessary since it basically describes the meaning of that is deductible from the constant's name.

library/network_connections.py
liangwen12year: Changed.

module_utils/network_lsr/argument_validator.py
liangwen12year: The comment dropped. Thanks.

library/network_connections.py
```diff
@@ -2012,7 +2019,13 @@ def run_prepare(self):
                                 "interface exists" % (connection["interface_name"]),
                             )
                         elif connection["type"] == "infiniband":
-                            if connection["infiniband"]["p_key"] != -1:
+                            # When `p_key` is unset (default), the physical infiniband
+                            # device is required to be present for creating the
+                            # connection on it
```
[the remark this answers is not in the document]
liangwen12year: Ok, I will bear this in mind, thanks

liangwen12year:
The commit in this PR has been merged into https://github.com/linux-system-roles/network/pull/493, let us close this PR.

PR #504, opened 2022-05-31
[Draft] Support the nmstate network state configuration

liangwen12year:
The users want to apply the nmstate network state configuration to the
interface directly through the role, which necessitates the less
complexity of the network configuration and allows the partial
configuration on the network.

To warrant that the users are capable to apply the nmstate network state
configuration, add the support for the `network_state` variable.


liangwen12year:
References:
1. https://github.com/nmstate/ansible-nmstate/blob/main/library/nmstate_apply.py

tests/playbooks/tests_network_state.yml
```diff
@@ -0,0 +1,45 @@
...
+              type: ethernet
+              state: up
+              ipv4:
+                address:
+                  - ip: 192.0.2.251
+                    prefix-length: 30
+                dhcp: false
+          routes:
+            config:
+            - destination: 192.0.2.100/30
```
richm, line 39: https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#yaml-and-jinja2-syntax
"Indent list contents beyond the list definition"

richm:
will need to document this in the README and examples/

richm:
```
************* Module library.network_state
E:  7, 0: Unable to import 'libnmstate' (import-error)
ERROR: InvocationError for command /usr/bin/bash /opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/tox_lsr/test_scripts/runpylint.sh --errors-only --rcfile /opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/tox_lsr/config_files/pylintrc (exited with code 2)
```
need to add a dependency on `libnmstate` to pylint_extra_requirements.yml and/or pytest_extra_requirements.yml

.github/run_test.sh
```diff
@@ -45,6 +45,9 @@ EXCLUDE_TESTS_C8S='
 # exclude tests/tests_wireless_wpa3_owe_nm.yml and tests/tests_wireless_wpa3_sae_nm.yml
 # since failing to install mac80211_hwsim kernel module
 # exclude tests/tests_infiniband_nm.yml since missing the infiniband device
+# the `nmstate-devel` package is shipped in CRB repo, after running `dnf upgrade -y`,
+# the CRB repo will be removed and therefore the `nmstate-devel` package will not be
+# avaiable
```
ffmancera, line 50: This should be possible.. I think it should be possible to add the repo. I think it is important for us to have the tests in C9S too.

README.md
```diff
@@ -91,6 +101,20 @@ network_connections:
...
+  interfaces:
+    - name: eth0
+    #...
+  routes:
+    config:
+      #...
+  dns-resolver:
+    config:
+      #...
+```
```
ffmancera, line 116: As this is an example.. should we use complete states instead of using `#...`? What do you think?

library/network_state.py
```diff
@@ -0,0 +1,74 @@
...
+        required: true
+        type: dict
+author: "Wen Liang (@liangwen12year)"
+"""
+
+RETURN = r"""
+state:
+    description: Network state after running the module
+    type: dict
+    returned: always
```
ffmancera, line 29: I noticed it is returned always.. but what would it return if Nmstate fails?

library/network_state.py
```diff
@@ -0,0 +1,74 @@
...
+            self.result["changed"] = True
+
+        self.result["state"] = current_state
+
+        self.module.exit_json(**self.result)
+
+
+def run_module():
+    module_args = dict(
+        state=dict(type="dict", required=True),
```
ffmancera, line 58: I think a better name would be `desired_state` because is what the user is specifying. `state` seems a little bit ambiguous..

library/network_state.py
```diff
@@ -0,0 +1,74 @@
...
+DOCUMENTATION = r"""
+---
+module: network_state
+version_added: "2.9"
+short_description: module for network role to apply network state configuration
+description:
+    - This module allows to apply the network state configuration through nmstate,
+      https://github.com/nmstate/nmstate
+options:
+    state:
```
ffmancera, line 18: what do you think about calling this `current_state`? So it is clear it is the result of the changes..

tasks/main.yml
```diff
@@ -19,6 +35,26 @@
...
+  package:
+    name:
+      - nmstate-devel
+    state: present
+  when:
+    - network_state is defined
+    - ansible_distribution == 'Fedora' and
+      ansible_distribution_major_version | int > 34 or
+      ansible_distribution != 'Fedora' and
+      ansible_distribution_major_version | int > 8
```
ffmancera, line 56: Shouldn´t this install NetworkManager too? As the test above is doing? Also the check on nmstate is also nice to have.

ffmancera (changes requested):
Nice work! I made some comments and will re-review again! It is a really good draft PR! :-)

README.md
liangwen12year: If you see the few lines before it, 
```yaml
network_provider: nm
network_connections:
  - name: eth0
    #...
network_allow_restart: yes
```
I just followed the format above, I also added the complete state examples  below.

richm:
[citest bad]

richm:
[citest pending]

tasks/main.yml
liangwen12year: NM is always installed in the previous task.

library/network_state.py
liangwen12year: I changed into `desired_state`.

tasks/main.yml
liangwen12year: I added the system version check on the nmstate.

.github/run_test.sh
liangwen12year: Yes, the package that contains the libnmstate is not nmstate-devel, but is python3-libnmstate.

tests/playbooks/tests_network_state.yml
```diff
@@ -0,0 +1,189 @@
...
+                address:
+                  - ip: 192.168.122.250
+                    prefix-length: 24
+                dhcp: false
+              ipv6:
+                address:
+                  - ip: 2001:db8::1:1
+                    prefix-length: 64
+                autoconf: false
+                dhcp: false
```
ffmancera, line 53: This is wrong, the `enabled` field is missing for IPv4 and IPv6 configuration. It should indicate `enabled: True` for both of them. Btw, Nmstate should raise a warning or error here and do not ignore it silently.

ffmancera (changes requested):

README.md
```diff
@@ -1173,6 +1197,83 @@ network_connections:
...
+          - ip: 192.168.100.192
+            prefix-length: 24
+        auto-dns: false
+        dhcp: false
+      ipv6:
+        address:
+          - ip: 2001:db8::2:1
+            prefix-length: 64
+        autoconf: false
+        dhcp: false
```
ffmancera, line 1236: This is missing the `enabled` section.. in ipv4 and ipv6.

examples/network_state.yml
```diff
@@ -0,0 +1,47 @@
...
+            address:
+              - ip: 192.168.122.250
+                prefix-length: 24
+            dhcp: false
+          ipv6:
+            address:
+              - ip: 2001:db8::1:1
+                prefix-length: 64
+            autoconf: false
+            dhcp: false
```
ffmancera, line 20: This is missing the `enabled` field in ipv4 and ipv6

ffmancera (changes requested):

README.md
liangwen12year: Thanks. I updated.

liangwen12year: Thanks. I updated.

ffmancera (approved):
LGTM!

Issue #507, opened 2022-06-01
Parser for routing table database is too strict


bsteinb:
The parser for the `iproute2` routing table database (`/etc/iproute2/rt_tables` etc.) in `module_utils/network_lsr/argument_validator.py:IPRouteUtils` is too strict. It asserts that entries/lines in the file are of the form `254 main` where the numeric id and the name must be separated by a single space character only. However, the [original parser](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/rt_names.c#n38) uses something like

```c
sscanf(p, "%d %s\n", id, namebuf)
```

Note that [the semantics of `sscanf` etc.](https://en.cppreference.com/w/c/io/fscanf) are such that any single whitespace character in the format string consumes all available consecutive whitespace characters in the input (spaces, tabs, etc.). The [default `/etc/iproute2/rt_tables`](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/etc/iproute2/rt_tables) indeed separates the two fields by tabs instead of spaces.

bsteinb:
> That is known to be more strict, because iproute2's parsing seems too lax to me, accepting things that it shouldn't accept.

Is the file format documented outside of the implementation of the original parser in iproute2? I could not find anything. If it is not, how do you arrive at the conclusion that it is too lax?

> a more useful report would be to list the exact files that you have

I provided a link to the default configuration file that is included in the iproute2 package, above, but I can reproduce it here verbatim if that helps:

```
#
# reserved values
#
255	local
254	main
253	default
0	unspec
#
# local
#
#1	inr.ruhep
```

This file is used as `/etc/iproute2/rt_tables` as is on RHEL 8 and its derivatives (installed via the `iproute` RPM). The role fails to parse this, since the file uses tabs instead of spaces as the field separator. The original parser in the general case will allow a sequence of one or more white space characters (according to `isspace(3)`) as field separators whereas the role only allows a single space (`' '`).

thom311:
you are completely right. Opened an MR. Thanks.

The issue was caused by not realizing that `scanf()` treats whitespace they way you say (which is of course sensible, on part of `scanf()`).

bsteinb:
No problem. I had to look up `scanf` as well.

I left a review with some things I noticed on the MR, I hope that was not too forward of me.

thom311:
that was very useful. So many plunders :) Thank you!

PR #508, opened 2022-06-02
argument_validator: fix IPRouteUtils.get_route_tables_mapping() to accept any whitespace sequence

thom311:
Accept any whitespace sequence, according to Python's `rb"\s"` class.
This way, tabs are also accepted.

Also add a unit test for the default iproute2 file.

Use just one regex for this.

---

https://github.com/linux-system-roles/network/issues/507

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -2568,65 +2568,61 @@ class ArgValidator_ListConnections(ArgValidatorList):
 
 class IPRouteUtils(object):
 
+    # iproute2 does not care much about the valid characters of a
+    # table alias (it doesn't even require UTF-8 encoding, the only
+    # forbidden parts are whitespace).
+    #
+    # We don't allow such flexibility. Aliases must only contain a
+    # certain set of ASCII characters. These aliases are what we accept
+    # as input (in the playbook), and there is no need to accept
+    # user input with unusual characters or non-ASCII names.
     ROUTE_TABLE_ALIAS_RE = re.compile("^[a-zA-Z0-9_.-]+$")
 
     @classmethod
     def _parse_route_tables_mapping(cls, file_content, mapping):
+
+        # This parses the /etc/iproute2/rt_tables file and constructs
+        # the mapping from table aliases the table numeric IDs.
+        #
+        # It is thus similar to rtnl_rttable_a2n(), from here:
+        # https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/rt_names.c?id=11e41a635cfab54e8e02fbff2a03715467e77ae9#n447
+        regex = re.compile(
+            b"^\\s*(0x[0-9a-fA-F]+|[0-9]+)\\s+([a-zA-Z0-9_.-]+)(\\s*|\\s+#.*)$"
+        )
         for line in file_content.split(b"\n"):
 
-            # iproute2 only skips over leading ' ' and '\t'.
-            line = line.lstrip()
-
-            # skip empty lines or comments
-            if not line:
+            rmatch = regex.match(line)
+            if not rmatch:
                 continue
 
-            # In Python 2.x, there is no new type called `bytes`. And Python 2.x
-            # `bytes` is just an alias to the str type
-            if Util.PY3:
-                if line[0] == ord(b"#"):
-                    continue
-            else:
-                if line[0] == "#":
-                    continue
+            table = rmatch.group(1)
+            name = rmatch.group(2)
 
-            # iproute2 splits at the first space.
-            ll = line.split(b" ", 1)
-            if len(ll) != 2:
-                continue
-            line1 = ll[0]
-            line2 = ll[1]
+            name = name.decode("utf-8")
 
-            comment_space = line2.find(b" ")
-            if comment_space >= 0:
-                # when commenting the route table entry in the same line,
-                # iproute2 only accepts one ' ', followed by '#'
-                if not re.match(b"^[a-zA-Z0-9_.-]+ #", line2):
-                    continue
-                line2 = line2[:comment_space]
-
-            # convert to UTF-8 and only accept benign characters.
-            try:
-                line2 = line2.decode("utf-8")
-            except Exception:
-                continue
-
-            if not cls.ROUTE_TABLE_ALIAS_RE.match(line2):
-                continue
+            if not cls.ROUTE_TABLE_ALIAS_RE.match(name):
+                raise AssertionError(
+                    "bug: table alias contains unexpected characters: %s" % (name,)
+                )
 
             tableid = None
             try:
-                tableid = int(line1)
+                tableid = int(table)
             except Exception:
-                if line.startswith(b"0x"):
+                if table.startswith(b"0x"):
                     try:
-                        tableid = int(line1[2:], 16)
+                        tableid = int(table[2:], 16)
                     except Exception:
                         pass
             if tableid is None or tableid < 0 or tableid > 0xFFFFFFFF:
                 continue
 
-            mapping[line2] = tableid
+            # In case of duplicates, the latter wins. That is unlike iproute2's
+            # rtnl_rttable_a2n(), which does a linear search over the
+            # hash table (thus, the first found name depends on the content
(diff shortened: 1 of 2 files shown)


module_utils/network_lsr/argument_validator.py
```diff
@@ -2572,61 +2572,32 @@ class IPRouteUtils(object):
 
     @classmethod
     def _parse_route_tables_mapping(cls, file_content, mapping):
+        rr = re.compile(b"^\\s*(0x[0-9]+|[0-9]+)\\s*([a-zA-Z0-9_.-]+)(\\s*|\\s+#.*)$")
```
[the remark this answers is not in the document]
thom311: right!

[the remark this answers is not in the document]
thom311: right!

module_utils/network_lsr/argument_validator.py
```diff
@@ -2572,61 +2572,32 @@ class IPRouteUtils(object):
...
-                continue
+            name = name.decode("utf-8")
+            assert cls.ROUTE_TABLE_ALIAS_RE.match(name)
 
             tableid = None
             try:
-                tableid = int(line1)
+                tableid = int(table)
             except Exception:
                 if line.startswith(b"0x"):
```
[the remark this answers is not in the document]
thom311: ACK

tests/unit/test_network_connections.py
```diff
@@ -4550,13 +4550,35 @@ def parse(file_content):
...
+        )
+
         self.assertEqual(parse(b""), {})
         self.assertEqual(parse(b"5 x"), {"x": 5})
-        self.assertEqual(parse(b"   7   y   "), {})
+        self.assertEqual(parse(b"   7   y   "), {"y": 7})
         self.assertEqual(parse(b"5 x\n0x4 y"), {"x": 5, "y": 4})
         self.assertEqual(parse(b"5 x #df\n0x4 y"), {"x": 5, "y": 4})
-        self.assertEqual(parse(b"5 x #df\n0x4 y\n7\ty"), {"x": 5, "y": 4})
+        self.assertEqual(parse(b"5 x #df\n0x4 y\n7\ty"), {"x": 5, "y": 7})
```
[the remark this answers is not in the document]
thom311: I will add a case for all the things you pointed out.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2572,61 +2572,34 @@ class IPRouteUtils(object):
 
     @classmethod
     def _parse_route_tables_mapping(cls, file_content, mapping):
+        rr = re.compile(
+            b"^\\s*(0x[0-9a-fA-F]+|[0-9]+)\\s+([a-zA-Z0-9_.-]+)(\\s*|\\s+#.*)$"
```
richm, line 2576: so a line like
```
0xffff somevalue# this is a comment
```
would not match - is this line allowed by iproute2?

thom311: yes, it would not match. You can test by adding
```
44  ttt1# comment
45  ttt2 # comment
```
to `/etc/iproute2/rt_tables` and type `ip route add 1.2.3.4/32 dev eth0 table ttt1` (while `ttt2` works).

thom311:
hm, from testing it seems that if you specify the same alias more than once, then iproute2 would use the first occurrence. From review, it's not clear to me how that could happen... dunno. The role behaves differently, where later values overwrite the previous ones.

thom311:
> hm, from testing it seems that if you specify the same alias more than once, then iproute2 would use the first occurrence. From review, it's not clear to me how that could happen... dunno. The role behaves differently, where later values overwrite the previous ones.

When iproute2 converts a numeric ID to name (`rtnl_rttable_n2a()`), it will prefer the latter lines in the file.
When it converts from name to ID (`rtnl_rttable_a2n()`) it searches all the names in a non-predictable order ([here](https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/rt_names.c?id=11e41a635cfab54e8e02fbff2a03715467e77ae9#n447)). That's because it uses a simple hash table with 256 entries to map from ID to name. IDs that hash to the same value get chained.

The role only converts from names to numeric IDs, and it does not follow the inconsistent behavior of `rtnl_rttable_a2n()`. Instead, the latter entries win.

I think this makes sense. The lesson is: don't configure duplicate aliases, if you want that what the role configures agrees to what `ip route show` will show.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2568,65 +2568,55 @@ def _ipv6_is_not_configured(connection):
...
-        for line in file_content.split(b"\n"):
 
-            # iproute2 only skips over leading ' ' and '\t'.
-            line = line.lstrip()
+        # This parses the /etc/iproute2/rt_tables file and constructs
+        # the mapping from table aliases the table numeric IDs.
+        #
+        # It is thus similar to rtnl_rttable_a2n(), from here:
+        # https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/rt_names.c?id=11e41a635cfab54e8e02fbff2a03715467e77ae9#n447
+        rr = re.compile(
```
tyll, line 2587: Please use more descriptive variable names than `rr` or `m`.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2568,65 +2568,55 @@ def _ipv6_is_not_configured(connection):
...
                         pass
             if tableid is None or tableid < 0 or tableid > 0xFFFFFFFF:
                 continue
 
-            mapping[line2] = tableid
+            # In case of duplicates, the latter wins. That is unlike iproute2's
+            # rtnl_rttable_a2n(), which does a linear search over the
+            # hash table (thus, the first found name depends on the content
+            # of the hash table and the result in face of duplicates is
+            # not well defined).
```
tyll, line 2618: IMHO it would be better to fail and notify the user so that they can fix this.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2568,65 +2568,55 @@ def _ipv6_is_not_configured(connection):
 
 class IPRouteUtils(object):
 
+    # iproute2 does not care much about the valid characters of a
+    # table alias (it doesn't even require UTF-8 encoding, the only
+    # forbidden parts are whitespace).
+    #
+    # We don't allow such flexibility. Aliases must only contain a
+    # certain set of ASCII characters.
```
tyll, line 2576: Please add a justification for diverting from iproute2 here.

module_utils/network_lsr/argument_validator.py
thom311: we already fail if we cannot resolve a name to a number. We only won't indicate whether the reason is due to syntax error in the rt_tables file.

But sure, follow-up patch welcome. This patch does not change any of this.

module_utils/network_lsr/argument_validator.py
thom311: done.

module_utils/network_lsr/argument_validator.py
thom311: it's a bugfix, so there is a visible change in behavior. It accepts now *more* than before, which tends to be less of a problem.

What does not change, is that invalid lines are ignored (only what makes a line invalid was relaxed)... to do anything about invalid lines is not part of this change.

tyll: The unit test that I linked changes which table entry is parsed from the first one to the last one:

```
-        self.assertEqual(parse(b"5 x #df\n0x4 y\n7\ty"), {"x": 5, "y": 4})
+      self.assertEqual(parse(b"5 x #df\n0x4 y\n7\ty"), {"x": 5, "y": 7})
```

which is the change that is described in this comment.

thom311: I am not against a followup patch to somehow handle errors differently. But I don't think that should be part of this patch. This patch doesn't change how invalid lines are handled (they are still ignored). It only changes, that it now considers strictly more lines as valid than before.

liangwen12year (approved):
LGTM.

PR #509, opened 2022-06-10
Add AlmaLinux to the role

smooge:
AlmaLinux is a Red Hat Enterprise Linux clone which would need testing and additional code like Rocky Linux did.

Signed-off-by: Stephen Smoogen <ssmoogen@redhat.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 [Unreleased]
 --------------------
+
 ### New features
 
 - Add AlmaLinux to various groups
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+[Unreleased]
+--------------------
+### New features
+
+- Add AlmaLinux to various groups
+- Add Rocky and Alma to testing
+
 [1.8.0] - 2022-05-19
 --------------------
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,7 +10,8 @@ network_provider_os_default: "{{
       'RedHat',
       'CentOS',
       'OracleLinux',
-      'Rocky'
+      'Rocky',
+      'AlmaLinux'
     ] and ansible_distribution_major_version is version('7', '<')
     else 'nm' }}"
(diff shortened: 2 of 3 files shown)


richm:
Thanks - do we need to add this to the README?

Note that we do not have any testing for Alma or Rocky, CI or otherwise (nor OracleLinux) - so we can add this, but there is no guarantee it will work for all test cases, now or in the future

[citest commit:18067ad97310f282146a1302747ab2dfa1822b49]

smooge:
Understood. I saw that things were acting differently on my Alma system and was trying to pin it down. This wasn't the fix (the problem being in ansible-core and not this), but it looked like a starting thing to help.

richm:
[citest commit:82dd06b1ac348f1248a6eacd9925f3feff3f34b5]

Conan-Kudo:
Why aren't we just using `RedHat` family instead of identifying all these distributions individually?

richm:
> Why aren't we just using `RedHat` family instead of identifying all these distributions individually?

Does Ansible 2.9 recognize these all as "RedHat" os_family?  If so, ok - we still have many customers using Ansible 2.9 that we will have to support for the foreseeable future.

Conan-Kudo:
I'm not sure Ansible 2.9 recognizes those distros properly at all.

richm:
> I'm not sure Ansible 2.9 recognizes those distros properly at all.

If not, then we can't use `os_family` here.

If the author desires, you could add `vars/main.yml` (or since the network role doesn't use `vars/`, maybe add near the top of `defaults/main.yml` like
```yaml
__network_redhat_family: [RedHat, Fedora, CentOS, OracleLinux, AlmaLinux, Rocky]
```
or even
```
__network_redhat_family: "{{ (ansible_facts['os_family']['RedHat'] +
  ['OracleLinux', 'AlmaLinux', 'Rocky']) | unique }}"
```
then do
```yaml
if ansible_distribution in __network_redhat_family and
```
There are several places in the network role that would be able to make use of such a variable rather than having the hardcoded lists of RedHat family OSes

liangwen12year:
The tox unit test (yamllint) failed, you may take a look.
```
Error: [line-length] line too long (88 > 80 characters)
```

PR #520, opened 2022-07-25
network_state: improve state comparison for achieving idempotency

liangwen12year:
`libnmstate.show()` would include the info like bridge timer etc which
should not be considered when comparing between the previous state and
current state. Instead, using the the `libnmstate.show_running_config()`
which would filter out these kind of data.

### DIFF
--- a/library/network_state.py
+++ b/library/network_state.py
@@ -39,18 +39,27 @@ class NetworkState:
         self.params = module.params
         self.result = dict(changed=False)
         self.module_name = module_name
-        self.previous_state = libnmstate.show()
+        self.previous_state = self.get_state_config()
 
     def run(self):
         desired_state = self.params["desired_state"]
         libnmstate.apply(desired_state)
-        current_state = libnmstate.show()
+        current_state = self.get_state_config()
         if current_state != self.previous_state:
             self.result["changed"] = True
 
         self.result["state"] = current_state
 
         self.module.exit_json(**self.result)
+
+    def get_state_config(self):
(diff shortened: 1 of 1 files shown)


richm:
[citest]

richm:
[citest bad]

richm:
lgtm

richm:
[citest]

richm:
@liangwen12year the tests like `CentOS-7-latest/ansible-2.13/(citool)` with the tft-bot robot icon are the baseos ci tests - they won't be triggered automatically - you must include the citest comment like above

library/network_state.py
```diff
@@ -39,12 +39,22 @@ def __init__(self, module, module_name):
...
     def run(self):
         desired_state = self.params["desired_state"]
         libnmstate.apply(desired_state)
-        current_state = libnmstate.show()
+        if hasattr(libnmstate, "show_running_config") and callable(
+            getattr(libnmstate, "show_running_config")
+        ):
+            current_state = libnmstate.show_running_config()
+        else:
+            current_state = libnmstate.show()
```
ffmancera, line 57: just a nit, as this is being used in two different places, what do you think about putting this block in its own function? That function should return what `show()` or `show_running_config()` is returning.

What do you think?

ffmancera (changes requested):

library/network_state.py
liangwen12year: I updated.

liangwen12year:
[citest]

ffmancera (approved):

PR #521, opened 2022-07-27
Use GITHUB_REF_NAME as name of push branch; fix error in branch detection [citest skip]

richm:
We need to get the name of the branch to which CHANGELOG.md was pushed.
For now, it looks as though `GITHUB_REF_NAME` is that name.  But don't
trust it - first, check that it is `main` or `master`.  If not, then use
a couple of other methods to determine what is the push branch.


nhosoi (approved):
lgtm

liangwen12year (approved):
lgtm.

PR #525, opened 2022-08-23
adding cloned_mac

35niavlys:
First draft to add cloned-mac/MACADDR config.
I do not have the skills to know if it is consistent on all types of interfaces

Tests will fail at the moment but let me know if it's far to be mergeable first


module_utils/network_lsr/argument_validator.py
```diff
@@ -596,6 +596,28 @@ def _validate_impl(self, value, name):
             )
         return Util.mac_ntoa(addr)
 
+class ArgValidatorClonedMac(ArgValidatorStr):
```
thom311, line 599: instead of adding a new type, just extend (and use) `ArgValidatorMac`. Add an `enum_values` parameter to `ArgValidatorMac.__init__()` and set it to `enum_values=["preserve", "permanent", "random", "stable"]`.

But of course, this `enum_values` argument needs to be implemented by `ArgValidatorMac` itself, and *not* via `v = ArgValidatorStr._validate_impl(self, value, name, enum_values=enum_values)`.

liangwen12year:
Please consider signing off the commit to remove the DCO failure, https://github.com/linux-system-roles/network/blob/main/contributing.md#sign-off-your-commit

library/network_connections.py
```diff
@@ -336,6 +336,7 @@ def ifcfg_create(
             ifcfg["ONBOOT"] = "no"
 
         ifcfg["DEVICE"] = connection["interface_name"]
+        ifcfg["MACADDR"] = connection["cloned_mac"]
```
liangwen12year, line 340: @thom311  are we still planning to support the new feature for the initscripts provider  ? IIRC,  we only do bugfix in the initscripts provider.

module_utils/network_lsr/argument_validator.py
```diff
@@ -596,6 +596,28 @@ def _validate_impl(self, value, name):
...
+            return v
+
+        try:
+            addr = Util.mac_aton(v, self.force_len)
+        except MyError:
+            raise ValidationError(
+                name, "value '%s' is not a valid MAC address nor a valid mode ['preserve', 'permanent', 'random', 'stable']" % (value)
+            )
+        if not addr:
+            raise ValidationError(
```
liangwen12year, line 607: Maybe we should validate this before calling `Util.mac_aton()`, so that for the case of empty mac,  we can raise an exception earlier.

And in the NM, there is a default value `ethernet.cloned-mac-address=preserve`, maybe we should also set the default value preserve here. @thom311 , what do you think ?

thom311: > And in the NM, there is a default value ethernet.cloned-mac-address=preserve, maybe we should also set the default value preserve here. @thom311 , what do you think ?

The default value of the ArgValidator should be chosen so, that it makes sense and so that it's most convenient to the callers of the code. The current `None` value is chosen, so that the callers can distinguish whether the value was explicitly set in YAML or whether it's unset (`None`). I don't know whether any of the callers actually uses that information.

library/network_connections.py
```diff
@@ -336,6 +336,7 @@ def ifcfg_create(
...
```
thom311, line 339: setting this to special values like `preserve` wouldn't work... the role should do something about this.

for example, if the initscripts provider is used, then `validate_connection_one()` would handle that (e.g. by logging a warning and ignoring the unsupported value, or failing the playbook as invalid altogether).

module_utils/network_lsr/argument_validator.py
35niavlys: According a [gnome blog post](https://blogs.gnome.org/thaller/tag/clone-mac-address/) the default value was different on older version so I think we should not define a default

thom311: The default value for `cloned-mac-address` is "unset" in the profile, which then allows to fallback to a connection default in NetworkManager.conf (if specified), and otherwise fallback to "preserve" (or "permanent" on earlier versions).

In any case, the default value in the profile is neither "preserve" nor "permanent". It only ends up effectively being one of those settings, if it's still unspecified in NetworkManager.conf.

As far as the role is concerned, it can implement whatever it wants and -- if the user does not specify cloned-mac-address in the playbook -- it may set the per profile value to "preserve". In practice, it makes however sense to follow what NetworkManager would do, and leave the value unset, if it's unset in the playbook...

In short, having `None` as default makes a lot of sense.

library/network_connections.py
35niavlys: added an exception in `validate_connection_one()`

README.md
```diff
@@ -364,6 +364,18 @@ with double quotes and sometimes it is necessary.
 
 - For `initscripts`,  `mac` is the currently configured MAC address of the device (`HWADDR`).
 
+### `cloned_mac`
+
+The `cloned_mac` address is optional and allow to specify the strategy to get the default
+mac or to set your own mac. The value of the `cloned_mac` address needs to be specified in
```
liangwen12year, line 370: How about "Besides explicitly specifying the value as a MAC address with hexadecimal notation, the following special values are also supported:" ?

liangwen12year:
The integration test and example need to be added.

README.md
35niavlys: done

README.md
```diff
@@ -364,6 +364,17 @@ with double quotes and sometimes it is necessary.
...
+hexadecimal notation like `mac` property. Besides explicitly specifying the value as a MAC
+address with hexadecimal notation, the following special values are also supported:
+- `permanent`: use the permanent MAC address of the device
```
liangwen12year, line 373: Lists should be surrounded by blank lines. (Maybe insert a blank line above it.)

35niavlys:
I will try to prepare my environment to run tests easily and add some in the following weeks but it will take a little work time

tests/playbooks/tests_cloned_mac.yml
```diff
@@ -0,0 +1,105 @@
...
+          register: cloned_mac_address
+          ignore_errors: yes
+          changed_when: false
+          loop:
+            - { name: "{{ controller_profile }}", mac: "12:23:34:45:56:60"}
+            - { name: "{{ port1_profile }}", mac: "12:23:34:45:56:61"}
+            - { name: "{{ port2_profile }}", mac: "--"}
+            - { name: "{{ bridge_interface }}", mac: "12:23:34:45:56:70"}
+
+        - name: "Assert that DNS addresses are configured correctly"
```
liangwen12year, line 70: Assert that the cloned-mac-address is configured correctly ?

tests/playbooks/tests_cloned_mac.yml
```diff
@@ -0,0 +1,105 @@
...
+                type: bridge
+                ip:
+                  dhcp4: no
+                  auto6: yes
+                cloned_mac: "12:23:34:45:56:70"
+
+        - name: Verify nmcli cloned-mac-address entry
+          shell: |
+            set -euxo pipefail
+            nmcli connection show {{ item.name }} | grep cloned-mac-address
```
liangwen12year, line 60: how about `nmcli -f 802-3-ethernet.cloned-mac-address c show {{ item.name }}`  ?

liangwen12year:
```
[wenliang@localhost network]$ tox -e yamllint 
./tests/tests_cloned_mac_initscripts.yml
  5:81      error    line too long (82 > 80 characters)  (line-length)
```
You may consider to ignore the line-length check of `tests_cloned_mac_initscripts.yml` in .yamllint.yml file.

liangwen12year:
[citest]

richm:
[citest pending]

35niavlys:
I did small corrections on the test. Not all devices are tested, especially wifi, but I can't get the existing tests to work :(

liangwen12year:
[citest]

liangwen12year:
A nit, maybe you can also provide an example under the `./examples/` directory.

liangwen12year:
yamllint check failed:
```
./tests/playbooks/tests_cloned_mac.yml
  Error: 62:81 [line-length] line too long (84 > 80 characters)
  Error: 72:81 [line-length] line too long (83 > 80 characters)

ERROR:   yamllint: commands failed
```

tests/playbooks/tests_cloned_mac.yml
```diff
@@ -0,0 +1,107 @@
...
+              - name: "{{ bridge_interface }}"
+                state: up
+                type: bridge
+                ip:
+                  dhcp4: no
+                  auto6: yes
+                cloned_mac: "12:23:34:45:56:70"
+
+        - name: Verify nmcli cloned-mac-address entry
+          shell: nmcli -f 802-3-ethernet.cloned-mac-address con show {{ item.name }}
```
richm, line 62: use `command` instead of `shell`
wrap like this:
```yaml
command: >-
  nmcli -f 802-3-ethernet.cloned-mac-address con show {{ item.name }}
```

tests/playbooks/tests_cloned_mac.yml
```diff
@@ -0,0 +1,107 @@
...
+          register: cloned_mac_address
+          ignore_errors: yes
+          changed_when: false
+          loop:
+            - { name: "{{ controller_profile }}", mac: "12:23:34:45:56:60"}
+            - { name: "{{ port1_profile }}", mac: "12:23:34:45:56:61"}
+            - { name: "{{ port2_profile }}", mac: "--"}
+            - { name: "{{ bridge_interface }}", mac: "12:23:34:45:56:70"}
+
+        - name: "Assert that cloned-mac-address addresses are configured correctly"
```
richm, line 72:
```suggestion
        - name: >
            Assert that cloned-mac-address addresses are configured correctly
```

liangwen12year:
[citest]

README.md
```diff
@@ -364,6 +364,18 @@ with double quotes and sometimes it is necessary.
...
 
+### `cloned_mac`
+
+The `cloned_mac` address is optional and allow to specify the strategy to get the default
+mac or to set your own mac. The value of the `cloned_mac` address needs to be specified in
+hexadecimal notation like `mac` property. Besides explicitly specifying the value as a MAC
+address with hexadecimal notation, the following special values are also supported:
+
+- `permanent`: use the permanent MAC address of the device
+- `preserve`: don’t change the MAC address of the device upon activation
```
liangwen12year, line 375: The quote here needs to be changed according to the CI checks.


```
Running sanity test "no-smart-quotes"
ERROR: Found 2 no-smart-quotes issue(s) which need to be resolved:
ERROR: docs/README_network.md:375:18: use ASCII quotes `'` and `"` instead of Unicode quotes
ERROR: roles/network/README.md:375:18: use ASCII quotes `'` and `"` instead of Unicode quotes
See documentation for help: https://docs.ansible.com/ansible-core/2.12/dev_guide/testing/sanity/no-smart-quotes.html
```

liangwen12year:
[citest]

liangwen12year:
> [citest]

BaseCI have to be triggered this way.

35niavlys:
> > [citest]
> 
> BaseCI has to be triggered this way.

I well understood but I don't see how my changes cause this error, I will check later

liangwen12year:
[citest]

35niavlys:
> [citest]

Can you restart CI on main branch ? because it's no more green :(

richm:
[citest]

richm:
[citest bad]

tests/playbooks/tests_bridge_cloned_mac.yml
```diff
@@ -0,0 +1,62 @@
...
+          dhcp4: no
+          auto6: yes
+        cloned_mac: "{{ cloned_mac }}"
+  roles:
+    - linux-system-roles.network
+  tasks:
+    - name: Verify ifcfg MACADDR entry
+      command: >-
+        grep ^MACADDR=
+        /etc/sysconfig/network-scripts/ifcfg-{{ bridge_interface }}
```
liangwen12year, line 29: The CI failed. The error was: 'bridge_interface' is undefined.

Should this be ? 
```
        grep ^MACADDR=
        /etc/sysconfig/network-scripts/ifcfg-{{ interface }}
```

tests/playbooks/tests_bridge_cloned_mac.yml
```diff
@@ -0,0 +1,62 @@
...
+      register: cloned_mac_address_init
+      changed_when: false
+      when:
+        - network_provider == "initscripts"
+
+    - name: Verify nmcli cloned-mac-address entry
+      command: >-
+        nmcli -f 802-3-ethernet.cloned-mac-address con show
+        {{ bridge_interface }}
```
liangwen12year, line 38: same here.

liangwen12year:
@richm , the `./tests/tests_bond_cloned_mac_initscripts.yml` failed with the following error, just wondered why this happened because I can not reproduce locally.
```
TASK [linux-system-roles.network : Re-test connectivity] ***********************
task path: /WORKDIR/dist-git-network-add_cloned_mac-BwEDGn/tests/roles/linux-system-roles.network/tasks/main.yml:149
Monday 26 September 2022  14:39:08 +0000 (0:00:01.811)       0:04:29.473 ****** 
fatal: [sut]: UNREACHABLE! => {
    "changed": false,
    "unreachable": true
}

MSG:

Data could not be sent to remote host "10.31.8.243". Make sure this host can be reached over ssh: mux_client_request_session: read from master failed: Broken pipe
ssh: connect to host 10.31.8.243 port 22: Connection timed out

	to retry, use: --limit @/tmp/tests_bond_cloned_mac_initscripts.retry

PLAY RECAP *********************************************************************
sut                        : ok=23   changed=5    unreachable=1    failed=0    skipped=14   rescued=0    ignored=0   
```

tests/playbooks/tests_bridge_cloned_mac.yml
35niavlys: don't understand how i commit that but yes definitly an error

35niavlys:
I can't run all tests on the HEAD of current main branch. You #526 pull request @liangwen12year seems to break for the same reason

liangwen12year:
[citest]

liangwen12year:
> I can't run all tests on the HEAD of current main branch. You #526 pull request @liangwen12year seems to break for the same reason

What do you mean specifically ?

richm:
> @richm , the `./tests/tests_bond_cloned_mac_initscripts.yml` failed with the following error, just out of curiosity, this is the CI problem, right? because I can not reproduce locally. Thanks.
> 
> ```
> TASK [linux-system-roles.network : Re-test connectivity] ***********************
> task path: /WORKDIR/dist-git-network-add_cloned_mac-BwEDGn/tests/roles/linux-system-roles.network/tasks/main.yml:149
> Monday 26 September 2022  14:39:08 +0000 (0:00:01.811)       0:04:29.473 ****** 
> fatal: [sut]: UNREACHABLE! => {
>     "changed": false,
>     "unreachable": true
> }
> 
> MSG:
> 
> Data could not be sent to remote host "10.31.8.243". Make sure this host can be reached over ssh: mux_client_request_session: read from master failed: Broken pipe
> ssh: connect to host 10.31.8.243 port 22: Connection timed out
> 
> 	to retry, use: --limit @/tmp/tests_bond_cloned_mac_initscripts.retry
> 
> PLAY RECAP *********************************************************************
> sut                        : ok=23   changed=5    unreachable=1    failed=0    skipped=14   rescued=0    ignored=0   
> ```

not sure - can you send me a link to the log?
Is it possible that the test `tests_bond_cloned_mac_initscripts` can temporarily interrupt ssh networking?  If so, then the task `Re-test connectivity` is going to have to be more robust to handle that case.

and yes - there is a difference between the way tests work when run locally with qemu/kvm machine managed node and  local virtual networking, vs. the CI system which launches an openstack or aws machine for the managed node, and uses a "real" network connection which has a lot more latency and opportunities for drops.

liangwen12year:
> > @richm , the `./tests/tests_bond_cloned_mac_initscripts.yml` failed with the following error, just out of curiosity, this is the CI problem, right? because I can not reproduce locally. Thanks.
> > ```
> > TASK [linux-system-roles.network : Re-test connectivity] ***********************
> > task path: /WORKDIR/dist-git-network-add_cloned_mac-BwEDGn/tests/roles/linux-system-roles.network/tasks/main.yml:149
> > Monday 26 September 2022  14:39:08 +0000 (0:00:01.811)       0:04:29.473 ****** 
> > fatal: [sut]: UNREACHABLE! => {
> >     "changed": false,
> >     "unreachable": true
> > }
> > 
> > MSG:
> > 
> > Data could not be sent to remote host "10.31.8.243". Make sure this host can be reached over ssh: mux_client_request_session: read from master failed: Broken pipe
> > ssh: connect to host 10.31.8.243 port 22: Connection timed out
> > 
> > 	to retry, use: --limit @/tmp/tests_bond_cloned_mac_initscripts.retry
> > 
> > PLAY RECAP *********************************************************************
> > sut                        : ok=23   changed=5    unreachable=1    failed=0    skipped=14   rescued=0    ignored=0   
> > ```
> 
> not sure - can you send me a link to the log? Is it possible that the test `tests_bond_cloned_mac_initscripts` can temporarily interrupt ssh networking? If so, then the task `Re-test connectivity` is going to have to be more robust to handle that case.
> 
> and yes - there is a difference between the way tests work when run locally with qemu/kvm machine managed node and local virtual networking, vs. the CI system which launches an openstack or aws machine for the managed node, and uses a "real" network connection which has a lot more latency and opportunities for drops.

https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-525-21d25a3_RHEL-8.8.0-20220923.0_20220927-004002/artifacts/summary.html

richm:
> > > @richm , the `./tests/tests_bond_cloned_mac_initscripts.yml` failed with the following error, just out of curiosity, this is the CI problem, right? because I can not reproduce locally. Thanks.
> > > ```
> > > TASK [linux-system-roles.network : Re-test connectivity] ***********************
> > > task path: /WORKDIR/dist-git-network-add_cloned_mac-BwEDGn/tests/roles/linux-system-roles.network/tasks/main.yml:149
> > > Monday 26 September 2022  14:39:08 +0000 (0:00:01.811)       0:04:29.473 ****** 
> > > fatal: [sut]: UNREACHABLE! => {
> > >     "changed": false,
> > >     "unreachable": true
> > > }
> > > 
> > > MSG:
> > > 
> > > Data could not be sent to remote host "10.31.8.243". Make sure this host can be reached over ssh: mux_client_request_session: read from master failed: Broken pipe
> > > ssh: connect to host 10.31.8.243 port 22: Connection timed out
> > > 
> > > 	to retry, use: --limit @/tmp/tests_bond_cloned_mac_initscripts.retry
> > > 
> > > PLAY RECAP *********************************************************************
> > > sut                        : ok=23   changed=5    unreachable=1    failed=0    skipped=14   rescued=0    ignored=0   
> > > ```
> > 
> > 
> > not sure - can you send me a link to the log? Is it possible that the test `tests_bond_cloned_mac_initscripts` can temporarily interrupt ssh networking? If so, then the task `Re-test connectivity` is going to have to be more robust to handle that case.
> > and yes - there is a difference between the way tests work when run locally with qemu/kvm machine managed node and local virtual networking, vs. the CI system which launches an openstack or aws machine for the managed node, and uses a "real" network connection which has a lot more latency and opportunities for drops.
> 
> https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-525-21d25a3_RHEL-8.8.0-20220923.0_20220927-004002/artifacts/summary.html

It might also be a test flake - I guess we'll find out in 5 hours or so when the tests complete.

35niavlys:
> > I can't run all tests on the HEAD of current main branch. You #526 pull request @liangwen12year seems to break for the same reason
> 
> What do you mean specifically ?

I mean that [this error](https://github.com/linux-system-roles/network/actions/runs/3137547556/jobs/5096107779#step:3:4815) is not due to one my changes. Same error on [your PR](https://github.com/linux-system-roles/network/actions/runs/3137547556/jobs/5096107779#step:3:4440)

liangwen12year:
> > > I can't run all tests on the HEAD of current main branch. You #526 pull request @liangwen12year seems to break for the same reason
> > 
> > 
> > What do you mean specifically ?
> 
> I mean that [this error](https://github.com/linux-system-roles/network/actions/runs/3137547556/jobs/5096107779#step:3:4815) is not due to one my changes. Same error on [your PR](https://github.com/linux-system-roles/network/actions/runs/3137547556/jobs/5096107779#step:3:4440)

Yes, we have been noticed that, and it is an bug in NM, it has been fixed in the NM upstream, https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/3871c670ab9417fc54d3c0450e91e08ced4a98b4. We may also backport the fix to the earlier version.

tests/playbooks/tests_bond_cloned_mac.yml
```diff
@@ -0,0 +1,96 @@
...
+                    persistent_state: absent
+                    state: down
+                  - name: "{{ controller_profile }}"
+                    persistent_state: absent
+                    state: down
+              ignore_errors: true
+            - command: ip link del {{ controller_device }}
+              ignore_errors: true
+            - import_tasks: tasks/remove_test_interfaces_with_dhcp.yml
+            - name: "Restore the /etc/resolv.conf for initscript"
```
richm, line 91: You do not need the double quotes for this string

tests/playbooks/tests_bridge_cloned_mac.yml
```diff
@@ -0,0 +1,60 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Test configuring bridges
+  hosts: all
+  tasks:
+    - name: "define vars"
```
richm, line 6: you do not need the double quotes for this string

tests/playbooks/tests_bridge_cloned_mac.yml
```diff
@@ -0,0 +1,60 @@
...
+      changed_when: false
+      when:
+        - network_provider == "nm"
+
+    - name: Assert that cloned-mac-address addresses are configured correctly
+      assert:
+        that:
+          - >-
+            cloned_mac_address_init.stdout is not defined or
+            cloned_mac_address_init.stdout.find('{{ cloned_mac }}') != -1
```
richm, line 48: Since this is being evaluated in a jinja context, you do not need the `{{ ... }}` here - so just
```
            cloned_mac_address_init.stdout is not defined or
            cloned_mac_address_init.stdout.find(cloned_mac) != -1
```
and also below

35niavlys:
Everything should be ok => [citest]

richm:
[citest]

richm:
please rebase on top of the latest `main` branch - this should help fix some of the ci test failures

richm:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

tests/playbooks/tests_bond_cloned_mac.yml
```diff
@@ -69,6 +70,31 @@
               - item.stdout.endswith(item.item.mac)
             msg: cloned-mac-address is configured incorrectly
           loop: "{{ cloned_mac_address.results }}"
+          when: network_provider == 'nm'
```
liangwen12year, line 73: The verification scheme `nmcli -f 802-3-ethernet.cloned-mac-address con show` did not work for initscripts provider, thus I verify the MAC address in the `ip addr show` for initscripts provider.

liangwen12year:
@35niavlys , the CI passed, please rebase and squash the commits, thanks.

liangwen12year:
[citest]

tests/playbooks/tests_bond_cloned_mac.yml
35niavlys: ok thanks a lot for the work

35niavlys:
Rebased and squashed everything, again thank you for all the given help for my first pull request

liangwen12year:
[citest]

liangwen12year (commented):
For the commit message, maybe can be changed as :
```
Support cloned MAC address

The users want to create the bonding interface with the MAC address
specified by themselves or specify the strategy (e.g. random, preserve) 
to get the default MAC for the bonding interface. Thus, add support for
the cloned MAC address.
```
Notice that the commit message should focus more on what is the reason precipitating this patch, here is a nice blog about how to write commit message, https://cbea.ms/git-commit/.

liangwen12year (dismissed):
LGTM.

library/network_connections.py
```diff
@@ -960,6 +961,17 @@ def connection_create(self, connections, idx, connection_current=None):
...
+        if connection["type"] == "wireless":
+            s_wireless = self.connection_ensure_setting(con, NM.SettingWireless)
+            s_wireless.set_property(
+                NM.SETTING_WIRELESS_CLONED_MAC_ADDRESS, connection["cloned_mac"]
+            )
+        else:
+            s_wired = self.connection_ensure_setting(con, NM.SettingWired)
+            s_wired.set_property(
+                NM.SETTING_WIRED_CLONED_MAC_ADDRESS, connection["cloned_mac"]
+            )
```
thom311, line 973: if `connection["cloned_mac"] is None`, then `s_wired.set_property()` indeed has no effect. However, we would still call `self.connection_ensure_setting()` and add a setting. That seems not desirable. For example, if you add a bridge profile, then adding an otherwise empty `[ethernet]` setting seems wrong. Let's add this block inside a
```
    if connection["cloned_mac"]:
        if connection["type"] == "wireless":
            ...
```

README.md
```diff
@@ -364,6 +364,18 @@ with double quotes and sometimes it is necessary.
...
+
+The `cloned_mac` address is optional and allow to specify the strategy to get the default
+mac or to set your own mac. The value of the `cloned_mac` address needs to be specified in
+hexadecimal notation like `mac` property. Besides explicitly specifying the value as a MAC
+address with hexadecimal notation, the following special values are also supported:
+
+- `permanent`: use the permanent MAC address of the device
+- `preserve`: don't change the MAC address of the device upon activation
+- `random`: generate a randomized value upon each connect
+- `stable`: generate a stable, hashed MAC address
```
thom311, line 377: if you omit the new `cloned_mac` property in the playbook, then `ArgValidatorMac()` gives `None` and `connection["cloned_mac"]` will be `None`.

That translates to `ethernet.cloned-mac-address=` in nmcli (or NULL in C code of the daemon).

So beside what is documented here, there is yet another value in the playbook. Which is to omit `cloned_mac` and has a special meaning (that none of the other values covers): to honor the global connection default.


I think, there should be one more enum value:
```
- `permanent`: use the permanent MAC address of the device
- `preserve`: don't change the MAC address of the device upon activation
- `random`: generate a randomized value upon each connect
- `stable`: generate a stable, hashed MAC address
- `default`: honor the default behavior in NetworkManager, which can be configured
   in NetworkManager.conf or falls back to "preserve".

Omitting the `cloned_mac` key is the same as setting `default`.
```

The advantage is, that in the playbook you can now do:
```
    cloned_mac: "{{ cloned_mac }}"
```
where the variable can also mean to use the default. Well, I am not familiar with Jinja templates, so I suspect it's even possible with the current approach to clear the `cloned_mac` key from the playbook if `{{ cloned_mac }}` is unset. But that seems non-obvious (I wouldn't know the syntax for that). And in general it seems nice that all the possible values have an explicit string representation (`{{ cloned_mac }}`, including "default") and omitting the `cloned_mac:` key is the same as setting one of those values.


And below:
```
                ArgValidatorMac(
                    "cloned_mac",
                    default_value="default",
                    enum_values=["default", "preserve", "permanent", "random", "stable"],
                ),
```
(and handle `"default"` in the code).

library/network_connections.py
35niavlys: True, I did the change

README.md
35niavlys: I do not necessarily agree, in my opinion, when we want the default conf we do not define it in jinja or according to [the documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#making-variables-optional) with something like: 
```
cloned_mac: "{{ cloned_mac | default(omit) }}"
```
@liangwen12year what do you think about it?

thom311: > I do not necessarily agree, in my opinion, when we want the default conf 

Note that "the default conf" in this case is another mode that is not configurable otherwise.

For example, for "mtu" all valid values range from `0` to `0xFFFFFFFF` (whereas 0 means that NetworkManager does not configure the MTU). Indeed, omitting the "mtu" key in the playbook is the same as setting it to zero "0", so you could say "the mtu key in the playbook defaults to zero". It is the same as setting `mtu: 0`, and you could explicitly set the value that happens to be the default.

With the `cloned_mac` property as it is, omitting the key defaults to a certain behavior, which is not accessibly otherwise. Omitting the key is a special setting itself (and not a default for an explicitly configurable settings).

> cloned_mac: "{{ cloned_mac | default(omit) }}"

Ah right. I thought this would be possible to achieve. Seems non-obvious though.

thom311: to be clear, we probably have properties in the playbook where omitting them is some genuine behavior. So if others also think this is fine, then I don't mind :)

liangwen12year: Honestly, I would prefer to add the `default` value into the enum values `enum_values=["default", "preserve", "permanent", "random", "stable"],` and the have the `default` value documented as "honor the default behavior in NetworkManager ... ". The benefit is that when omiting the `cloned_mac` setting, the `default` value can still be set in the controlled level but not at the user level with the hack  `cloned_mac: "{{ cloned_mac | default(omit) }}"`. Plus, all the possible value for the `cloned_mac` can be listed with explicit string represenation.

liangwen12year:
[citest]

README.md
35niavlys: Ok, I did a simple implementation that do nothing when `default` is set, is it what you expect ?

liangwen12year:
[citest]

README.md
liangwen12year: I think so, when default is set, then the setting `cloned_mac` in the connection profile is unset,  then NM lookup the global connection defaults (in NetworkManager.conf). If it's not specified there either, then the current version will default to "preserve".

module_utils/network_lsr/argument_validator.py
```diff
@@ -1822,6 +1835,10 @@ def __init__(self):
                 ArgValidatorDeprecated("master", deprecated_by="controller"),
                 ArgValidatorStr("interface_name", allow_empty=True),
                 ArgValidatorMac("mac"),
+                ArgValidatorMac(
+                    "cloned_mac",
```
liangwen12year, line 1839: default_value="default" ?

liangwen12year:
`tox -e black` failed, you may use  `black module_utils/network_lsr/argument_validator.py --line-length 88` to format the code.

module_utils/network_lsr/argument_validator.py
35niavlys: I removed it to work with default value, without it, yes I can define it here

liangwen12year:
[citest]

library/network_connections.py
```diff
@@ -334,6 +334,8 @@ def ifcfg_create(
             ifcfg["ONBOOT"] = "no"
 
         ifcfg["DEVICE"] = connection["interface_name"]
+        if connection["cloned_mac"] and connection["cloned_mac"] != "default":
```
liangwen12year, line 337: With initscripts provider, `connection["cloned_mac"]` will not be 'default' at this step, I think that we can remove it and only keep it as `if connection["cloned_mac"]:`.

liangwen12year:
[citest]

liangwen12year (approved):
lgtm.

PR #526, opened 2022-09-22
Support looking up named route table in routing rule

liangwen12year:

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,10 @@ The IP configuration supports the following options:
    - `suppress_prefixlength` -
       Reject routing decisions that have a prefix length of the specified or less.
    - `table` -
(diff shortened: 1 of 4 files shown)


liangwen12year:
[citest]

ffmancera (approved):

ffmancera:
I noticed the CI failures are related to the bond arp_all_targets issue, do you think we could mark them as xfail or skip until the new builds are in c8s and c9s? Thanks!

liangwen12year:
> I noticed the CI failures are related to the bond arp_all_targets issue, do you think we could mark them as xfail or skip until the new builds are in c8s and c9s? Thanks!

I think that we may not test the corner cases anymore, instead, enable the arp_interval in the testing.

PR #527, opened 2022-09-30
bond: Test arp_all_targets only when arp_interval is enabled

liangwen12year:
Kernel allows to set `arp_all_targets` when `arp_interval` is disabled (disable ARP monitoring). But `arp_all_targets` specifies the quantity of `arp_ip_targets` that must be reachable in order for the ARP monitor to consider a slave as being up. It makes more sense to set the `arp_all_targets` while enabling the `arp_interval`.

### DIFF
--- a/tests/playbooks/tests_bond_options.yml
+++ b/tests/playbooks/tests_bond_options.yml
@@ -40,7 +40,6 @@
                   ad_select: stable
                   ad_user_port_key: 1023
                   all_ports_active: True
-                  arp_all_targets: all
                   downdelay: 0
                   lacp_rate: slow
                   lp_interval: 128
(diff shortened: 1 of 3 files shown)


liangwen12year:
[citest]

ffmancera (changes requested):
Hi! I think it should be better to enable arp_interval.. it will make the test cover both options instead of stop covering them

liangwen12year:
> Hi! I think it should be better to enable arp_interval.. it will make the test cover both options instead of stop covering them

Sure, but note that  arp_interval only works for mode "balance-rr", "active-backup", "balance-xor", "broadcast".

liangwen12year:
[citest]

tests/playbooks/tests_bond_options.yml
```diff
@@ -124,17 +122,31 @@
                 interface_name: "{{ controller_device }}"
                 bond:
                   mode: active-backup
+                  arp_all_targets: any
```
ffmancera, line 125: I think this should be "all" as "any" is the default value so we wouldn't notice even if NetworkManager is not setting it. Thank you!

ffmancera (changes requested):

liangwen12year:
[citest]

tests/playbooks/tests_bond_options.yml
liangwen12year: Nice catch.

liangwen12year:
[citest]

liangwen12year:
[citest]

ffmancera (approved):

PR #528, opened 2022-10-04
Add support for __network_test_flags to workaround CI problems

richm:
On some CI systems the tests that create routes can create them in such
a way that they override the "default" default routes, interrupting
network connectivity and causing Ansible to hang.  The parameter
`__network_test_flags` gives network role developers a way to ensure
that the newly created route is given a lower priority and does not
override the "default" default route.  A `dict` was chosen as the type
of this variable so that it could be easily extended in the future for
future use cases.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -323,6 +323,17 @@ ansible-playbook --skip-tags tests::cleanup \
     -i testhost, tests/playbooks/tests_802_1x.yml
 ```
 
+### Test Hacks
+
+There is a parameter `__network_test_flags` which is a `dict` and can be used to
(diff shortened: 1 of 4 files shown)


richm:
> lgtm, but it seems that the `tox -e black` failed.

yes - I don't want to merge this yet - I'm anticipating other reviewers may have suggested changes, or an entirely different approach - this is here just as a PoC - I'm using this PR to test in baseos CI

liangwen12year:
This pr can be closed as we prefer the solution in https://github.com/linux-system-roles/network/pull/532.

PR #534, opened 2022-10-27
add ip.route_metric4: 65535 to failing bond tests

richm:
When creating a bond, the bond also creates a default route with a
default metric of 0.  This causes test failures on CI systems as
it overrides the system default route.  Use the new `ip.route_metric4`
parameter to set a high metric value so as not to override the
default system route.
Some systems cannot use a metric value of 32 bit unsigned int max value.  To ensure
the broadest possible support, use a metric value of 16 bit signed int max value,
which should be high enough to ensure the routes always have the lowest priority.


richm:
[citest]

richm:
[citest]

PR #535, opened 2022-10-28
use rpm -i instead of yum install for epel7

richm:
On BaseOS CI systems, `yum install` for the epel7 rpm does not work.
Instead, use `rpm -i` which should work on any system.  We should not
need to use `yum install` here because the epel7 setup rpm does not have
additional dependencies.

### DIFF
--- a/tests/tasks/enable_epel.yml
+++ b/tests/tasks/enable_epel.yml
@@ -3,10 +3,18 @@
 - block:
     - name: Create EPEL {{ ansible_distribution_major_version }}
       # yamllint disable-line rule:line-length
-      command: yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
+      command: rpm -iv https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
       args:
         warn: false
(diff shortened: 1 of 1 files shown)


richm:
This should fix the CI failures on el7
[citest]

liangwen12year:
The code looks good to me. I was testing in the VM using both `rpm -iv`  and `yum install`, and the warning was prompted:
```
Many EPEL packages require the CodeReady Builder (CRB) repository.
It is recommended that you run /usr/bin/crb enable to enable the CRB repository.
```
Should we enable the CRB repository ? wdyt?

richm:
> The code looks good to me. I was testing in the VM using both `rpm -iv` and `yum install`, and the warning was prompted:
> 
> ```
> Many EPEL packages require the CodeReady Builder (CRB) repository.
> It is recommended that you run /usr/bin/crb enable to enable the CRB repository.
> ```
> 
> Should we enable the CRB repository ? wdyt? (it seems not needed to be enabled. )

No, not unless we need it.  If we enable it, that may cause unwanted packages to be installed.

liangwen12year (approved):
lgtm.

richm:
[citest]

richm:
The failures are known:
* cs8 - need CI image rebuilt
* rhel6 - needs some work from the networking team to fix the tests
* f35, rhel9 - more work needed on bond tests

PR #538, opened 2022-11-10
support ansible-core-2.14, ansible-lint 6.x

richm:
ansible-core 2.14 is now the current version of Ansible.  This version
does not support `args: warn: false` so we have to remove it from the
network role.  Users will need to use `COMMAND_WARNINGS` in their
Ansible configuration in order to suppress the warning in older versions
of Ansible such as 2.9, 2.11.
In addition, the gating tests are getting stricter about using new best
practices, such as using `true`, `false` for booleans instead of `yes`,
`no`; use of spaces in Jinja expressions; etc.  These issues were
addressed.
The `tests/` directory contains far too many cases of non-recommended
practices, so this directory has been exempted.


richm:
Once this merges, I will use https://github.com/linux-system-roles/tox-lsr/pull/95 for the network role.

richm:
[citest]

.ansible-lint
```diff
@@ -1,8 +1,10 @@
 ---
+exclude_paths:
+  - tests/
```
tyll, line 3: Is excludiding "tests" already needed at this point or only after the tox-lsr update? Do you have a list of failures for the tests directory? @liangwen12year can maybe address them.

.sanity-ansible-ignore-2.13.txt
```diff
@@ -0,0 +1,77 @@
+plugins/module_utils/network_lsr/nm/__init__.py empty-init!skip
```
tyll, line 1: just wondering: Is the filename ".sanity-ansible..." required by an Ansible project or is it a filename pattern that tox-lsr uses? It would  be nice to not have many files with the same or very similar content.

defaults/main.yml
```diff
@@ -2,24 +2,26 @@
 ---
 network_connections: []
 
-network_allow_restart: no
+network_allow_restart: false
+
+# List of RedHat family distributions supported by the role
```
tyll, line 7: It seems better to avoid "supported" here, since only RedHat and CentOS are supported. The others are recognized as RHEL clones without investing in supporting them.

tyll:
Thank you. It would be great to also update the README (and possibly other documentation) to use true/false instead of yes/no.

.ansible-lint
richm: > Is excludiding "tests" already needed at this point or only after the tox-lsr update?

It will be needed after the update to use ansible-core 2.14 and ansible-lint 6.x

> Do you have a list of failures for the tests directory? @liangwen12year can maybe address them.

It would probably be easier for @liangwen12year to run `tox -e ansible-lint` locally and see the list.  It is a pretty big list . . . nothing really complicated to fix, just a lot of tedious work.

.sanity-ansible-ignore-2.13.txt
richm: > just wondering: Is the filename ".sanity-ansible..." required by an Ansible project or is it a filename pattern that tox-lsr uses?

It is a filename pattern used by the tools that build the collection - https://github.com/linux-system-roles/auto-maintenance/blob/master/release_collection.py#L227 - we use these to build the `tests/sanity/ignore-xx.txt` files required by Ansible.

> It would be nice to not have many files with the same or very similar content.

There are version differences - for example, the 2.14 file must contain ignores for python-3.11, but earlier versions must not have the ignores for python-3.11.  So yes, we could have
* a config file that lists the ignores and suppressions in a more compact format
* a script that generates the correct ignore-xxx.txt format for each version
* a workflow to ensure that files are generated properly (e.g. git post-commit, github actions)

But at the moment that seems like a lot of work . . . otoh this would also be used by other roles that have to do something similar - I'll create a jira task to investigate doing this

defaults/main.yml
richm: > It seems better to avoid "supported" here, since only RedHat and CentOS are supported. The others are recognized as RHEL clones without investing in supporting them.

ok - what wording would you suggest?

tyll, line 7:
```suggestion
# Ansible distribution identifiers that the role treats like RHEL
```

.ansible-lint
tyll: > > Is excludiding "tests" already needed at this point or only after the tox-lsr update?
> 
> It will be needed after the update to use ansible-core 2.14 and ansible-lint 6.x

What is the timeline for this? Would it be ok to do the change to exclude `tests` together with the update? Or is this update happening only externally in the CI?

richm: > > > Is excludiding "tests" already needed at this point or only after the tox-lsr update?
> > 
> > 
> > It will be needed after the update to use ansible-core 2.14 and ansible-lint 6.x
> 
> What is the timeline for this?

QE is blocked - RHEL 8.8/9.2 have switched to ansible-core 2.14 - it isn't just the network role that has issues with ansible-core 2.14, but network QE is ready and waiting

> Would it be ok to do the change to exclude `tests` together with the update? Or is this update happening only externally in the CI?

Not sure what you mean.

richm:
> Thank you. It would be great to also update the README (and possibly other documentation) to use true/false instead of yes/no. @liangwen12year can you help with this?

done with the latest commit - I also made some more corrections in the examples

.ansible-lint
tyll: Thank you, now I understand what is updated. I was assuming it would be some configuration in the role's files or the CI. @liangwen12year please track the need to update the tests so it can be planned for in a sprint.

richm: Note that ansible-lint 6.x has an option `--write` - this will attempt to fix some of the issues - but you'll have to read the docs, and configure it carefully - by default it will change things like line wrapping which we do not want to touch.

richm:
[citest]

meta/main.yml
```diff
@@ -6,7 +6,6 @@ galaxy_info:
   company: Red Hat, Inc.
   license: BSD-3-Clause
   min_ansible_version: "2.9"
-  github_branch: main
```
liangwen12year, line 9: This is also not supported in ansible-core-2.14, ansible-lint 6.x ?

.sanity-ansible-ignore-2.13.txt
liangwen12year: I just noticed that the 2.11 file also contains ignores for python-3.11, how can we exactly know the version difference ?
 (e.g. ignores for python-3.11 )

meta/main.yml
richm: > This is also not supported in ansible-core-2.14, ansible-lint 6.x ?

Right.  It violates the schema of meta/main.yml - now enforced by ansible-lint 6.x

richm:
Any further comments?  Can this be merged?

.sanity-ansible-ignore-2.11.txt
```diff
@@ -6,41 +6,53 @@ plugins/module_utils/network_lsr/nm/__init__.py import-3.6!skip
...
+plugins/module_utils/network_lsr/nm/__init__.py import-3.11!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-2.6!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-2.7!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-3.5!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-3.6!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-3.7!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-3.8!skip
 plugins/module_utils/network_lsr/nm/active_connection.py import-3.9!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.10!skip
+plugins/module_utils/network_lsr/nm/active_connection.py import-3.11!skip
```
liangwen12year, line 19: Maybe this can be dropped, it seems to me that python 3.11 is not supported with ansible 2.11.

liangwen12year:
> Any further comments? Can this be merged?

I just added the comment, except for that the pr lgtm. Thanks.

richm:
> > Any further comments? Can this be merged?
> 
> I just added the comment, except for that the pr lgtm. Thanks.

Done.

liangwen12year (approved):
lgtm.

PR #540, opened 2022-11-15
add weekly-ci action

ukulekek:


ukulekek:
This action will trigger the CI system to test the role every weekend. It will ensure the role didn't break because of dependencies and other factors while being inactively developed. See https://url.corp.redhat.com/1ab3ab1 for more details

PR #542, opened 2022-11-28
python version depends on platform; upgrade checkout, setup-python; support py311 [citest skip]

richm:
The python version used now requires a corresponding os version e.g. python 2.7 and
python 3.6 are no longer supported on ubuntu-latest - must use 20.04.  Update
the python matrix to include the os to use as well.

Use checkout@v3 and setup-python@v4

python 3.11 stable is now supported by setup-python

Add `push` action for status reporting on role main page if missing

Use `docker` for ansible-test if not already doing that

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -12,17 +12,29 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 jobs:
   python:
-    runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


richm:
ok to merge?  If so, I will also fix the branch protection rules

liangwen12year:
> ok to merge? If so, I will also fix the branch protection rules

review now.

liangwen12year (approved):
lgtm.

PR #544, opened 2022-12-08
Use tox-lsr 2.13.1; force the use of tox 3.x [citest skip]

richm:
CI is broken pretty hard right now due to tox 4.x being released
It will take non-trivial effort to make tox-lsr work with tox 4.x. So,
in the meantime, tox-lsr 2.13.1 forces the use of tox 3.x


liangwen12year (approved):
lgtm.

PR #547, opened 2023-01-04
ansible-lint 6.x updates

richm:
The only thing we need to skip currently is using FQCN for ansible
builtin modules, plugins
Add `kinds` - otherwise, Ansible thinks anything not in a traditional
role path is a plain YAML file, and we don't get the additional
checking.
Ensure all plays are named.
Fix some other minor problems.

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 ---
-extends: yamllint_defaults.yml
+extends: default
 # possible customizations over the base yamllint config
 # skip the yaml files in the /tests/ directory
 # NOTE: If you want to customize `ignore` you'll have to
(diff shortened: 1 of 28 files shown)


richm:
[citest]

.yamllint.yml
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 ---
-extends: yamllint_defaults.yml
+extends: default
```
liangwen12year, line 3: what is the reason switching from `extends: yamllint_defaults.yml` into `extends: default` ?

richm: We're getting rid of `yamllint_defaults.yml` and we'll just use the built-in defaults in `default`

liangwen12year (approved):
lgtm.

PR #549, opened 2023-01-09
Support running the tests with ANSIBLE_GATHERING=explicit

liangwen12year:
Almost all the network role tests fail with fact gathering disabled. Therefore, in order to support the tests running with ANSIBLE_GATHERING=explicit, gather the minimum subset of facts required.

### DIFF
--- a/tests/tasks/el_repo_setup.yml
+++ b/tests/tasks/el_repo_setup.yml
@@ -1,4 +1,17 @@
 # SPDX-License-Identifier: BSD-3-Clause
+- name: Gather the minimum subset of ansible_facts required by the network
+    role test
+  setup:
+    gather_subset: min
+  when: not ansible_facts.keys() | list |
+    intersect(network_test_required_facts) == network_test_required_facts
+  vars:
+    network_test_required_facts:
+      - distribution
+      - distribution_major_version
+      - distribution_version
+      - os_family
+
 - name: Fix CentOS6 Base repo
   copy:
     dest: /etc/yum.repos.d/CentOS-Base.repo


tests/tasks/el_repo_setup.yml
```diff
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
+- name: Gather the minimum subset of ansible_facts required
+  setup:
+    gather_subset: min
```
richm, line 4: please add a `when` condition like this https://github.com/linux-system-roles/network/blob/main/tasks/set_facts.yml#L6
gathering facts is an expensive operation, even for `gather_subset: min` - if the playbook already gathered facts, you do not want to gather them again

tests/tasks/el_repo_setup.yml
```diff
@@ -1,4 +1,19 @@
...
+      - distribution_major_version
+      - distribution_version
+      - os_family
+
+- name: Gather the minimum subset of ansible_facts required by the network
+    role test
+  setup:
+    gather_subset: min
+  when: not ansible_facts.keys() | list |
+    intersect(network_test_required_facts) == network_test_required_facts
```
richm, line 15:
```yaml
  vars:
    network_test_required_facts:
      - distribution
      - distribution_major_version
      - distribution_version
      - os_family
```
unless you need `network_test_required_facts` to be a global variable, use it elsewhere

richm (approved):
lgtm

ffmancera (approved):

PR #552, opened 2023-01-20
ansible-lint: remediate the test failures

liangwen12year:


liangwen12year:
All the patches are committed, I will resolve the conflict and squash in the second step.

.ansible-lint
```diff
@@ -13,12 +13,16 @@ kinds:
   - playbook: "**/examples/*.yml"
 skip_list:
   - fqcn-builtins
+  - name[play]
+  - key-order[task]
 warn_list:
   - load-failure
+  - parser-error  # AnsibleParserError.
```
richm, line 20: Add `- playbook: "**/tests/get_coverage.yml" to `kinds` instead

richm: I think this will fix the `schema[tasks]` issue too (the error is because ansible-lint thinks `tests/get_coverage.yml` is a tasks file instead of a playbook

richm: and may fix the `schema[playbook]` issue as well

tasks/main.yml
```diff
@@ -113,7 +113,7 @@
   when:
     - network_provider == "initscripts"
 
-- name: Configure networking connection profiles
+- name: Configure networking connection profiles  # noqa jinja[invalid]
```
richm, line 116: What is the error here?  The jinja to me looks valid

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -32,22 +32,22 @@
   hosts: all
```
richm, line 32: The error in this file is here:
```yaml
- name: Install dependencies for integration tests
  hosts: all
  vars:
    - rpmdependencies:
        - git
        - python3-pip
        - rsync
```
`vars` should be a `dict`, not a `list`, so
```yaml
- name: Install dependencies for integration tests
  hosts: all
  vars:
    rpmdependencies:
      - git
      - python3-pip
      - rsync
```

tests/playbooks/integration_pytest_python3.yml
```diff
@@ -73,19 +73,21 @@
...
 
-        - name: Untar testrepo.tar
+        - name: Untar testrepo.tar  # noqa command-instead-of-module
           command: tar xf {{ temptar.path | basename }}
           args:
             chdir: "{{ _rundir.path }}"
+          changed_when: false
 
-        - name: "Create {{ _rundir.path }}/ansible"
+        - name: "Create {{ _rundir.path }}/ansible"  # noqa name
```
richm, line 90:
```suggestion
        - name: Create /ansible directory in {{ _rundir.path }}
```

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -132,28 +136,34 @@
...
-                    ip:
-                      dhcp4: "no"
-                      auto6: "no"
-                    ethtool:
-                      features:
-                        tx_tcp_segmentation: "no"
-                        tx-tcp-segmentation: "no"
-                __header: "# Ansible managed test header"
-              register: __network_connections_result
+            - name: Import network role
```
richm, line 139: This changes the semantic - before, the test is calling the `network_connections` module directly - with the change you are calling the role (which calls the module) - does this work exactly the same way?

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -132,28 +136,34 @@
...
+                      state: up
+                      type: ethernet
+                      ip:
+                        dhcp4: "no"
+                        auto6: "no"
+                      ethtool:
+                        features:
+                          tx_tcp_segmentation: "no"
+                          tx-tcp-segmentation: "no"
+                  __header: "# Ansible managed test header"
```
richm, line 156:
```suggestion
```
The reason this was here was that the old code was calling the module directly, which needed to add the `__header` - since you changed it to use the role, you should omit `__header`, since the role will add it

tests/playbooks/tests_ethtool_features.yml
```diff
@@ -132,28 +136,34 @@
...
-            - assert:
+            - name: Assert the duplicate key "tx_tcp_segmentation"
+                configuration error happened
+              assert:
                 that:
-                  - '{{  "fatal error: configuration error:
+                  - '{{ "fatal error: configuration error:
                       connections[0].ethtool.features: duplicate key
                       ''tx_tcp_segmentation''" in
                       __network_connections_result.msg }}'
```
richm, line 169:
```yaml
              vars:
                errmsg: >-
                  fatal error: configuration error: connections[0].ethtool.features:
                  duplicate key 'tx_tcp_segmentation'
              assert:
                that:
                  - errmsg in __network_connections_result.msg
```

tests/tests_eth_pci_address_match_nm.yml
```diff
@@ -5,21 +5,21 @@
...
+    - block:  # noqa name
         - name: Install NetworkManager
           package:
             name: NetworkManager
             state: present
-        - name: Get NetworkManager version
+        - name: Get NetworkManager version  # noqa command-instead-of-module
           command: rpm -q --qf "%{version}" NetworkManager
-          register: NetworkManager_version
+          register: networkmanager_version
```
richm, line 22: Where is this used?

tests/tests_ethtool_coalesce_nm.yml
```diff
@@ -5,21 +5,21 @@
...
+    - block:  # noqa name
         - name: Install NetworkManager
           package:
             name: NetworkManager
             state: present
-        - name: Get NetworkManager version
+        - name: Get NetworkManager version  # noqa command-instead-of-module
           command: rpm -q --qf "%{version}" NetworkManager
-          register: NetworkManager_version
+          register: networkmanager_version
```
richm, line 22: where is this used?

tests/tests_ethtool_features_nm.yml
```diff
@@ -5,21 +5,21 @@
...
+    - block:  # noqa name
         - name: Install NetworkManager
           package:
             name: NetworkManager
             state: present
-        - name: Get NetworkManager version
+        - name: Get NetworkManager version  # noqa command-instead-of-module
           command: rpm -q --qf "%{version}" NetworkManager
-          register: NetworkManager_version
+          register: networkmanager_version
```
richm, line 22: where is this used?

tests/tests_ethtool_ring_nm.yml
```diff
@@ -5,21 +5,21 @@
...
+    - block:  # noqa name
         - name: Install NetworkManager
           package:
             name: NetworkManager
             state: present
-        - name: Get NetworkManager version
+        - name: Get NetworkManager version  # noqa command-instead-of-module
           command: rpm -q --qf "%{version}" NetworkManager
-          register: NetworkManager_version
+          register: networkmanager_version
```
richm, line 22: where is this used?

tests/tests_provider_nm.yml
```diff
@@ -5,21 +5,21 @@
...
+    - block:  # noqa name
         - name: Install NetworkManager
           package:
             name: NetworkManager
             state: present
-        - name: Get NetworkManager version
+        - name: Get NetworkManager version  # noqa command-instead-of-module
           command: rpm -q --qf "%{version}" NetworkManager
-          register: NetworkManager_version
+          register: networkmanager_version
```
richm, line 22: where is this used?

liangwen12year: It is used in - networkmanager_version.stdout is version({minimum_nm_version}, '>=').

tests/tests_ethtool_ring_nm.yml
liangwen12year: It is used in - networkmanager_version.stdout is version({minimum_nm_version}, '>=').

tests/tests_provider_nm.yml
richm: ok - because https://github.com/linux-system-roles/network/blob/54c556c06096eb29ce6a97a1f3d07fd17609a5e3/tests/tests_provider_nm.yml#L37 is still using `NetworkManager_version`

tasks/main.yml
liangwen12year:
```
jinja[invalid]: An unhandled exception occurred while templating '{{ lookup('template', 'get_ansible_managed.j2') }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'template'. Error was a <class 'ansible.errors.AnsibleError'>, original message: the template file get_ansible_managed.j2 could not be found for the lookup. the template file get_ansible_managed.j2 could not be found for the lookup
~/network_wl_ansible_lint/network/tasks/main.yml:116 Task/Handler: Configure networking connection profiles
```

richm: What version of ansible-lint are you using?  How are you running it?

tests/ensure_provider_tests.py
```diff
@@ -39,7 +39,7 @@
 - hosts: all
   name: Run playbook '{test_playbook}' with nm as provider
   tasks:
-    - include_tasks: tasks/el_repo_setup.yml
+    - include_tasks: tasks/el_repo_setup.yml  # noqa name
```
tyll, line 42: It seems to me that adding a name for these tasks would not hurt.

tests/playbooks/tests_ethtool_features.yml
tyll: IIRC, calling the module was done as a proof-of-concept to speed up testing since we do not need to run the whole role for every test.

tests/playbooks/tests_auto_gateway.yml
```diff
@@ -91,16 +91,16 @@
             type: ethernet
             state: up
             ip:
-              auto_gateway: no
-              dhcp4: no
-              auto6: no
+              auto_gateway: false
+              dhcp4: false
+              auto6: false
```
tyll, line 96: IMHO it would be nicer to have individual commits for different type of changes, then you can move forward with those that are straight-forward like always using `true`/`false` for booleans. Put this in a separate PR and it can be merged IMHO.

tests/playbooks/tests_bond_options.yml
```diff
@@ -83,6 +83,7 @@
             /sys/class/net/{{ controller_device }}/bonding/'{{ item.key }}'
           name: "** TEST check bond settings"
           register: result
+          changed_when: false
```
tyll, line 86: The `changed_when` changes also seem to be a good candidate for one commit to be merged sooner.

tests/playbooks/tests_bridge_cloned_mac.yml
```diff
@@ -3,7 +3,7 @@
 - name: Test configuring bridges
   hosts: all
   tasks:
-    - name: define vars
+    - name: Define vars
```
tyll, line 6: Using upper-case for names is another candidate for an individual commit to be merged.

tests/ensure_provider_tests.py
```diff
@@ -11,14 +11,14 @@
 
 
 GET_NM_VERSION = """
-    - block:
+    - block:  # noqa name
```
tyll, line 14: I guess you could easily add a name to the block and it would also help.

tyll: The changes to the script and all changes that result from running it should be its own commit, too (unless it fites already in another category) to avoid a review of dozens of generated changes.

tasks/main.yml
liangwen12year: The version is ansible-lint 6.11.0. `ansible-lint -v`

liangwen12year: I upgraded the ansible-lint into 6.12.1, then I did not see the error.

richm: ok - then please remove the noqa comment

tests/ensure_provider_tests.py
richm: Basically, change `tests/ensure_provider_tests.py` so that it produces code that is ansible-lint clean, rather than having it generate code with ansible-lint exceptions - this applies to this `noqa name` and any other exceptions.

tyll: yes, and move this and related changes to their own commit, so they can independently be reviewed. Reviewing the changes for the script and the resulting generated code together with small changes might result in small, unrelated changes to be missed.

PR #554, opened 2023-01-21
tag all bond tests with expfail

richm:
The bond tests fail regularly in our single host downstream testing,
so mark them as `expfail` (expect to fail) so that these failures
do not clutter up our test results.

We have tried to fix this in the past - several times over the past
three years, at a cost of many man weeks of effort  - and the fixes just
don't seem to "take".  We have reached the point where we need to cut
our losses and just skip these tests.  Perhaps at some point in the
future we can revisit this issue.


richm:
Can this be merged?  Any objection?

liangwen12year (approved):
The  changes look good to me.

PR #557, opened 2023-01-25
address shellcheck issues

richm:
Use `grep -E` instead of deprecated `egrep`
Ignore the "use double quotes" rule because we want the
variables to be expanded with all of the spaces.


liangwen12year (approved):
LGTM.

PR #558, opened 2023-01-30
Address CodeQL warnings

nhosoi:
library/network_connections.py
- 'except' clause does nothing but pass and there is no explanatory comment.
    
module_utils/network_lsr/argument_validator.py
- 'except' clause does nothing but pass and there is no explanatory comment.
    
module_utils/network_lsr/utils.py
- Mixing implicit and explicit returns may indicate an error as implicit returns always return None.
    
tests/unit/test_network_connections.py
- Module 'network_lsr.argument_validator' is imported with both 'import' and 'import from'.
- assertTrue(a == b) cannot provide an informative message. Using assertEqual(a, b) instead will give more informative messages.

### DIFF
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -818,12 +818,14 @@ class NMUtil:
             try:
                 con_a.normalize()
             except Exception:
+                # We ignore any errors that might happen
                 pass
         if normalize_b:
             con_b = NM.SimpleConnection.new_clone(con_b)
             try:
                 con_b.normalize()
             except Exception:
+                # We ignore any errors that might happen
                 pass
         if compare_flags is None:
             compare_flags = NM.SettingCompareFlags.IGNORE_TIMESTAMP
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -258,20 +258,17 @@ class ArgValidatorRouteTable(ArgValidator):
 
     def _validate_impl(self, value, name):
         table = None
-        try:
-            if isinstance(value, bool):
-                # bool can (probably) be converted to integer type,
-                # but here we don't want to accept a boolean value.
-                pass
-            elif isinstance(value, int):
-                table = int(value)
-            elif isinstance(value, Util.STRING_TYPE):
-                try:
-                    table = int(value)
-                except Exception:
-                    table = value
-        except Exception:
+        if isinstance(value, bool):
+            # bool can (probably) be converted to integer type,
+            # but here we don't want to accept a boolean value.
             pass
+        elif isinstance(value, int):
+            table = value
+        elif isinstance(value, Util.STRING_TYPE):
+            try:
+                table = int(value)
+            except Exception:
+                table = value
         if table is None:
             raise ValidationError(
                 name,
@@ -336,6 +333,7 @@ class ArgValidatorNum(ArgValidator):
                 if isinstance(value, Util.STRING_TYPE) or v2 == value:
                     v = v2
         except Exception:
+            # Exception handling is done next.
             pass
         if v is None:
             raise ValidationError(
@@ -374,11 +372,13 @@ class ArgValidatorRange(ArgValidator):
                 try:
                     range = (int(match_group.group(1)), int(match_group.group(2)))
                 except Exception:
+                    # Exception handling is done below.
                     pass
             else:
                 try:
                     range = (int(value), int(value))
                 except Exception:
+                    # Exception handling is done below.
                     pass
         elif isinstance(value, bool):
             # bool can (probably) be converted to integer type,
@@ -425,6 +425,7 @@ class ArgValidatorBool(ArgValidator):
             if isinstance(value, Util.STRING_TYPE) or isinstance(value, int):
                 return Util.boolean(value)
         except Exception:
+            # Exception handling is done next.
             pass
         raise ValidationError(name, "must be an boolean but is '%s'" % (value))
 
@@ -2665,6 +2666,7 @@ class IPRouteUtils(object):
                     try:
                         tableid = int(table[2:], 16)
                     except Exception:
+                        # Exception handling is done next.
                         pass
             if tableid is None or tableid < 0 or tableid > 0xFFFFFFFF:
                 continue
--- a/tests/unit/test_network_connections.py
+++ b/tests/unit/test_network_connections.py
@@ -19,10 +19,8 @@ sys.modules["ansible.module_utils.basic"] = mock.Mock()
 
 # pylint: disable=import-error, wrong-import-position
 
-import network_lsr
 import network_lsr.argument_validator
 from network_connections import IfcfgUtil, NMUtil, SysUtil, Util
-from network_lsr.argument_validator import ValidationError
 
 try:
     my_test_skipIf = unittest.skipIf
@@ -215,7 +213,9 @@ class TestValidator(Python26CompatTestCase):
         }
 
     def assertValidationError(self, v, value):
-        self.assertRaises(ValidationError, v.validate, value)
+        self.assertRaises(
+            network_lsr.argument_validator.ValidationError, v.validate, value
+        )
 
     def assert_nm_connection_routes_expected(self, connection, route_list_expected):
         parser = network_lsr.argument_validator.ArgValidatorIPRoute("route[?]")
@@ -257,7 +257,7 @@ class TestValidator(Python26CompatTestCase):
         for idx, connection in enumerate(connections):
             try:
                 ARGS_CONNECTIONS.validate_connection_one(mode, connections, idx)
-            except ValidationError:
+            except network_lsr.argument_validator.ValidationError:
                 continue
             if "type" in connection:
                 con_new = nmutil.connection_create(connections, idx)
@@ -304,7 +304,7 @@ class TestValidator(Python26CompatTestCase):
         for idx, connection in enumerate(connections):
             try:
                 ARGS_CONNECTIONS.validate_connection_one(mode, connections, idx)
-            except ValidationError:
+            except network_lsr.argument_validator.ValidationError:
                 continue
             if "type" not in connection:
                 continue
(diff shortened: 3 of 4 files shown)


nhosoi:
Hmmm, maybe I'd better close this pr since the network role does not have the codeql action added. What do you think, @richm?  If so, sorry for the noise...

richm:
> Hmmm, maybe I'd better close this pr since the network role does not have the codeql action added. What do you think, @richm? If so, sorry for the noise...

I think the plan is for network to drop lgtm and use codeql, so they will need these fixes (unless they want to fix the code in some other way) - I guess it's up to @liangwen12year and the rest of the network team to decide what they want to do.

tests/unit/test_network_connections.py
```diff
@@ -3461,15 +3463,15 @@ def test_interface_name_ethernet_default(self):
         """Use the profile name as interface_name for ethernet profiles"""
         cons_without_interface_name = [{"name": "eth0", "type": "ethernet"}]
         connections = ARGS_CONNECTIONS.validate(cons_without_interface_name)
-        self.assertTrue(connections[0]["interface_name"] == "eth0")
+        self.assertEqual(connections[0]["interface_name"], "eth0")
```
tyll, line 3466: https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertIs  this is new in Python 2.7 - in the past we needed to support Python 2.6 - is this not necessary anymore?

tyll: https://github.com/linux-system-roles/network/blob/main/contributing.md#where-to-start needs to be updated if we don't need to support Python 2.6 anymore.

richm: I'm not sure if we still need to support 2.6, but if we don't, it should be a conscious, deliberate decision, in a separate PR.  So for this change, please use a method that is compatible with 2.6

library/network_connections.py
```diff
@@ -818,12 +818,14 @@ def connection_compare(
             try:
                 con_a.normalize()
             except Exception:
+                # It's already checked normalize is successful.
```
tyll, line 821: @thom311 what kind of exceptions are ignored here?

nhosoi:
Thank you, @tyll, @richm.
I'm reverting the changes complained about by the py26, and py27 tests.

library/network_connections.py
```diff
@@ -2550,6 +2552,8 @@ def forget_nm_connection(self, path):
                 ]
             )
         except Exception:
+            # Assume that NetworkManager is not present.
```
tyll, line 2555: This is duplicating the doc string for the function, so IMHO it would be better to remove this and/or to properly ignore only the error cases when busctl is not available or if the NetworkManager dbus is not available instead of ignoring all exceptions. I feel  guilty since this is technical debt from me. :see_no_evil:

tests/unit/test_network_connections.py
```diff
@@ -19,10 +19,10 @@
 
 # pylint: disable=import-error, wrong-import-position
 
-import network_lsr
 import network_lsr.argument_validator
 from network_connections import IfcfgUtil, NMUtil, SysUtil, Util
-from network_lsr.argument_validator import ValidationError
+
+# from network_lsr.argument_validator import ValidationError
```
tyll, line 25: Is this comment intentional?

tests/unit/test_network_connections.py
```diff
@@ -215,7 +215,9 @@ def setUp(self):
         }
 
     def assertValidationError(self, v, value):
-        self.assertRaises(ValidationError, v.validate, value)
+        self.assertRaises(
+            network_lsr.argument_validator.ValidationError, v.validate, value
+        )
```
tyll, line 220: If others agree with this change, please put it into its own commit/PR, this seems to be a valid change that could be merged - also the other, similar lines. :+1:

richm: Is this change due to a CodeQL error?  If so, did CodeQL suggest this as the correction?

library/network_connections.py
```diff
@@ -1493,8 +1495,7 @@ def check_activated(ac, dev):
             except AttributeError:
                 ac_reason = None
 
-            if dev:
-                dev_state = dev.get_state()
+            dev_state = dev.get_state() if dev else None
```
tyll, line 1498: I believe my previous comment was lost, if not, please excuse the duplicate.
CodeQL properly identifies this code as problematic. This change only makes CodeQL happy but does not really work towards refactoring the code to make it easier to understand. In the following lines `dev` and `dev_state` are  used without checking whether they are None and I guess it is also not needed since there is some implicit understanding whether this function will actually be called with `dev` not being a true value. I would rather see this function being properly refactored to make full use of CodeQL and creating additional benefit for the future.

module_utils/network_lsr/argument_validator.py
```diff
@@ -271,6 +271,7 @@ def _validate_impl(self, value, name):
                 except Exception:
                     table = value
         except Exception:
```
tyll, line 273: @thom311 @liangwen12year what exceptions are we excepting here? Is it a ValueError from line 267? It seems like it could be excepted there specifically.

library/network_connections.py
nhosoi: Thanks for the detailed explanation, @tyll ! Reverted.

tests/unit/test_network_connections.py
nhosoi: CodeQL complains about duplicate import statements: `Module 'network_lsr.argument_validator' is imported with both 'import' and 'import from'.` And no, CodeQL did not give us suggestions.

library/network_connections.py
nhosoi: @liangwen12year, @thom311, could you please take a look at this exception? Thanks!

tests/unit/test_network_connections.py
nhosoi: No, thanks for finding it out!

library/network_connections.py
thom311: Do you mean, which exceptions `normalize()` is expected to throw? Probably they are all subtypes of `GLib.Error`. But I don't know, and I don't want to find out -- because even if I check *now*, it's not clear that this check will hold for the future. The code *really* wants to catch and ignore *any* errors (that is, all `Exception` types -- but not all `BaseException`).

The code seems pretty clear in that. This is not a bug but done unintentionally. The code wants to normalize the profile, and if that fails, ignore the error. *Any* error.

The new code comment seems not clear to me. The proper comment would be `# We ignore any errors that might happen`, which seems a pretty meaningless thing to say. This linter rule seems questionable to me, but OK, it needs to be worked around.

---

In the commit message, I would not say "fix issues" (but false positives or warnings) -- or does the patch also fix actual bugs? In that case, the actual bug fixes should be separate commits.

module_utils/network_lsr/argument_validator.py
thom311: I don't think any exception can be raised here. The outer try/except should be dropped.

library/network_connections.py
tyll: I mean what can go wrong with normalization that it can be ignored but the connection is still fine without normalization so that we can safely ignore the error.

thom311: > In the commit message, I would not say "fix issues" (but false positives or warnings) -- or does the patch also fix actual bugs? In that case, the actual bug fixes should be separate commits.

Well, maybe "issues" means something quite benign (like a warning). In my ear, it sounds severe to fix "issues". I would not use the wording "fix issue" for adding a code comment. If the patch fixes actual issues, this patch should not be cluttered with low-severity changes. If it doesn't fix any bugs, the word "issue" seems unnecessarily alarming.

module_utils/network_lsr/argument_validator.py
tyll: Ah, I see now. So `table = value` would be better in line 267, wouldn't it?

liangwen12year: If the value is already the int type, then the function `int()` seems to me is not needed.

module_utils/network_lsr/argument_validator.py
```diff
@@ -2665,6 +2666,7 @@ def _parse_route_tables_mapping(cls, file_content, mapping):
                     try:
                         tableid = int(table[2:], 16)
                     except Exception:
+                        # Exception handling is done next.
```
tyll, line 2669: Which exception is expected here? The regex "(0x[0-9a-fA-F]+|[0-9]+)" seems to be sufficient to protect here. This could be simplified to

```python
if table.startswith(b"0x"):
    tableid = int(table, 16)
else:
    tableid = int(table)
 if tableid > 0xFFFFFFFF:
    continue
```
The regex does not seem to allow negative values. To re-iterate my previous point, these small changes to please CodeQL should be reviewed indepedently and the straight-forward changes such as the import changes in the unit tests should move to their own commit in their own PR so they can be merged.

library/network_connections.py
```diff
@@ -818,12 +818,14 @@ def connection_compare(
             try:
                 con_a.normalize()
             except Exception:
+                # We ignore any errors that might happen
```
tyll, line 821: This comment does not provide any additional value since it just describes the code. The comment should include the reason why it is the right thing to do this here.

module_utils/network_lsr/argument_validator.py
liangwen12year: Thanks for the proposal of code changes, IMHO, the exception here can be dropped (no exception).

tyll: and it should be its own commit with appropriate commit message.

tyll: and also note, that the slice in `table[2:]` is not necessary, `int()` takes care of this already.

library/network_connections.py
nhosoi: > This comment does not provide any additional value since it just describes the code. The comment should include the reason why it is the right thing to do this here.

https://github.com/linux-system-roles/network/pull/558#discussion_r1101921414
> The proper comment would be # We ignore any errors that might happen, which seems a pretty meaningless thing to say. This linter rule seems questionable to me, but OK, it needs to be worked around.

We could just say `# Workaround CodeQL warning`, then?

nhosoi:
Thank you for the discussions, @tyll, @richm, @liangwen12year, @thom311.
Instead of discussing each warning in this PR, I believe it should be much faster and cleaner for an expert to clean up the code. So, I'm going to close this PR for now, and open an issue with the raw output from CodeQL pointing to this PR.

module_utils/network_lsr/argument_validator.py
thom311: `isinstance()` does not check whether it's the same type, but of the type or a subtype.

the point of the code is of course to ensure that the result is really an int, and not a subtype.

"not needed", is if we never call the function with a subtype of int (`bool` is also a subtype of `int`, but already handled above).

I would leave it.

tyll: What's the problem if it is a subclass? Regarding `bool`, the check is there to avoid accepting boolean values but other subclasses are not rejected, so it is not clear to me why this would be useful.

module_utils/network_lsr/argument_validator.py
tyll: Submitted the change in #566

module_utils/network_lsr/argument_validator.py
thom311: the idea is that the validation+normalization returns ... normalized values. That is, `int` and not a subclass of an `int` that was passed as input to the validation.

Anyway, since nobody is actually subclassing `int` (aside `bool`, which is handled already), it doesn't have any effect.

tyll: Normalization can also mean to normalize it to `int` or a subclass of it.

tests/unit/test_network_connections.py
tyll: https://github.com/linux-system-roles/network/security/code-scanning/13 has the suggestion to do

```
import network_lsr
ValidationError = network_lsr.argument_validator.ValidationError
```

to avoid importing it twice.

PR #564, opened 2023-02-09
mdl_style: Fix rule URLs to use main branch

tyll:


tyll:
Thanks @richm, it seems the markdownlint job is broken.

PR #568, opened 2023-02-10
add back python-26 unit test

richm:
The PR https://github.com/linux-system-roles/network/pull/556
erroneously removed the python-26 job for unit testing.  This
PR adds it back.

See also https://github.com/linux-system-roles/.github/pull/10

### DIFF
--- a/.github/workflows/python-unit-test.yml
+++ b/.github/workflows/python-unit-test.yml
@@ -69,3 +69,13 @@ jobs:
           *) toxenvs="${toxenvs},coveralls" ;;
           esac
           TOXENV="$toxenvs" lsr_ci_runtox
+  python-26:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout PR
+        uses: actions/checkout@v3
+
+      - name: Run py26 tests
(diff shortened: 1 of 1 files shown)


richm:
Using the new codecov github app integration too I see

tyll (commented):
LGTM :+1:

liangwen12year (approved):
Looks good.

richm:
there should be a way to not require integration tests for changes like this . . . of course you could say the same for any other check that doesn't have to run, but the integration tests waste a lot more time and resources . . .

tyll:
> there should be a way to not require integration tests for changes like this . . . of course you could say the same for any other check that doesn't have to run, but the integration tests waste a lot more time and resources . . .

Some automated tools detect themselves if there are relevant changes, so it would be nice to implement an ignorelist for the different test runners and if the changes only match the ignorelist, the tests are skipped. It would also be possible to use LD_PRELOAD to wrap open/stat to detect which files are accessed by which test run and create a dependency map and use that. Maybe these could be some GSoC projects.

PR #571, opened 2023-02-13
Fix CentOS 7 integration tests

liangwen12year:

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   integration:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:


.github/workflows/integration.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   integration:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
tyll, line 12: Please provide an explanation to the commit message why this is the correct change.

liangwen12year: Updated.

tyll (approved):
:+1: LGTM

liangwen12year:
Notice that, this PR will make all the c7, c8s, c9s integration tests  run on the ubuntu-20.04 runner image, which is not the ideal case. Because we still want the c8s, c9s integration tests  run on the ubuntu-latest runner image.  Later on, we may add patch to improve it.

PR #573, opened 2023-02-13
initscripts: Configure output device in routes

liangwen12year:
Without an explicit output device, the kernel might use a different
    output device than intended by the user. Therefore, use the interface
    name of connections to specify it if it is available. Otherwise,
    educate the user about this potential problem with a warning. This
    aligns the behavior with NetworkManager which configures the output
    device in routes when activating profiles on devices.
    
    Fixes: https://bugzilla.redhat.com/2168735


liangwen12year:
I will add the test case soon.

library/network_connections.py
```diff
@@ -2511,6 +2513,11 @@ def run_prepare(self):
                     "unsupported type %s for initscripts provider"
                     % (connection["type"]),
                 )
+            if not connection["interface_name"] and connection["ip"]["route"]:
+                self.log_warn(
+                    idx,
+                    "the route will be configured without the 'dev' attribute",
+                )
```
tyll, line 2520: What's the reason to put the warning here instead of in an else clause above? Also, the message needs to be more specific so it is clear which route is meant and @ffmancera proposed some different wording to explain the behavior.

liangwen12year: I put it here because I found out almost all the loggings for the initscripts provider was put in the `Cmd_initscripts` class rather than  `IfcfgUtil.ifcfg_create()` function.

liangwen12year:
[citest]

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- hosts: all
```
richm, line 3:
```suggestion
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+
+- name: Test configuring ethernet devices
+  hosts: all
+  vars:
+    type: veth
+    interface0: ethtest0
+    interface1: ethtest1
+
+  tasks:
+    - name: "set type={{ type }} and interface={{ interface0 }}"
```
richm, line 13:
```suggestion
    - name: Set type and interface
```
might as well try to make the code ansible-lint clean, even if the other tests don't pass ansible-lint

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+      set_fact:
+        type: "{{ type }}"
+        interface: "{{ interface0 }}"
+    - include_tasks: tasks/show_interfaces.yml
```
richm, line 17:
```suggestion
    - name: Show interfaces
      include_tasks: tasks/show_interfaces.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+    - include_tasks: tasks/manage_test_interface.yml
```
richm, line 18:
```suggestion
    - name: Manage test interface
      include_tasks: tasks/manage_test_interface.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+      vars:
+        state: present
+    - include_tasks: tasks/assert_device_present.yml
```
richm, line 21:
```suggestion
    - name: Assert device is present
      include_tasks: tasks/assert_device_present.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+    - name: "set type={{ type }} and interface={{ interface1 }}"
```
richm, line 22:
```suggestion
    - name: Set type and interface1
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+      set_fact:
+        type: "{{ type }}"
+        interface: "{{ interface1 }}"
+    - include_tasks: tasks/show_interfaces.yml
```
richm, line 26:
```suggestion
    - name: Show interfaces
      include_tasks: tasks/show_interfaces.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+    - include_tasks: tasks/manage_test_interface.yml
```
richm, line 27:
```suggestion
    - name: Manage test interface
      include_tasks: tasks/manage_test_interface.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+      vars:
+        state: present
+    - include_tasks: tasks/assert_device_present.yml
```
richm, line 30:
```suggestion
    - name: Assert device is present
      include_tasks: tasks/assert_device_present.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+    - block:
```
richm, line 31:
```suggestion
    - name: Run the test
      block:
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+        - name: Configure the IP addresses
+          import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: "{{ interface0 }}"
+                interface_name: "{{ interface0 }}"
+                state: up
+                type: ethernet
+                autoconnect: yes
```
richm, line 41:
```suggestion
                autoconnect: true
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+                ip:
+                  dhcp4: no
```
richm, line 43:
```suggestion
                  dhcp4: false
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+                      metric: 4
+                    - network: 2001:db6::4
+                      prefix: 128
+                      gateway: 2001:db8::1
+                      metric: 2
+              - name: "{{ interface1 }}"
+                interface_name: "{{ interface1 }}"
+                state: up
+                type: ethernet
+                autoconnect: yes
```
richm, line 60:
```suggestion
                autoconnect: true
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+                ip:
+                  dhcp4: no
```
richm, line 62:
```suggestion
                  dhcp4: false
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+        - name: Assert that the route table main contains the specified IPv6
+            routes
+          assert:
+            that:
+              - route_table_main_ipv6.stdout is search("2001:db6::4 via
+                2001:db8::1 dev ethtest0 metric 2")
+            msg: "the route table main does not contain the specified IPv6
+              route"
+      always:
+        - block:
```
richm, line 103:
```suggestion
        - name: Remove test configuration
          block:
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+            - include_role:
```
richm, line 104:
```suggestion
            - name: Shut down test devices
              include_role:
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+              vars:
+                network_connections:
+                  - name: "{{ interface0 }}"
+                    persistent_state: absent
+                    state: down
+                  - name: "{{ interface1 }}"
+                    persistent_state: absent
+                    state: down
+              ignore_errors: true
+            - include_tasks: tasks/assert_profile_absent.yml
```
richm, line 115:
```suggestion
            - name: Assert test devices are down
              include_tasks: tasks/assert_profile_absent.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+              vars:
+                profile: "{{ interface0 }}"
+            - include_tasks: tasks/assert_profile_absent.yml
```
richm, line 118:
```suggestion
            - name: Assert test profiles are absent
              include_tasks: tasks/assert_profile_absent.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+              vars:
+                profile: "{{ interface1 }}"
+            - include_tasks: tasks/delete_interface.yml
```
richm, line 121:
```suggestion
            - name: Delete interfaces
              include_tasks: tasks/delete_interface.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+            - include_tasks: tasks/assert_device_absent.yml
```
richm, line 122:
```suggestion
            - name: Assert devices are absent
              include_tasks: tasks/assert_device_absent.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+            - name: "set type={{ type }} and interface={{ interface }}"
```
richm, line 123:
```suggestion
            - name: Set type and interface0
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+              set_fact:
+                type: "{{ type }}"
+                interface: "{{ interface0 }}"
+            - include_tasks: tasks/delete_interface.yml
```
richm, line 127:
```suggestion
            - name: Delete interface0
              include_tasks: tasks/delete_interface.yml
```

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,131 @@
...
+            - include_tasks: tasks/assert_device_absent.yml
```
richm, line 128:
```suggestion
            - name: Assert interface0 is absent
              include_tasks: tasks/assert_device_absent.yml
```

liangwen12year:
Thanks for commenting, I was just testing the CI, maybe I should mark it as "Draft".

richm:
> Thanks for commenting, I was just testing the CI, maybe I should mark it as "Draft".

https://github.com/linux-system-roles/network/pull/573#discussion_r1106468534 is likely to cause CI issues - not sure if 
```yaml
- hosts: all

- hosts: all
  ...
```
is valid Ansible

liangwen12year:
[citest]

library/network_connections.py
```diff
@@ -531,6 +531,16 @@ def ifcfg_create(
                 line = r["network"] + "/" + str(r["prefix"])
                 if r["gateway"]:
                     line += " via " + r["gateway"]
+                if connection["interface_name"]:
```
cathay4t, line 534: I am not sure whether ansible always set `interface_name` or not, but this might lead to `KeyError`. Try `connection.get("interface_name")`

cathay4t (changes requested):

library/network_connections.py
ffmancera: The new warning message is fine for me :-)

tyll:
The warning message is missing an explanation for the use what triggered
the warning and a recommendation what to do to get rid of it.

library/network_connections.py
```diff
@@ -531,6 +531,16 @@ def ifcfg_create(
...
+                    line += " dev " + connection["interface_name"]
+                else:
+                    warn_fcn(
+                        "the route '{0}' from the connection '{1}' will be configured "
```
tyll, line 538: The message is missing the reason for this and recommendations how to get rid of the warning:

"The connection ... does not specify an interface name. Therefore, the route ... will be configured without the output device and the kernel will choose it automatically which might result in an unwanted device being used. To avoid this, specify `interface_name` in the connection appropriately."

richm:
@liangwen12year this test is failing:
```
TASK [Assert that the route table main contains the specified IPv4 routes] *****
task path: /WORKDIR/dist-git-network-route_dev_initscripts-tpmdz930/tests/playbooks/tests_route_next_hop_ethernet.yml:88
Wednesday 15 February 2023  15:13:55 +0000 (0:00:05.337)       0:04:18.470 **** 
fatal: [sut]: FAILED! => {
    "assertion": "route_table_main_ipv4.stdout is search(\"198.51.10.64/26 via 198.51.100.6 dev ethtest0 metric 4\")",
    "changed": false,
    "evaluated_to": false
}
```
the problem is that the route output has an extra `proto static` in it:
```
TASK [Get the IPv4 routes from the route table main] ***************************
task path: /WORKDIR/dist-git-network-route_dev_initscripts-tpmdz930/tests/playbooks/tests_route_next_hop_ethernet.yml:76
Wednesday 15 February 2023  15:13:41 +0000 (0:00:19.466)       0:04:04.995 **** 
ok: [sut] => {
    "changed": false,
    "cmd": [
        "ip",
        "-4",
        "route"
    ],
    "delta": "0:00:00.003912",
    "end": "2023-02-15 15:13:43.607055",
    "rc": 0,
    "start": "2023-02-15 15:13:43.603143"
}

STDOUT:

default via 10.31.12.1 dev eth0 proto dhcp metric 100 
10.31.12.0/22 dev eth0 proto kernel scope link src 10.31.14.13 metric 100 
198.51.10.64/26 via 198.51.100.6 dev ethtest0 proto static metric 4 
198.51.12.128/26 via 198.51.100.1 dev ethtest1 proto static metric 2 
198.51.100.0/24 dev ethtest0 proto kernel scope link src 198.51.100.3 metric 101 
198.51.100.0/24 dev ethtest1 proto kernel scope link src 198.51.100.6 metric 102 
```
is it ok if the `ip route` output has the `proto static` in it?  If so, then I suggest changing the test like this:
```yaml
        - name: Assert that the route table main contains the specified IPv4
            routes
          assert:
            that:
              - route_table_main_ipv4.stdout is search("198.51.10.64/26 via
                198.51.100.6 dev ethtest0 .* metric 4")
              - route_table_main_ipv4.stdout is search("198.51.12.128/26 via
                198.51.100.1 dev ethtest1 .* metric 2")
            msg: "the route table main does not contain the specified IPv4
              route"
```

richm:
I get the same error in local qemu/kvm testing - are you able to test locally?

liangwen12year:
> I get the same error in local qemu/kvm testing - are you able to test locally?

I updated the PR,  there are couple issues causing the CI failed:
            1. When running with nm provider, the route will be configured
                with `proto static`
             2.  In RHEL-6.10 managed host, the attribute in the route is not
                  equally spaced

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,148 @@
...
+        - name: Assert that the route table main contains the specified IPv4
+            routes
+          assert:
+            that:
+              # When running with nm provider, the route will be configured
+              # with `proto static`
+              # In RHEL-6.10 managed host, the attribute in the route is not
+              # equally spaced
+              - route_table_main_ipv4.stdout is search("198.51.10.64/26 via
+                198.51.100.6 dev ethtest0 (proto static)?(\s+)metric 4")
```
richm, line 97:
```suggestion
                198.51.100.6 dev ethtest0 (proto static )?metric 4")
```
if the output is `198.51.100.6 dev ethtest0 metric 4` then I don't think `search("198.51.10.64/26 via 198.51.100.6 dev ethtest0 (proto static)?(\s+)metric 4")` will match - it will look for "ethtest0" followed by a space followed by one or more spaces which will fail.  If you are concerned that in output there will be more than one space delimiter then you will need to adjust all of your search regexes to use `\s+` instead of ` `.  Otherwise, the suggested search pattern should suffice.

richm:
> > I get the same error in local qemu/kvm testing - are you able to test locally?
> 
> I updated the PR, there are couple issues causing the CI failed: 1. When running with nm provider, the route will be configured with `proto static` 2. In RHEL-6.10 managed host, the attribute in the route is not equally spaced

ok - then perhaps `search("198.51.10.64/26 via 198.51.100.6 dev ethtest0\s+(proto static )?metric 4")` ?
that should match the following
* `198.51.10.64/26 via 198.51.100.6 dev ethtest0 proto static metric 4`
* `198.51.10.64/26 via 198.51.100.6 dev ethtest0 metric 4`
* `198.51.10.64/26 via 198.51.100.6 dev ethtest0          metric 4`
* `198.51.10.64/26 via 198.51.100.6 dev ethtest0        proto static metric 4`

liangwen12year:
> > > I get the same error in local qemu/kvm testing - are you able to test locally?
> > 
> > 
> > I updated the PR, there are couple issues causing the CI failed: 1. When running with nm provider, the route will be configured with `proto static` 2. In RHEL-6.10 managed host, the attribute in the route is not equally spaced
> 
> ok - then perhaps `search("198.51.10.64/26 via 198.51.100.6 dev ethtest0\s+(proto static )?metric 4")` ? that should match the following
> 
> * `198.51.10.64/26 via 198.51.100.6 dev ethtest0 proto static metric 4`
> * `198.51.10.64/26 via 198.51.100.6 dev ethtest0 metric 4`
> * `198.51.10.64/26 via 198.51.100.6 dev ethtest0          metric 4`
> * `198.51.10.64/26 via 198.51.100.6 dev ethtest0        proto static metric 4`

In the CI log,  https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-573-5f5e070_RHEL-6.10-updates-20201110.17_20230215-161536/artifacts/tests_route_next_hop_ethernet_initscripts-FAILED.log,  it shows the following pattern:

```
198.51.12.128/26 via 198.51.100.1 dev ethtest1  metric 2 
198.51.10.64/26 via 198.51.100.6 dev ethtest0  metric 4 
198.51.100.0/24 dev ethtest0  proto kernel  scope link  src 198.51.100.3 
198.51.100.0/24 dev ethtest1  proto kernel  scope link  src 198.51.100.6 
10.31.12.0/22 dev eth0  proto kernel  scope link  src 10.31.12.78 
169.254.0.0/16 dev eth0  scope link  metric 1002 
169.254.0.0/16 dev ethtest0  scope link  metric 1004 
169.254.0.0/16 dev ethtest1  scope link  metric 1006 
default via 10.31.12.1 dev eth0 
```

richm:
> > > > I get the same error in local qemu/kvm testing - are you able to test locally?
> > > 
> > > 
> > > I updated the PR, there are couple issues causing the CI failed: 1. When running with nm provider, the route will be configured with `proto static` 2. In RHEL-6.10 managed host, the attribute in the route is not equally spaced
> > 
> > 
> > ok - then perhaps `search("198.51.10.64/26 via 198.51.100.6 dev ethtest0\s+(proto static )?metric 4")` ? that should match the following
> > 
> > * `198.51.10.64/26 via 198.51.100.6 dev ethtest0 proto static metric 4`
> > * `198.51.10.64/26 via 198.51.100.6 dev ethtest0 metric 4`
> > * `198.51.10.64/26 via 198.51.100.6 dev ethtest0          metric 4`
> > * `198.51.10.64/26 via 198.51.100.6 dev ethtest0        proto static metric 4`
> 
> In the CI log, https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-573-5f5e070_RHEL-6.10-updates-20201110.17_20230215-161536/artifacts/tests_route_next_hop_ethernet_initscripts-FAILED.log, it shows the following pattern:
> 
> ```
> 198.51.12.128/26 via 198.51.100.1 dev ethtest1  metric 2 
> 198.51.10.64/26 via 198.51.100.6 dev ethtest0  metric 4 
> 198.51.100.0/24 dev ethtest0  proto kernel  scope link  src 198.51.100.3 
> 198.51.100.0/24 dev ethtest1  proto kernel  scope link  src 198.51.100.6 
> 10.31.12.0/22 dev eth0  proto kernel  scope link  src 10.31.12.78 
> 169.254.0.0/16 dev eth0  scope link  metric 1002 
> 169.254.0.0/16 dev ethtest0  scope link  metric 1004 
> 169.254.0.0/16 dev ethtest1  scope link  metric 1006 
> default via 10.31.12.1 dev eth0 
> ```

ok - then the pattern I suggested `search("198.51.10.64/26 via 198.51.100.6 dev ethtest0\s+(proto static )?metric 4")` should work - it should match both `198.51.12.128/26 via 198.51.100.1 dev ethtest1  metric 2` and `198.51.12.128/26 via 198.51.100.1 dev ethtest1 proto static metric 2`

richm:
[citest]

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,191 @@
...
+                  dhcp4: false
+                  address:
+                    - 198.51.100.4/24
+                    - 2001:db8::2/32
+                  route:
+                    - network: 198.58.10.64
+                      prefix: 26
+                      gateway: 198.51.100.102
+                      metric: 4
+          ignore_errors: true  # noqa ignore-errors
```
liangwen12year, line 137: It seems like that the udev rule needs to be changed in order to bring up the connection with only the MAC specified.  I will keep investigate later.

```
# ifup /etc/sysconfig/network-scripts/ifcfg-ethtest1
WARN      : [ifup] You are using 'ifup' script provided by 'network-scripts', which are now deprecated.
WARN      : [ifup] 'network-scripts' will be removed in one of the next major releases of RHEL.
WARN      : [ifup] It is advised to switch to 'NetworkManager' instead - it provides 'ifup/ifdown' scripts as well.
ERROR     : [/etc/sysconfig/network-scripts/ifup-eth] Device ethtest1@peerethtest1 does not seem to be present, delaying initialization.
```

liangwen12year:
[citest]

library/network_connections.py
thom311: this is not correct.

It's not about "whether ansible [...] set `interface_name`. The `connection` dictionary is the result of `ArgValidator.validate()` which validates and normalizes the dictionary. The major point of normalization is that the code later can rely on `connection` containing certain values and not need to do another round of error checking.

Also, the code above already accesses `connection["interface_name"]`

tests/playbooks/tests_route_next_hop_ethernet.yml
ffmancera: While this test could be a nice to have I think it is not needed. Checking that the device is being specified in the route when `interface_name` is set for `initscripts` provider should be enough if this is a problem. @tyll what do you think?

ffmancera (commented):
I would rephrase the commit message.. something like:

```
initscripts: add missing output interface when creating a route

Initscripts provider is not configuring the output interface, `dev`, when creating a route.
If the user configures routes for 2 different interfaces in the same subnet kernel could set
the incorrect output interface and therefore the traffic would be routed through the wrong
interface.

Now initscripts configures the output interface according to the `interface_name` set. 
If the `interface_name` property is missing then kernel will guess the output interface. 
This behaviour is trying to honor the NetworkManager provider workflow which always set the output interface implicitly.
```

Please, this is a suggestion, feel free to ignore if you think the other one is clearer :-)

Thank you!

tests/playbooks/tests_route_next_hop_ethernet.yml
tyll: IIRC there are two acceptance criteria (it works with interface name, there is a warning if it is missing), they should both be tested. FYI: The most modern test format for the role is in https://github.com/linux-system-roles/network/blob/main/tests/playbooks/tests_states.yml - unfortunately, it is using "user stories" instead of "acceptance criteria" to describe the tests because I learned about acceptance criteria later.

lsr_setup corresponds to the given part, lsr_test is the when part and lsr_assert_* the then part of an acceptance criterion.

tyll: Also, I am not 100% sure what you are asking, I don't have the time to read the full test. Adjusting udev seems to be wrong, if veths are a problem, maybe try dummy interfaces.

ffmancera: I think udev shouldn't be adjusted. I think use dummy interfaces is a great idea. Let's do that. Thank you!

liangwen12year: In the current role, the dummy interface is not supported by the initscripts provider, and the mac address is only allowed for infiniband or ethernet

liangwen12year: Since we are testing the warning message, we can just omit the `state: up`, and the warning message will also be logged.

liangwen12year:
> 

I reworded a bit. Thanks for proposal.

liangwen12year:
> I would rephrase the commit message.. something like:
> 
> ```
> initscripts: add missing output interface when creating a route
> 
> Initscripts provider is not configuring the output interface, `dev`, when creating a route.
> If the user configures routes for 2 different interfaces in the same subnet kernel could set
> the incorrect output interface and therefore the traffic would be routed through the wrong
> interface.
> 
> Now initscripts configures the output interface according to the `interface_name` set. 
> If the `interface_name` property is missing then kernel will guess the output interface. 
> This behaviour is trying to honor the NetworkManager provider workflow which always set the output interface implicitly.
> ```
> 
> Please, this is a suggestion, feel free to ignore if you think the other one is clearer :-)
> 
> Thank you!

I reworded a bit. Thanks for proposal.

library/network_connections.py
liangwen12year: Thanks for the suggestion.

liangwen12year:
[citest]

liangwen12year:
[citest]

richm:
If we want to get this into the release, we need this reviewed and approved ASAP.  @tyll @thom311 @ffmancera @cathay4t if there are no further objections, I would like to merge this PR.

thom311 (approved):

tyll:
Regarding the commit message:
```
initscripts: Add missing output interface when creating a route
```
"missing" is somewhat implied, otherwise we don't need a patch. Strictly speaking, it is not missing since it is not always necessary. 

```Initscripts provider is not configuring the output interface when
creating a route.
```
Regarding "when creating":  this is seems to correspond to when actually configuring the routes in kernel but this is about the time when writing the config files 

```
 If the user configures routes for 2 different
interfaces in the same subnet, kernel could set the incorrect output
interface and therefore the traffic would be routed through the wrong
interface.
```

The unintended behavior only needs the user to configure one route, this is not about configuring multiple routes.


```
Now initscripts provider configures the output interface according to
the `interface_name` setting. If the `interface_name` setting is
missing then the kernel will guess the output interface.
```

The kernel does not know about `interface_name`.

```
 This behaviour
is trying to honor the behavior of NetworkManager provider which always
sets the output interface implicitly when creating a route.
```

This is not about trying, this is about doing. Also, NetworKmanager (implicitly) sets (explicitly) the output interface. NM's behavior is implicitly configuring the output interface and the output interface is explicitly configured in the kernel. If the kernel is "guessing", this would be an implicit output interface.

my proposal:

```
initscripts: Configure output device in routes

Without an explicit output device, the kernel might use a different output device than intended by the user. Therefore, use the interface name of connections to specify it if it is available. Otherwise, educate the user about this potential problem with a warning. This aligns the behavior with NetworkManager which configures the output device in routes when activating profiles on devices.

Fixes: https://bugzilla.redhat.com/2168735
```

liangwen12year:
> Regarding the commit message:
> 
> ```
> initscripts: Add missing output interface when creating a route
> ```
> 
> "missing" is somewhat implied, otherwise we don't need a patch. Strictly speaking, it is not missing since it is not always necessary.
> 
> ```
> creating a route.
> ```
> 
> Regarding "when creating": this is seems to correspond to when actually configuring the routes in kernel but this is about the time when writing the config files
> 
> ```
>  If the user configures routes for 2 different
> interfaces in the same subnet, kernel could set the incorrect output
> interface and therefore the traffic would be routed through the wrong
> interface.
> ```
> 
> The unintended behavior only needs the user to configure one route, this is not about configuring multiple routes.
> 
> ```
> Now initscripts provider configures the output interface according to
> the `interface_name` setting. If the `interface_name` setting is
> missing then the kernel will guess the output interface.
> ```
> 
> The kernel does not know about `interface_name`.
> 
> ```
>  This behaviour
> is trying to honor the behavior of NetworkManager provider which always
> sets the output interface implicitly when creating a route.
> ```
> 
> This is not about trying, this is about doing. Also, NetworKmanager (implicitly) sets (explicitly) the output interface. NM's behavior is implicitly configuring the output interface and the output interface is explicitly configured in the kernel. If the kernel is "guessing", this would be an implicit output interface.
> 
> my proposal:
> 
> ```
> initscripts: Configure output device in routes
> 
> Without an explicit output device, the kernel might use a different output device than intended by the user. Therefore, use the interface name of connections to specify it if it is available. Otherwise, educate the user about this potential problem with a warning. This aligns the behavior with NetworkManager which configures the output device in routes when activating profiles on devices.
> 
> Fixes: https://bugzilla.redhat.com/2168735
> ```

Thanks for explanation, I updated.

liangwen12year:
[citest]

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Test configuring ethernet devices
```
tyll, line 3: The name is wrong, maybe a result from copying another test? Also, the name of the test file "next hop" - this term is not used in the code, warning or commit message. Seems more appropriate to call it `tests_route_device.yml`. Are there plans to test other devices than ethernet? If not, there is no need for the ethernet suffix.

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+        interface: "{{ interface1 }}"
+    - name: Show interfaces
+      include_tasks: tasks/show_interfaces.yml
+    - name: Manage test interface
+      include_tasks: tasks/manage_test_interface.yml
+      vars:
+        state: present
+    - name: Assert device is present
+      include_tasks: tasks/assert_device_present.yml
+    - name: Run the test
```
tyll, line 35: This should be a descriptive name of the test scenario, there does not seem to be one that explains the acceptance criterion.

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+      block:
+        - name: Configure the IP addresses
```
tyll, line 37: This does not only configure IP addresses but also a route.

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+          import_role:
+            name: linux-system-roles.network
+          vars:
+            network_connections:
+              - name: "{{ interface0 }}"
+                interface_name: "{{ interface0 }}"
+                state: up
+                type: ethernet
+                autoconnect: true
```
tyll, line 46: Why autoconnect:true?

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+          ignore_errors: true
+          changed_when: false
+
+        - name: Get the IPv6 routes from the route table main
+          command: ip -6 route
+          register: route_table_main_ipv6
+          ignore_errors: true
+          changed_when: false
+
+        - name: Assert that the route table main contains the specified IPv4
```
tyll, line 88: IMHO it is cleaner to keep getting and assertions together, so first get IPv4 then assert IPv4, then the same for IPv6. But this is a nitpick.

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+                  dhcp4: false
+                  address:
+                    - 198.51.100.4/24
+                    - 2001:db8::6/32
+                  route:
+                    - network: 198.58.10.64
+                      prefix: 26
+                      gateway: 198.51.100.102
+                      metric: 4
+          when: network_provider == "initscripts"
```
tyll, line 136: IMHO only the assertion should depend on the provider, this is also a perfect test case for NM, just the expectation/assertion is different.

tests/playbooks/tests_route_next_hop_ethernet.yml
```diff
@@ -0,0 +1,188 @@
...
+                    state: down
+              ignore_errors: true
+            - name: Assert interface0 profile is absent
+              include_tasks: tasks/assert_profile_absent.yml
+              vars:
+                profile: "{{ interface0 }}"
+            - name: Assert interface1 profile is absent
+              include_tasks: tasks/assert_profile_absent.yml
+              vars:
+                profile: "{{ interface1 }}"
```
tyll, line 173: Seems that this could be a loop.

tyll (changes requested):
At least the descriptions in the tests need to be accurate.

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: We can omit this setting, since the default value is True for `autoconnect`, but it is true that we may not want this autoconnect behavior to happen during any of the restart scenario. (e.g. restart NM, systemctl reboot). I will change into `autoconnect: false`.

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: > The name is wrong, maybe a result from copying another test? 

Yes, honestly, I should be more smart about writing descriptive name. 

> Are there plans to test other devices than ethernet? If not, there is no need for the ethernet suffix.

Currently, there is no plan.

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: I changed into `-name: Test the route or the warning log when configuring the route with
or without the interface name`

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: I updated.

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: The condition was removed and I also added the assertion for the nm provider scenario.

tests/playbooks/tests_route_next_hop_ethernet.yml
liangwen12year: I used the loop to simplify the testing code.

liangwen12year:
@tyll , thanks for taking time to review it. I addressed all you previous comments and updated respectively.

liangwen12year:
[citest]

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,202 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Test configuring the output device of the route
```
tyll, line 3:
```suggestion
- name: Test configuring the output device of routes
```

The article `the` is not correct here, since this is not about one specific route.

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,202 @@
...
+      include_tasks: tasks/show_interfaces.yml
+    - name: Manage test interface
+      include_tasks: tasks/manage_test_interface.yml
+      vars:
+        state: present
+    - name: Assert device is present
+      include_tasks: tasks/assert_device_present.yml
+    - name: Set type and interface1
+      set_fact:
+        type: "{{ type }}"
```
tyll, line 25: it is not necessary to set the type again
```suggestion
```

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,202 @@
...
+                    - 2001:db8::4/32
+                  route:
+                    - network: 198.51.12.128
+                      prefix: 26
+                      gateway: 198.51.100.1
+                      metric: 2
+        - name: Get the IPv4 routes from the route table main
+          command: ip -4 route
+          register: route_table_main_ipv4
+          ignore_errors: true  # noqa ignore-errors
```
tyll, line 81: which errors need to be ignored?

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,195 @@
...
+        - name: Assert that the warning about specifying the route without
+            the output device is not logged for nm provider
+          assert:
+            that:
+              - __network_connections_result.stderr is not search("The
+                connection .* does not specify an interface name. Therefore,
+                the route .* will be configured without the output device and
+                the kernel will choose it automatically which might result in
+                an unwanted device being used. To avoid this, specify
+                `interface_name` in the connection appropriately.")
```
tyll, line 156: can you assert that there is no warning in this case? Are there any warnings usually for this configuration?

ffmancera (approved):
IMO, this looks good to me. I just made a comment about the description of an assert but I would not like to block on that.

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,195 @@
...
+            msg: The warning about specifying the route without the output
+              device is logged for nm provider
+          when: network_provider == "nm"
```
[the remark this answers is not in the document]
liangwen12year: Sure, all the warning logs are tagged by `<warn>`.

tests/playbooks/tests_route_device.yml
liangwen12year: We can assert that there is no warning in this case. There is no warning for this configuration.

liangwen12year:
[citest]

liangwen12year:
@tyll @ffmancera ,Thanks for reviewing, I addressed all the comments.

liangwen12year:
@tyll , @ffmancera , @thom311 , @cathay4t ,if no other comments. Very likely,  @richm will get it merged today.

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,188 @@
...
+          assert:
+            that:
+              - route_table_main_ipv6.stdout is search("2001:db6::4 via
+                2001:db8::1 dev ethtest0\s+(proto static )?metric 2")
+            msg: "the route table main does not contain the specified IPv6
+              route"
+        - name: Get the interface1 MAC address
+          command: cat /sys/class/net/{{ interface1 }}/address
+          register: interface1_mac
+          ignore_errors: true  # noqa ignore-errors
```
tyll, line 111: What's the reason to have ignore_errors:true here?

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,188 @@
...
+                device being used. To avoid this, specify `interface_name` in
+                the connection appropriately.")
+            msg: The warning about specifying the route without the output
+              device is not logged for initscripts provider
+          when: network_provider == "initscripts"
+
+        - name: Assert that no warning is logged for nm provider
+          assert:
+            that:
+              - __network_connections_result.stderr is not search("<warn>")
```
tyll, line 150: :+1:

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,188 @@
...
+        - name: Assert that the warning about specifying the route without
+            the output device is logged for initscripts provider
+          assert:
+            that:
+              - __network_connections_result.stderr is search("The connection
+                .* does not specify an interface name. Therefore, the route .*
+                will be configured without the output device and the kernel
+                will choose it automatically which might result in an unwanted
+                device being used. To avoid this, specify `interface_name` in
+                the connection appropriately.")
```
tyll, line 142: The acceptance criteria state that this is a warning, so the assertion should also check that this message is prefixed with `<warn>`, I guess. Not sure, what the exact syntax is.

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,188 @@
...
+                    persistent_state: absent
+                    state: down
+              ignore_errors: true  # noqa ignore-errors
+            - name: Assert interface0 profile and interface1 profile are absent
+              include_tasks: tasks/assert_profile_absent.yml
+              vars:
+                profile: "{{ item }}"
+              loop:
+                - "{{ interface0 }}"
+                - "{{ interface1 }}"
```
tyll, line 174: When using ignore_errors above (which seems to be correct), it does not seem to be necessary to assert that the profiles are gone. And if, then maybe only after trying to delete everything to ensure that as much as possible is cleaned up.

tests/playbooks/tests_route_device.yml
liangwen12year: I was thinking that the interface1 is possible not created and configured properly. So the  directory `/sys/class/net/{{ interface1 }}/` may not exist, but after the second thought, drop `ignore_errors: true  # noqa ignore-errors`  is better, so that error will be raised when  the  directory `/sys/class/net/{{ interface1 }}/` is not existed.

tests/playbooks/tests_route_device.yml
liangwen12year: I honestly think that we should clean up all the tests interfaces and profiles at the end of the tests so that we know that the current tests will not affect the next running tests (imagine that we are running the test playbooks in sequences). Therefore, from my point of view, it is necessary to assert the profiles are absent.

tests/playbooks/tests_route_device.yml
liangwen12year: I prefixed with <warn> in the assertion.

liangwen12year:
[citest]

liangwen12year:
@tyll I addressed all the comments. Thanks.

liangwen12year:
[citest]

liangwen12year:
If no other objection, @richm may consider merge it today.

richm:
> If no other objection, @richm may consider merge it today.

Both @tyll and @cathay4t have requested changes that haven't been resolved yet.  I would like them to clear their change requests first.

tests/playbooks/tests_route_device.yml
tyll: In that case, it does not seem to make sense to ignore the errors in line 166. Also, your response does not address the concern to first delete the interfaces before asserting that the profiles are gone.

library/network_connections.py
```diff
@@ -531,6 +531,17 @@ def ifcfg_create(
                 line = r["network"] + "/" + str(r["prefix"])
                 if r["gateway"]:
                     line += " via " + r["gateway"]
+                if connection["interface_name"]:
+                    line += " dev " + connection["interface_name"]
+                else:
+                    warn_fcn(
+                        "The connection '{0}' does not specify an interface name. "
```
tyll, line 538: There seems to be a unit test missing for testing the route configuration without an interface_name.

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,187 @@
...
+                      prefix: 26
+                      gateway: 198.51.100.102
+                      metric: 4
+
+        - name: Assert that the warning about specifying the route without
+            the output device is logged for initscripts provider
+          assert:
+            that:
+              - __network_connections_result.stderr is search("<warn> .* The
+                connection .* does not specify an interface name. Therefore,
```
tyll, line 137: The name of the connection is known, what's the reason to use a broad wildcard here?

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,187 @@
...
+                    - network: 198.58.10.64
+                      prefix: 26
+                      gateway: 198.51.100.102
+                      metric: 4
+
+        - name: Assert that the warning about specifying the route without
+            the output device is logged for initscripts provider
+          assert:
+            that:
+              - __network_connections_result.stderr is search("<warn> .* The
```
tyll, line 136: What is the wildcard matching here?

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,187 @@
...
+                connection .* does not specify an interface name. Therefore,
+                the route .* will be configured without the output device and
```
tyll, line 138: This broad wildcard is hiding that warning is not properly formatting the information about the route.

library/network_connections.py
```diff
@@ -531,6 +531,17 @@ def ifcfg_create(
...
+                        "Therefore, the route '{1}' will be configured without the "
```
tyll, line 539: The result of `{1}` is very hard to read. How about
```"...the route to {1}/{2}...".format(name, r["network"], r["prefix"])```

Comment (tyll) [reply library/network_connections.py:539]: @ffmancera FYI

Comment (ffmancera) [reply library/network_connections.py:539]: Nice catch. I agree.

File: library/network_connections.py (hunk 45 above)
Comment (tyll) [reply library/network_connections.py:538]: https://github.com/linux-system-roles/network/commit/ac7cc6fe94cc4fd856a323b6d0fe26e3309c2c6b shows how to check warnings for initscripts. You will have to rebase on that commit before adding a new test.

File: library/network_connections.py (hunk 49 above)
Comment (tyll) [reply library/network_connections.py:539]: The unit tests caught it. :grin:

Comment (tyll) [reply library/network_connections.py:539]: This also shows the importance of a demo as part of the review.

Comment (ffmancera) [reply library/network_connections.py:539]: Yes, you are right here. +1

File: tests/playbooks/tests_route_device.yml (hunk 46 above)
Comment (liangwen12year) [reply tests/playbooks/tests_route_device.yml:137]: nice catch. yes, we can avoid the wildcard.

File: tests/playbooks/tests_route_device.yml (hunk 47 above)
Comment (liangwen12year) [reply tests/playbooks/tests_route_device.yml:136]: The warning is:

```
stderr": "[003] <warn>  #0, state:None persistent_state:present, 'ethtest1': The connection 'ethtest1' does not specify an interface name. Therefore, the route '{'network': '198.58.10.64', 'prefix': 26, 'gateway': '198.51.100.102', 'metric': 4, 'table': None, 'family': <AddressFamily.AF_INET: 2>}' will be configured without the output device and the kernel will choose it automatically which might result in an unwanted device being used. To avoid this, specify `interface_name` in the connection appropriately.\n[004] <info>  #0, state:None persistent_state:present, 'ethtest1': update ifcfg-rh profile 'ethtest1'\n",
```

File: library/network_connections.py (hunk 49 above)
Comment (liangwen12year) [reply library/network_connections.py:539]: Yes, the proposed change is nice.

File: tests/playbooks/tests_route_device.yml (hunk 47 above)
Comment (tyll) [reply tests/playbooks/tests_route_device.yml:136]: You can mask problematic single characters with a dot, so for example:

`.003. <warn>  .0, state:None persistent_state:present, 'ethtest1':` - Assuming that the brackets or the hash sign caused the issues.

**Comment by liangwen12year** [comment]
[citest]

File: tests/playbooks/tests_route_device.yml (hunk 47 above)
Comment (richm) [reply tests/playbooks/tests_route_device.yml:136]: Another "trick" is to define pattern strings as `vars` e.g.
```yaml
          assert:
            that:
              - __network_connections_result.stderr is search(__search_pattern)
          vars:
            __search_pattern: string that can have problematic meta characters
```

liangwen12year: The problematic characters are hash sign and colon.

richm: > The problematic characters are hash sign and colon. I updated. Thanks.

https://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html
"When not to use Plain Scalars"

liangwen12year:
[citest]

tests/unit/test_network_connections.py
```diff
@@ -317,13 +317,19 @@ def do_connections_validate_ifcfg(self, input_connections, **kwargs):
...
             )
             # pprint("con[%s] = \"%s\"" % (idx, connections[idx]['name']), c)
-            exp = kwargs.get("initscripts_dict_expected", None)
-            if exp is not None:
-                self.assertEqual(exp[idx], c)
+            expected_config = kwargs.get("initscripts_dict_expected", None)
+            if expected_config is not None:
+                self.assertEqual(expected_config[idx], config)
+            expected_warnings = kwargs.get("initscripts_expected_warnings", [])
+            self.assertEqual(expected_warnings, warnings)
```
liangwen12year, line 332: This commit is taken from https://github.com/linux-system-roles/network/commit/ac7cc6fe94cc4fd856a323b6d0fe26e3309c2c6b which is submitted by @tyll . I honestly do not know how to  change the author into @tyll .  @tyll , please change the author's name if you'd like to.

tyll: I don't know how you applied the patch. If you use git to do it, it keeps the author. If you added my repo and fetched from it (which is my recommendation), you should be able to run `git cherry-pick  f6d6a79a01b8004527e15ef84715571aaa85f78a` to get the commit. If you do not want to do this, you can download and apply the patch:

```
wget https://github.com/linux-system-roles/network/commit/ac7cc6fe94cc4fd856a323b6d0fe26e3309c2c6b.patch
git am ac7cc6fe94cc4fd856a323b6d0fe26e3309c2c6b.patch
```

To change the author of the latest commit (you might need to rebase interactively to edit the commit if it is not the latest), use

```
git commit --amend --author "Till Maas ..."
```

Regarding the commit message, you also removed my Signed-off-by line. And about
"The unit test is missing warnings check when testing with initscritps provider, add support for that."  

- The summary already states "Unit tests/initscripts: Support checking warnings", this text does not seem to add any additional information, does it?
- "The unit test" is wrong, since the patch is about the TestValidator class, not one specific unit test. The unit test framework could be considered correct.
- "is missing warnings check" - checks would be part of a specific test, support for checks (which is added here) part of the framework. So "is missing support to check for warnings" might be correct
	- If this is about adjusting an existing unit test, it would be ok to write "The unit test for route devices is missing a check for warnings" but this is not the case here.
- If you want to add additional context to the commit message, I guess this would be the fact that the unit test to be added will need this to be written, so something like "This will be needed to test that warnings are generated when specifying routes in connections without an interface name." I am not convinced, that this is adding much value to the commit message, though.

tests/unit/test_network_connections.py
```diff
@@ -2786,13 +2792,176 @@
...
                     "rule": None,
                     "rule6": None,
                 }
             ],
+            initscripts_expected_warnings=[
+                "The connection e556 does not specify an interface name. Therefore, the route "
+                "to 192.168.45.0/24 will be configured without the output device and the "
+                "kernel will choose it automatically which might result in an unwanted device "
+                "being used. To avoid this, specify `interface_name` in the connection "
+                "appropriately.",
```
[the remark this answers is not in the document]
tyll: It seems worth to create a template string for the warning and then format it accordingly for each different route to reduce the duplication. It will also solve the CodeQL warning.

tests/unit/test_network_connections.py
```diff
@@ -317,13 +317,19 @@ def do_connections_validate_ifcfg(self, input_connections, **kwargs):
...
             if content_current:
                 content_current = content_current[idx]
-            c = IfcfgUtil.ifcfg_create(
-                connections, idx, content_current=content_current
+            warnings = []
+            config = IfcfgUtil.ifcfg_create(
+                connections,
+                idx,
+                content_current=content_current,
+                warn_fcn=lambda msg: warnings.append(msg),
```
tyll, line 325: Nice finding from pylint:

```suggestion
                warn_fcn=warnings.append,
```

tests/unit/test_network_connections.py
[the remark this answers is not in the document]
liangwen12year: I created the template string to reduce the duplication.

tests/unit/test_network_connections.py
liangwen12year: I just knew that adding the suffix `.patch` to the commit will get the raw patch. Thanks for sharing. I just felt weird for me to change the author without the permission from the author. For example, I created a commit which had all the code changes from another author, then I felt uncomfortable to sign off on behalf of another author and left the author's name as another author (without the explicit permission from another author).

Indeed, the commit message that I added did not introduce any new information.

library/network_connections.py
liangwen12year: The unit test was added. Thanks.

tests/unit/test_network_connections.py
```diff
@@ -2792,13 +2792,171 @@ def test_route_v6(self):
...
+            ],
+        )
+
+    def test_route_v6_without_interface_name(self):
+        route_device_warning = (
+            "The connection e556 does not specify an interface name. Therefore, the "
+            "route to {} will be configured without the output device and the kernel "
+            "will choose it automatically which might result in an unwanted device "
+            "being used. To avoid this, specify `interface_name` in the connection "
+            "appropriately."
```
ffmancera, line 2812: Python2.6 will require you to use indices in the format specification. Please, replace `{}` with `{0}` and that should solve the unit test failing.

tests/unit/test_network_connections.py
```diff
@@ -2792,13 +2792,171 @@ def test_route_v6(self):
...
+                    "192.168.45.0/24 dev e556 metric 545\n"
+                    "192.168.46.0/30 dev e556\n",
+                    "route6": "a:b:c:f::/64\na:b:c:d::/64 dev e556\n",
+                    "rule": None,
+                    "rule6": None,
+                }
+            ],
+        )
+
+    def test_route_v6_without_interface_name(self):
```
ffmancera, line 2806: The test name mention v6 but AFAICS, this is checking both IPv4 and IPv6.. isn't it?

tests/playbooks/tests_route_device.yml
liangwen12year: I dropped the ignore_errors in line 166. I also moved the assertion (about the profile absent) after deleting the interface.

tests/unit/test_network_connections.py
liangwen12year: Yes, I changed the test name into `test_route_without_interface_name()`.

liangwen12year:
@tyll @ffmancera , I addressed all the comments, if you agree with the changes, please resolve the comments/conversation.

liangwen12year:
[citest]

ffmancera (approved):
It looks good to me! Thank you for all the work @liangwen12year !

richm:
[citest]

tests/playbooks/tests_route_device.yml
```diff
@@ -0,0 +1,188 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Test configuring the output device of routes
```
tyll, line 3:
```suggestion
- name: Test output device of routes
```

richm:
[citest]

tyll (approved):
Thank you, LGTM!

tyll:
@richm can you see the unresolved comment, maybe from a notification e-mail and resolve it? Github does not show it to me anymore and I am not sure how to resolve it. It seems like a bug.

tyll:
> @richm can you see the unresolved comment, maybe from a notification e-mail and resolve it? Github does not show it to me anymore and I am not sure how to resolve it. It seems like a bug.

Found it in the hidden lines.

richm:
baseos ci is down - tests are getting cancelled due to error - that's why the statuses are all pending 

The github action integration test for centos 7 is failing in `tests_bond.yml`: https://github.com/linux-system-roles/network/actions/runs/4217581686/jobs/7322395842
```
TASK [Create test interfaces] **************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "set -euo pipefail\nip link add test1 type veth peer name test1p
...
"stderr": "RTNETLINK answers: File exists", "stderr_lines": ["RTNETLINK answers: File exists"], "stdout": "", "stdout_lines": []}
```

liangwen12year:
> baseos ci is down - tests are getting cancelled due to error - that's why the statuses are all pending
> 
> The github action integration test for centos 7 is failing in `tests_bond.yml`: https://github.com/linux-system-roles/network/actions/runs/4217581686/jobs/7322395842
> 
> ```
> TASK [Create test interfaces] **************************************************
> fatal: [localhost]: FAILED! => {"changed": false, "cmd": "set -euo pipefail\nip link add test1 type veth peer name test1p
> ...
> "stderr": "RTNETLINK answers: File exists", "stderr_lines": ["RTNETLINK answers: File exists"], "stdout": "", "stdout_lines": []}
> ```



> baseos ci is down - tests are getting cancelled due to error - that's why the statuses are all pending
> 
> The github action integration test for centos 7 is failing in `tests_bond.yml`: https://github.com/linux-system-roles/network/actions/runs/4217581686/jobs/7322395842
> 
> ```
> TASK [Create test interfaces] **************************************************
> fatal: [localhost]: FAILED! => {"changed": false, "cmd": "set -euo pipefail\nip link add test1 type veth peer name test1p
> ...
> "stderr": "RTNETLINK answers: File exists", "stderr_lines": ["RTNETLINK answers: File exists"], "stdout": "", "stdout_lines": []}
> ```

The CentOS 7 is failing in `tests/tests_bond_nm.yml`  which is not caused the code updates in this PR.

cathay4t (approved):

liangwen12year:
[citest]

PR #574, opened 2023-02-13
Use custom shellcheck github action

nhosoi:
The upstream repo for the differential shellcheck action:
https://github.com/redhat-plumbers-in-action/differential-shellcheck


thom311 (dismissed):

thom311:
could you please elaborate what the advantages of this shellcheck test are, compare to the previous one (which it replaces)?

also "differential shellcheck", I presume will only complain about newly added violations. Is that really desired, don't we always want to get all warnings -- and either fix them, or suppress them?

thom311:
or maybe leave the other shellcheck test? I takes 14 seconds to run and it's already configured and passing. Does it cause any problem to run them both?

richm:
@jamacku can you answer the questions above from @thom311 and explain the benefits of differential shellcheck over the "plain" shellcheck action?

jamacku:
Hey @richm,

@thom311, you are right. It will complain only about added defects on PR. Once PR is merged, it behaves the same way, but there is [`strict-check-on-push`](https://github.com/redhat-plumbers-in-action/differential-shellcheck#strict-check-on-push) option. When enabled action will complain about any defect in the code base (but it is only available on push event).

Differential ShellCheck also supports SARIF (you will see reports in GitHub UI - on PR and in the Security tab).

You can keep both actions, but I think it's not required, given that you don't have any ShellCheck warnings in the current code base and are not pushing directly to `main`.

_I'll be on PTO next week, so I might be slow to respond._

cathay4t (approved):
LGTM

tyll:
Good catch @thom311. I disabled the auto merge. What is the reason not to enable "strict-check-on-push"? @cathay4t since you approved this and @jamacku is unavailable, what's the benefit?

> You can keep both actions, but I think it's not required, given that you don't have any ShellCheck warnings in the current code base and are not pushing directly to `main`.

ShellCheck might also develop new checks/warnings, I guess, which might lead to new warnings being reported in `main`.

richm:
> Good catch @thom311. I disabled the auto merge. What is the reason not to enable "strict-check-on-push"? @cathay4t since you approved this and @jamacku is unavailable, what's the benefit?
> 
> > You can keep both actions, but I think it's not required, given that you don't have any ShellCheck warnings in the current code base and are not pushing directly to `main`.
> 
> ShellCheck might also develop new checks/warnings, I guess, which might lead to new warnings being reported in `main`.

I'll wait for @jamacku to address these concerns.

richm:
@jamacku Can you help?  I don't know how to answer the questions or address the objections that the network team has to using Differential Shellcheck vs. "plain" shellcheck.

jamacku:
Differential ShellCheck was developed mainly for codebases with many shell scripts and ShellCheck defects that would be tricky/dangerous to mitigate. It allows you to introduce ShellCheck immediately without dealing with all existing defects. So you can quickly check incoming changes and try to improve the state of existing scripts in your free time.

I'm thinking of ShellCheck as a linter suggesting the correct way to write shell scripts. ShellCheck defect doesn't mean a bug in the script (but it certainly can). So when a new release of ShellCheck introduces new kinds of defects, it doesn't mean that your script has a bug that should be addressed.

The current configuration proposed in this PR - ShellCheck check will only block development/fixes if PR/commit introduces new defects. Once changes are pushed to the default (main) branch SARIF report containing all defects will be uploaded to the GitHub repo and can be accessed via UI on [ `Security -> Code Scanning` tab](https://github.com/linux-system-roles/network/security/code-scanning?query=is%3Aopen+branch%3Amain+tool%3Ashellcheck) or by using GitHub API.

So when a new version of ShellCheck is released, defects won't show up due to the nature of differential scanning, but you will see them in the Security tab, and you can decide if it is indeed a defect or not.

Screenshot from 2023-03-08 09-32-57

thom311:
> I'm thinking of ShellCheck as a linter suggesting the correct way to write shell scripts. ShellCheck defect doesn't mean a bug in the script (but it certainly can). So when a new release of ShellCheck introduces new kinds of defects, it doesn't mean that your script has a bug that should be addressed.

That seems to be common with any linter or compiler warnings. New version may introduce new warnings, and most of the time they are false positives. So there is a constant effort required to fix them. But usually, we consider that a worthy effort. Based on the hope, that once in a while it catches an actual bug and a more uniform coding style is achieved.

I guess, the differential check makes a lot of sense for projects that don't want to (or cannot) invest that constant effort. That doesn't seem to apply to us however.

richm:
> > I'm thinking of ShellCheck as a linter suggesting the correct way to write shell scripts. ShellCheck defect doesn't mean a bug in the script (but it certainly can). So when a new release of ShellCheck introduces new kinds of defects, it doesn't mean that your script has a bug that should be addressed.
> 
> That seems to be common with any linter or compiler warnings. New version may introduce new warnings, and most of the time they are false positives. So there is a constant effort required to fix them. But usually, we consider that a worthy effort. Based on the hope, that once in a while it catches an actual bug and a more uniform coding style is achieved.
> 
> I guess, the differential check makes a lot of sense for projects that don't want to (or cannot) invest that constant effort. That doesn't seem to apply to us however.

I'm assuming that means you want to use "plain" shellcheck.  Does the rest of the network team feel the same way?   Note that in the interest of consistency among *all* of the other system roles, I would like to not use "plain" shellcheck in some roles, and differential in other roles, so whatever we decide here will likely determine the direction we take in all of the other system roles.

tyll:
I am also in favor of using plain ShellCheck and in case ShellCheck introduces new checks that we cannot trivially address, pin the version of ShellCheck to the previous/working version and plan to address the new issues as part of a regular development cycle.

richm:
> Differential ShellCheck was developed mainly for codebases with many shell scripts and ShellCheck defects that would be tricky/dangerous to mitigate

I agree with @tyll and @thom311.  For the network role, and all of the system roles, neither of these conditions is true.  We only have 10 roles that have shell scripts, 18 scripts total.  Of these, only 5 are run at runtime, and these are small - the rest are used for testing, building, git commit hooks, etc.

So based on this, and the feedback from the network team, I don't think we will use differential shellcheck in system roles at this time.

nhosoi:
> > Differential ShellCheck was developed mainly for codebases with many shell scripts and ShellCheck defects that would be tricky/dangerous to mitigate
> 
> I agree with @tyll and @thom311. For the network role, and all of the system roles, neither of these conditions is true. We only have 10 roles that have shell scripts, 18 scripts total. Of these, only 5 are run at runtime, and these are small - the rest are used for testing, building, git commit hooks, etc.
> 
> So based on this, and the feedback from the network team, I don't think we will use differential shellcheck in system roles at this time.

We've already merged the differential shellcheck pr's to the rest of the system roles. I'm going to revert them...

PR #576, opened 2023-02-20
Version 1.11.2 - CHANGELOG.md [citest skip]

richm:
[1.11.2] - 2023-02-20
--------------------

### New Features

- none

### Bug Fixes

- initscripts: Configure output device in routes

### Other Changes

- none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,21 @@
 Changelog
 =========
 
+[1.11.2] - 2023-02-20
+--------------------
+
+### New Features
+
+- none
+
+### Bug Fixes
+
+- initscripts: Configure output device in routes
+
+### Other Changes
+
+- none
+
 [1.11.1] - 2023-01-24
 --------------------
 


CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
...
+### Bug Fixes
+
+- initscripts: Configure output device in routes
+
+### Other Changes
+
+- Fix CentOS 7 integration tests
+- Refactor route table ID parsing
+- Unify naming of route table name constants
+- ansible-lint: Fix yaml[truthy] failures
```
tyll, line 20: IMHO these internal/non user-visible changes do not need to be mentioned in the changelog.

richm: They are in the `Other Changes` section - in the past, we have had similar issues in the past reported in this section of the changelog e.g. https://github.com/linux-system-roles/network/blob/main/CHANGELOG.md#1100---2022-11-01

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 Changelog
 =========
 
+[1.11.2] - 2023-02-20
+--------------------
+
+### New Features
+
+- none
+
```
tyll, line 10: Also it would be nicer to remove empty sections instead of stating none.

richm: Then this makes this particular changelog entry completely different and inconsistent with every other changelog entry in the network changelog and every other system roles changelog

CHANGELOG.md
tyll: Yes, this happened when I did not review them, I guess. I vaguely remember that I had similar requests in the past.

The changelog for [1.7.0](https://github.com/linux-system-roles/network/blob/main/CHANGELOG.md#170---2022-02-14) does not list all the commits from [1.6.0-1.7.0](https://github.com/linux-system-roles/network/compare/1.6.0...1.7.0) which has similar non-user facing changes. I am fine with accepting this for once but I would like not to have this used as precedent for future entries. This is also an example from the bug fixes section being removed when it was empty.

tyll (approved):
Approved with objections.

CHANGELOG.md
richm: Unfortunately, we cannot currently have empty sections - so what I will do is make `Other Changes` have contents of `- none` - then later we will need to change the way we generate the collection changelog from the individual role changelogs - see https://github.com/linux-system-roles/auto-maintenance/blob/master/release_collection.py#L267

CHANGELOG.md
tyll: It became inconsistent when using "none" for empty sections were introduced in 1.9.1 - also, the benefit of focusing on the user impact changes becomes visible in [the 1.11.1 changelog(https://github.com/linux-system-roles/network/blob/main/CHANGELOG.md#1111---2023-01-24) which seems to contain only noise - from the changelog, I am not sure why there was a release actually but the fact that there is a release implies that there is something hidden in there, I guess. Maybe it is the ansible-lint updates because they are needed for Galaxy? My point is, that the changelog is supposed to create a benefit for the user and that is limited if there is too much noise.

tyll (approved):

liangwen12year (approved):
LGTM.

CHANGELOG.md
richm: re: "none" sections - The thinking at the time was that if the changelog just contained something like
```
### Other Changes

- something
```
then the user might be confused as to whether or not there were any new features or bug fixes in the release.  I thought it would be better for it to be explicit rather than implicit.
OTOH, you might ask why we are doing a release if there are no new features or bug fixes?  It's a bit tricky because there are several times we have to make a change to the role and a release solely to fix a downstream test issue or Galaxy import issue.  These are not "user visible" to most users, but might be visible if we do have actual users who run tests or use some sort of linter on our roles before they use them in production.

tyll: If the release is solely for Galaxy or downstream the changelog should just state this like

```
### Other Changes
- Fix ansible-lint 6.x warnings to allow Galaxy import
```

or maybe just ``` - fix Ansible Galaxy import```. Btw. the commit message does not need to copy the changelog IMHO, currently it also contains "- none", twice.

Whether it needs to be explicit, I guess having "Other changes: none" in there is also misleading, since there are other, non-interesting changes. I agree, it could be debatable whether to explictly mention that there are no features or bug fixes in a release, if there are not both but maybe probably less prominent. Currently, 2/3 of the changelog are about stating that there are no changes and only 1/3 that there is a bugfix while the latter seems to be the most important information.

PR #577, opened 2023-02-27
[Draft] Add needed status check on the merge_group

liangwen12year:
When adding an approved PR to the merge queue, the required status checks of the workflows are missing, which prevents the PR getting merged. This commit is intended to triggering merge group checks with Github Actions, so that the workflows will report the needed status checks.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -2,6 +2,11 @@
 name: Ansible Lint
 on:  # yamllint disable-line rule:truthy
   pull_request:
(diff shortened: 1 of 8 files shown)


ffmancera:
This won't be enough I think but it is fine to check if it is the right solution. Let's check what happens when I add it to the merge queue

ffmancera (approved):

ffmancera:
Yay! It seems they are running now! @liangwen12year

ffmancera:
So this should be extended to all required tests

PR #578, opened 2023-02-28
[Draft] Add status check of more workflows to the merge_group

liangwen12year:


ffmancera (approved):

ffmancera:
It looks good but for some reason CodeQL is not being reported (but it is being triggered). Then whe have the DCO problem. DCO is not a github action so we need to see how we can trigger it from merge queue.

PR #580, opened 2023-02-28
ansible-lint: Fix name[template] warnings

liangwen12year:
Jinja templates should only be at the end of 'name'.


liangwen12year:
[citest]

tests/playbooks/tests_eth_dns_support.yml
```diff
@@ -10,7 +10,7 @@
 
 
   tasks:
-    - name: "set type={{ type }} and interface={{ interface }}"
+    - name: "set type=veth and interface={{ interface }}"
```
ffmancera, line 13: What about `set interface and type to {{ interface }} {{ type }}` this should avoid triggering the error and it will continue using the variable. What do you think?

ffmancera: Also this applies to others similar

richm: Not sure.  In general, Ansible considers using `{{ ... }}` in the `name:` of a task - at all, anywhere - to be a bad idea.  For one, it makes "grepping" the source code for errors from the logs much harder.

liangwen12year: @richm , so just avoid using `{{...}}` in the name ?  And replace the variable with the real value?

liangwen12year: >I think for the work you are doing, in order not to create more work, if you have already fixed name: values to have a single {{ ... }} at the end of the string, keep it. Do not change it.

>For name: values that still have one or more {{ ... }} in them not at the end of the string, consider rewriting the name: value to get rid of all {{ ... }} in the name: value.

If so, then this PR is ready for review.

ffmancera (dismissed):

tyll:
What is the reason for "Jinja templates should only be at the end of 'name'."? It makes the names harder to understand and also hard codes names where content should depend on a variable.

richm:
> What is the reason for "Jinja templates should only be at the end of 'name'."? It makes the names harder to understand and also hard codes names where content should depend on a variable.

https://ansible-lint.readthedocs.io/rules/name/

> name[template] - Jinja templates should only be at the end of 'name'. This helps with the identification of tasks inside the source code when they fail. The use of templating inside name keys is discouraged as there are multiple cases where the rendering of the name template is not possible.

For example, if you get a message in the Ansible output, and you want to `grep` the role to find out where it comes from, if the `name` has a lot of templating in it, it will make using `grep` quite difficult.

tyll:
> [ansible-lint.readthedocs.io/rules/name](https://ansible-lint.readthedocs.io/rules/name/)
> 
> > name[template] - Jinja templates should only be at the end of 'name'. This helps with the identification of tasks inside the source code when they fail. The use of templating inside name keys is discouraged as there are multiple cases where the rendering of the name template is not possible.
> 
> For example, if you get a message in the Ansible output, and you want to `grep` the role to find out where it comes from, if the `name` has a lot of templating in it, it will make using `grep` quite difficult.

Thank you for the explanation. IMHO this justification belongs into the  commit message. Given that the integration tests and their output are targeting the developers, is the problem with "grep" an actual issue that anyone dealing with the tests actually experienced?

richm:
> > [ansible-lint.readthedocs.io/rules/name](https://ansible-lint.readthedocs.io/rules/name/)
> > > name[template] - Jinja templates should only be at the end of 'name'. This helps with the identification of tasks inside the source code when they fail. The use of templating inside name keys is discouraged as there are multiple cases where the rendering of the name template is not possible.
> > 
> > 
> > For example, if you get a message in the Ansible output, and you want to `grep` the role to find out where it comes from, if the `name` has a lot of templating in it, it will make using `grep` quite difficult.
> 
> Thank you for the explanation. IMHO this justification belongs into the commit message. Given that the integration tests and their output are targeting the developers, is the problem with "grep" an actual issue that anyone dealing with the tests actually experienced?

I have experienced problems when I first started working on system roles and debugging problems e.g. as a system role developer unfamiliar with the network role, trying to debug errors reported in the logs, I could not easily figure out which task was causing the error.  For system roles developers, this has subsequently been solved by running our tests using `ansible-playbook -vv` which shows the file+line number of the task.  However, if we get logs from customers without `-vv`, it may be difficult for a support engineer, or someone unfamiliar with the network role, to know which parts of the log message are variable and which are fixed, and craft an appropriate `grep` regexp.

Ultimately it falls on the network role team to decide how best to support the network role.  If having the names of interfaces, devices, etc. in the ansible logs makes it easier to support the role, then use Jinja templates in the `name:` values, and add ansible-lint suppressions.  I'm fine with whatever you decide to do.

tyll:
The templates are only used in the test playbooks, not in the actual role playbook. So customers, support engineers etc will only greppable names from https://github.com/linux-system-roles/network/blob/main/tasks/main.yml

@ffmancera @liangwen12year what is your preference?

ffmancera:
> The templates are only used in the test playbooks, not in the actual role playbook. So customers, support engineers etc will only greppable names from https://github.com/linux-system-roles/network/blob/main/tasks/main.yml
> 
> @ffmancera @liangwen12year what is your preference?

I do not have a strong preference here to be honest. IMO, we should use templates when it is really needed and otherwise hardcode the values. Mainly because I am grepping the code all the time when developing so it would make my life easier. But as I said, I do not have a strong preference so this is why I approved it if it is fine for @richm and @liangwen12year.

liangwen12year:
> The templates are only used in the test playbooks, not in the actual role playbook. So customers, support engineers etc will only greppable names from https://github.com/linux-system-roles/network/blob/main/tasks/main.yml
> 
> @ffmancera @liangwen12year what is your preference?

I understood the concern from @richm for debugging purposes, I also understood that the variable and the task name may diverge if we hardcode the name of the type and the interface in the task name. But each test playbook is targeting specific feature support, honestly, I did not anticipate any change in the interface type for a specific test (maybe I am wrong). I prefer to leave it as the current changes in this PR, maybe in the future, we can avoid using the Jinja templates in the task name.

tyll:
> I also understood that the variable and the task name may diverge if we hardcode the name of the type and the interface in the task name.

As mentioned in https://github.com/linux-system-roles/network/pull/580#discussion_r1128469262_ the problem is that the type is hardcoded in the task name but in the corresponding task.

tyll:
> > > I also understood that the variable and the task name may diverge if we hardcode the name of the type and the interface in the task name.
> > 
> > 
> > As mentioned in [#580 (comment)](https://github.com/linux-system-roles/network/pull/580#discussion_r1128469262)_ the problem is that the type is hardcoded in the task name but in the corresponding task.
> 
> Since using the jinja templates in the task name only appeared in the test playbooks, then we may use `# noqa name` to suppress the ansible-lint 6.x warning.

I missed a "not" in my sentence:

As mentioned in https://github.com/linux-system-roles/network/pull/580#discussion_r1128469262_ the problem is that the type is hardcoded in the task name but **not** in the corresponding task. So hardcode it in both places or not, but don't mix it please.

liangwen12year:
> > > > I also understood that the variable and the task name may diverge if we hardcode the name of the type and the interface in the task name.
> > > 
> > > 
> > > As mentioned in [#580 (comment)](https://github.com/linux-system-roles/network/pull/580#discussion_r1128469262)_ the problem is that the type is hardcoded in the task name but in the corresponding task.
> > 
> > 
> > Since using the jinja templates in the task name only appeared in the test playbooks, then we may use `# noqa name` to suppress the ansible-lint 6.x warning.
> 
> I missed a "not" in my sentence:
> 
> As mentioned in [#580 (comment)](https://github.com/linux-system-roles/network/pull/580#discussion_r1128469262)_ the problem is that the type is hardcoded in the task name but **not** in the corresponding task. So hardcode it in both places or not, but don't mix it please.

Thanks for the clarification, I updated it.

ffmancera:
I personally don't like the usage of `#noqa` if there is not a good reason for it. When using `#noqa` it makes ansible-lint pointless. I understand @richm concerns and in my honest opinion the right solution is to drop both variables in that case and hardcode them. @tyll what is your opinion here?

Again, I do not want to block on this.. as in my opinion is more important to get things done and reduce the ansible-lint errors as much as possible.

richm:
> I personally don't like the usage of `#noqa` if there is not a good reason for it. When using `#noqa` it makes ansible-lint pointless.

IMO, the use of `# noqa` makes ansible-lint *more* useful, because it allows you to ignore *specific* items deliberately and explicitly, rather than putting things into .ansible-lint `skip_list` or `warn_list`.

The other nice thing about using `# noqa` inline in the code is that it stays with the code - we don't have to worry about transforming the .ansible-lint directives when we convert the network role into the collection for Galaxy and Automation Hub.

Of course, having too many `# noqa` inline in the code makes it a bit ugly, but so far it isn't too many.

> I understand @richm concerns and in my honest opinion the right solution is to drop both variables in that case and hardcode them. @tyll what is your opinion here?
> 
> Again, I do not want to block on this.. as in my opinion is more important to get things done and reduce the ansible-lint errors as much as possible.

PR #582, opened 2023-03-02
ansible-lint: Fix parser-error failure

liangwen12year:


.ansible-lint
```diff
@@ -1,7 +1,15 @@
...
+extra_vars:
+  # In order to fix the ansible-lint 6.x parser-error, the file
+  # `tests/get_coverage.yml` has to be declared as the playbook and an extra
+  # variable `test_playbook` has to be defined. The value of the variable
+  # `test_playbook`is less important which only meets the purpose of passing
+  # the ansible-lint 6.x check
+  test_playbook: tests_default.yml
 kinds:
   - yaml: "**/meta/collection-requirements.yml"
+  - playbook: "**/tests/get_coverage.yml"
```
ffmancera, line 12: I am not sure this is the right solution. Wouldn't it be better to configure .ansible-lint to skip `get_coverage.yml`?

richm: > I am not sure this is the right solution. Wouldn't it be better to configure .ansible-lint to skip `get_coverage.yml`?

No, I think it is better to tell ansible-lint that this is a playbook instead of a tasks file

ffmancera: In that case it LGTM!

ffmancera (approved):

.ansible-lint
liangwen12year: And notably, the current solution also addressed the schema[tasks] failure in "./tests/get_coverage.yml".

PR #589, opened 2023-03-17
ansible-lint: Fix no-changed-when failure

liangwen12year:


liangwen12year:
[citest]

richm:
> I see the same problem of #587, it seems to me that setting it to `false` is just the same than `#noqa`. Not sure if this is the right solution and would like to get your opinion.

I think using `changed_when: false` is very different than using `ignore_errors: true   # noqa ignore-errors` or using `failed_when: false`.  I think all of these cases in this PR it is appropriate to use `changed_when: false`, especially since we don't care if the tests are idempotent.

liangwen12year:
[citest]

ffmancera (dismissed):
I do not want to block here on my comment so LGTM

liangwen12year:
@ffmancera , I rebased and resolved the conflicts.

liangwen12year:
[citest]

PR #595, opened 2023-03-28
in lsr_assert_when use condition instead of when

richm:
ansible-lint is confused by the use of `when:` with `lsr_assert_when`.
It thinks the `when` string should be evaluated as a Jinja expression
but it is really just a plain string.  To make this more friendly to
ansible-lint, use `condition:` instead of `when:` with
`lsr_assert_when`.

### DIFF
--- a/tests/playbooks/tests_states.yml
+++ b/tests/playbooks/tests_states.yml
@@ -31,7 +31,7 @@
               # Device should be present because of autoconnect: true by
               # default for NM (this might be considered a bug)
               - what: tasks/assert_device_present.yml
-                when: network_provider == 'nm'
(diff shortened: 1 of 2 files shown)


richm:
[citest]

nhosoi (dismissed):
lgtm

+1 to "not using a keyword as a variable name." :)

richm:
> lgtm
> 
> +1 to "not using a keyword as a variable name." :)

yeah - that too

ffmancera (approved):

richm:
[citest]

liangwen12year (approved):
LGTM.

PR #603, opened 2023-04-24
ci: Add commitlint GitHub action to ensure conventional commits

spetrosi:
For more information, see Conventional Commits format in Contribute
https://linux-system-roles.github.io/contribute.html#conventional-commits-format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+[1.11.3] - 2023-04-13
+--------------------
+
+### Other Changes
+
+- ansible-lint: Fix ansible-lint issues in tests
+- Add README-ansible.md to refer Ansible intro page on linux-system-roles.github.io
+- Fingerprint RHEL System Role managed config files
+
 [1.11.2] - 2023-02-20
 --------------------
 


.ansible-lint
```diff
@@ -17,9 +16,12 @@ kinds:
   - playbook: "**/examples/*.yml"
 skip_list:
   - fqcn-builtins
+warn_list:
+  - load-failure
```
liangwen12year, line 20: It seems that the ansible-lint changes here are not intentional, right?  Because these changes would make the ansible-lint broken.

.yamllint.yml
```diff
@@ -21,8 +21,6 @@ rules:
 
       /tests/tests_ethtool_ring_initscripts.yml
 
-      /tests/tests_route_device_initscripts.yml
```
liangwen12year, line 24: And why this line was dropped?

richm: Right - all .ansible-lint and .yamllint.yml changes should be made https://github.com/linux-system-roles/.github/blob/main/inventory/host_vars/network.yml

spetrosi: I fixed this, extra_vars: in ansible lint have a different order but that's expected

liangwen12year: I see.

PR #606, opened 2023-04-27
style: ansible-lint - remove line-length files from .yamllint.yml

richm:
Rather than having to maintain .yamllint.yml with a list of files
exempt from yamllint line-length checking, either fix the code
to abide by the line length restriction, or put the yamllint
exceptions in-line in the code.


richm:
NOTE: Most of these changes are due to adding the "disable rule:line-length" to ensure_provider_tests.py which updated *all* of the generated files with this.

liangwen12year (approved):
LGTM.

PR #611, opened 2023-05-01
feat: Support configuring auto-dns setting

liangwen12year:
reason: The user wants to ignore auto dns record (DHCP dns record) via toggling the auto-dns setting.

result: The user is able to ignore auto dns record.


liangwen12year:
[citest]

module_utils/network_lsr/argument_validator.py
```diff
@@ -2461,6 +2463,14 @@ def _ipv6_is_not_configured(connection):
                     "if you need to use initscripts.",
                 )
 
+        # initscripts does not support ip.auto_dns, so raise errors when network
+        # provider is initscripts
```
richm, line 2467: What about `PEERDNS`? https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ip_networking_with_ifcg_files

or is it that we just don't want to provide any new features for the initscripts provider?

tests/playbooks/tests_auto_dns.yml
```diff
@@ -0,0 +1,80 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Test output device of routes
```
richm, line 3: `Test auto_dns settings`?

tests/playbooks/tests_auto_dns.yml
```diff
@@ -0,0 +1,80 @@
...
+      include_tasks: tasks/show_interfaces.yml
+    - name: Include the task 'manage_test_interface.yml'
+      include_tasks: tasks/manage_test_interface.yml
+      vars:
+        state: present
+    - name: Include the task 'assert_device_present.yml'
+      include_tasks: tasks/assert_device_present.yml
+
+    - name: Test the route or the warning log when configuring the route with
+        or without the interface name
```
richm, line 20: Is this really what it's testing?

tests/playbooks/tests_auto_dns.yml
```diff
@@ -0,0 +1,80 @@
...
+      block:
+        - name: Configure an interface with auto_dns enabled
```
richm, line 22: the test uses `auto_dns: false` - shoud the name be `Configure an interface with auto_dns disabled`?

richm:
README.md?

module_utils/network_lsr/argument_validator.py
liangwen12year: Yes, unless the user explicitly requested the new feature for the initscripts provider.

liangwen12year:
[citest]

liangwen12year:
> README.md?

I updated the Readme. Thanks.

library/network_connections.py
```diff
@@ -1173,6 +1173,18 @@ def connection_create(self, connections, idx, connection_current=None):
...
+                if ip["ipv4_ignore_auto_dns"]:
+                    s_ip4.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, True)
+                else:
+                    s_ip4.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, False)
+
+            if ip["ipv6_ignore_auto_dns"] is not None:
+                if ip["ipv6_ignore_auto_dns"]:
+                    s_ip6.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, True)
+                else:
+                    s_ip6.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, False)
```
thom311, line 1186: just:
```python
s_ip4.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, bool(ip["ipv4_ignore_auto_dns"]))
s_ip6.set_property(NM.SETTING_IP_CONFIG_IGNORE_AUTO_DNS, bool(ip["ipv6_ignore_auto_dns"]))
```

liangwen12year:
[citest]

tyll:
What is the reason to have two settings for IPv4 and IPv6? The acceptance criteria in https://issues.redhat.com/browse/NMT-507 mention only one setting "auto-dns" and Nmstate also only uses one [auto-dns setting](https://nmstate.io/devel/yaml_api.html#auto-dns) and the commit message also does not clarify the reason to distinguish between IPv4 and IPv6.

liangwen12year:
> What is the reason to have two settings for IPv4 and IPv6? The acceptance criteria in https://issues.redhat.com/browse/NMT-507 mention only one setting "auto-dns" and Nmstate also only uses one [auto-dns setting](https://nmstate.io/devel/yaml_api.html#auto-dns) and the commit message also does not clarify the reason to distinguish between IPv4 and IPv6.

The reason that we have two settings for IPv4 and IPv6 is to be compatible with NM (`ipv4.ignore-auto-dns` and `ipv6.ignore-auto-dns`) and nmstate. In nmstate, the user needs to specify the `auto-dns` under `ipv4:` or `ipv6:` section. Also, the user can get auto DNS from DHCPv4, DHCPv6, modem etc, and they might want to ignore auto DNS on Ipv4 but not on IPv6, in this case, it is better to have two settings based on the address family.

ffmancera:
> > What is the reason to have two settings for IPv4 and IPv6? The acceptance criteria in https://issues.redhat.com/browse/NMT-507 mention only one setting "auto-dns" and Nmstate also only uses one [auto-dns setting](https://nmstate.io/devel/yaml_api.html#auto-dns) and the commit message also does not clarify the reason to distinguish between IPv4 and IPv6.
> 
> The reason that we have two settings for IPv4 and IPv6 is to be compatible with NM (`ipv4.ignore-auto-dns` and `ipv6.ignore-auto-dns`) and nmstate. In nmstate, the user needs to specify the `auto-dns` under `ipv4:` or `ipv6:` section. Also, the user can get auto DNS from DHCPv4, DHCPv6, modem etc, and they might want to ignore auto DNS on Ipv4 but not on IPv6, in this case, it is better to have two settings based on the address family. Therefore, the acceptance criteria need to be updated.

I think the commit message should explain why there are two different properties. IMO, it is fine having them as we have them on nmcli and nmstate. It is true that on Nmstate we only have `auto-dns` but it matters if is set on `ipv4` or `ipv6` section.

ffmancera (changes requested):
Please modify the commit message explaining why there are two different properties for ipv4 and ipv6.

liangwen12year:
> Please modify the commit message explaining why there are two different properties for ipv4 and ipv6.

I modified the commit message.  Thanks.

liangwen12year:
[citest]

ffmancera (approved):

ffmancera:
Checking why python2.6 unit test failed.. not related I think

ffmancera:
Is a tox issue @liangwen12year can you check it out? I am merging this as I do not want to block the work here.

PR #621, opened 2023-06-09
ci: Add pull request template and run commitlint on PR title only

spetrosi:
We now ensure the conventional commits format only on PR titles and not on
commits to let developers keep commit messages targeted for other developers
i.e. describe actual changes to code that users should not care about.
And PR titles, on the contrary, must be aimed at end users.

For more info, see
https://linux-system-roles.github.io/contribute.html#write-a-good-pr-title-and-description

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+Enhancement:
+
+Reason:
+
+Result:
+
+Issue Tracker Tickets (Jira or BZ if any):
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -24,20 +24,6 @@ jobs:
       - name: Install conventional-commit linter
         run: npm install @commitlint/config-conventional @commitlint/cli
 
-      # Finding the commit range is not as trivial as it may seem.
-      #
-      # At this stage, git's HEAD does not refer to the latest commit in the
-      # PR, but rather to the merge commit inserted by the PR. So instead we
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+Enhancement:
+
+Reason:
+
+Result:
+
+Issue Tracker Tickets (Jira or BZ if any):
```
liangwen12year, line 7: IMHO, this is the format for RFE (new feature). And for bugfix, we usually use a different format in bugzilla:
```
Cause: 

Consequence: 

Fix: 

Result: 
```

richm: > IMHO, this is the format for RFE (new feature). And for bugfix, we usually use a different format in bugzilla:
> 
> ```
> Cause: 
> 
> Consequence: 
> 
> Fix: 
> 
> Result: 
> ```

We don't have a good way for users to select a specific template, and there isn't an automated way to do it.  So no matter what we do, the template isn't going to be applicable for all issues, and reviewers will need to guide users in how best to fill out these.  Note that the goal is that we can use these directly in BZ Doc Text without having to go back weeks or months after the fact and fill that in.

See also the discussion in https://github.com/linux-system-roles/ssh/pull/99

liangwen12year: Ok, then I am fine with this template, thanks.

liangwen12year (approved):

PR #623, opened 2023-06-21
ci: Rename commitlint to PR title Lint, echo PR titles from env var

spetrosi:

### DIFF
--- a/.github/workflows/pr-title-lint.yml
+++ b/.github/workflows/pr-title-lint.yml
@@ -1,4 +1,4 @@
-name: Commitlint
+name: PR Title Lint
 on:  # yamllint disable-line rule:truthy
   pull_request:
     types:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-title-lint.yml
```diff
@@ -1,4 +1,4 @@
-name: Commitlint
+name: PR Title Lint
```
liangwen12year, line 1: IMHO, `name: Commitlint` is better because the following job is to run commitlint on PR title.

richm: It depends on whether or not you care about the implementation.  If you care that the action is being implemented by the `commitlint` executable, then I guess it is better.  If you care that the action is linting the PR title, but you don't care about the implementation, then `PR Title Lint` is a better name, because it is not linting the commit, it is linting the PR.

spetrosi: The job that runs commitlint states that it runs commitlint, but the action itself lints PR titles so I think that's more clear as it is now.

PR #627, opened 2023-07-12
ci: ansible-test ignores file for ansible-core 2.15

richm:
Need a new ignores file for ansible-core 2.15 ansible-test


liangwen12year:
This seems to me that we need a new ignore file whenever a new ansible-core is released.

Btw, I noticed that `.sanity-ansible-ignore-2.15.txt` and `.sanity-ansible-ignore-2.14.txt` are identical.

richm:
> This seems to me that we need a new ignore file whenever a new ansible-core is released.

Yes.  I would like to find a better way to do this - or better yet, fix all of these issues.

> Btw, I noticed that `.sanity-ansible-ignore-2.15.txt` and `.sanity-ansible-ignore-2.14.txt` are identical.

Yes, but sometimes they are not e.g. if a new version of python is supported, or an old one is unsupported, or checks are added or deleted.

PR #628, opened 2023-07-13
fix: facts being gathered unnecessarily

richm:
Cause: The comparison of the present facts with the required facts is
being done on unsorted lists.

Consequence: The comparison may fail if the only difference is the
order.  Facts are gathered unnecessarily.

Fix: Use `difference` which works no matter what the order is.  Ensure
that the fact gathering subsets used are the absolute minimum required.

Result: The role gathers only the facts it requires, and does
not unnecessarily gather facts.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -132,3 +132,9 @@ __network_required_facts:
   - distribution_version
   - os_family
   - python
+
+# the subsets of ansible_facts that need to be gathered in case any of the
+# facts in required_facts is missing; see the documentation of
+# the 'gather_subset' parameter of the 'setup' module
+__network_required_facts_subsets: "{{ ['!all', '!min'] +
+  __network_required_facts }}"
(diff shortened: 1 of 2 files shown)


spetrosi (approved):

liangwen12year (approved):

richm:
[citest]

richm:
centos and rhel8 failures are due to copr and/or epel repo issues:

```
[MIRROR] kernel-core-4.18.0-499.el8.x86_64.rpm: Status code: 404 for https://koji.mbox.centos.org/pkgs/packages/kernel/4.18.0/499.el8/x86_64/kernel-core-4.18.0-499.el8.x86_64.rpm (IP: 8.43.84.204)
```

fedora failure is due to network error:
```
Can not load RPM file: kernel-modules*.rpm.
Could not open: kernel-modules*.rpm
```

PR #633, opened 2023-08-24
ci: Remove badges from README.md prior to converting to HTML

spetrosi:
- Remove thematic break after badges
- Remove badges from README.md prior to converting to HTML


liangwen12year (commented):
Maybe explain the reason that necessitates this commit ?

.github/workflows/markdownlint.yml
```diff
@@ -34,5 +34,5 @@ jobs:
         with:
           args: >-
             --ignore=CHANGELOG.md
-            **/*.md 
+            **/*.md
```
liangwen12year, line 37: This should be in another commit ?

spetrosi: I am doing it via automation, so pushing all in a single commit. That's a minor change, last time I made a bug that kept this extra space, fixing it here. I believe it's fine to have it within the fixing commit. This should be the last PR for badges, I made a lot of noise already anyway...

PR #635, opened 2023-09-13
test: Drop the workaround for configuring IP address to DHCP server

liangwen12year:
NM already fixed the bug which causes the IP addresses of the DHCP server wrongly removed, which can be tracked in
https://bugzilla.redhat.com/show_bug.cgi?id=2079054.

Therefore, there is no reason to still keep the workaround.

### DIFF
--- a/tests/tasks/create_test_interfaces_with_dhcp.yml
+++ b/tests/tasks/create_test_interfaces_with_dhcp.yml
@@ -43,17 +43,8 @@
       nmcli d set testbr managed false
     fi
     ip link set testbr up
-    timer=0
-    # The while loop following is a workaround for the NM bug, which can be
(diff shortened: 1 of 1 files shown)


richm:
Was the issue fixed on all applicable platforms - rhel6, rhel7, rhel8, rhel9, fedora?
[citest]

liangwen12year:
> Was the issue fixed on all applicable platforms - rhel6, rhel7, rhel8, rhel9, fedora? [citest]

The issue was fixed since rhel-8.7, rhel-9.1, and fedora 37.

liangwen12year:
Since we might still see the failure - `tests_bond_options_nm.yml` on rhel-7/centos7 downstream 1minutetip testing, close this PR for now.

PR #636, opened 2023-09-20
test: Use variable to hold infiniband interface name [citest skip]

liangwen12year:
Enhancement: Use variable to hold infiniband interface name

Reason: To avoid hard coded interface name used all over the place in `tests_infiniband.yml`, specify a variable to hold the variable name instead.

### DIFF
--- a/tests/playbooks/tests_infiniband.yml
+++ b/tests/playbooks/tests_infiniband.yml
@@ -49,7 +49,7 @@
             that:
               - virtual_device_setting.stdout is search("ipoib pkey 0x800a mode
                 datagram")
-            msg: the settings in the virtual device
+            msg: the settings in the virtual device do not
               match the specified settings
 
         - name: "** TEST check IPv4"
--- a/tests/playbooks/tests_infiniband.yml
+++ b/tests/playbooks/tests_infiniband.yml
@@ -2,6 +2,9 @@
 ---
 - name: Play for testing infiniband device
   hosts: all
+  vars:
+    interface: "{{ lookup('env', 'INFINIBAND_INTERFACE_NAME') |
+      default('ib0', true) }}"
   tags:
     - "tests::infiniband"
   tasks:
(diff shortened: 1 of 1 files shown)


spetrosi:
[citest]

richm:
How will this be used?  How will the user override the variable name?  How will the user know what will be the generated name for the infiniband interface?

tests/playbooks/tests_infiniband.yml
```diff
@@ -2,6 +2,8 @@
 ---
 - name: Play for testing infiniband device
   hosts: all
+  vars:
+    interface: ib0
```
richm, line 6:
```suggestion
    interface: "{{ if lookup('env', 'INFINIBAND_INTERFACE_NAME')
      then lookup('env', 'INFINIBAND_INTERFACE_NAME')
      else 'ib0' }}"
```
Then the name can be set in various ways.  The most convenient is setting env. var. - in most test frameworks we use, it is easy to add an env. var. to the test environment (e.g. 1minutetip, beaker, etc.) e.g.
```bash
INFINIBAND_INTERFACE_NAME=ibxyz runtest.sh ....
```
you can also use `ansible-playbook -e interface=ibxyz` but then `interface` is used by many network tests, so you would need to be careful to only use it for this test.

We should never require the modification of a test playbook.

liangwen12year: Then, I agree to add a variable to the test environment. let me prepare an update.

tests/playbooks/tests_infiniband.yml
```diff
@@ -2,6 +2,9 @@
 ---
 - name: Play for testing infiniband device
   hosts: all
+  vars:
+    interface: "{{ lookup('env', 'INFINIBAND_INTERFACE_NAME') |
+      default('INFINIBAND_INTERFACE_NAME', 'ib0') }}"
```
richm, line 7:
```suggestion
      default(lookup('env', 'INFINIBAND_INTERFACE_NAME'), 'ib0') }}"
```

richm: if you don't want to have to do the lookup twice - then define another var:
```yaml
vars:
  env_interface: "{{ lookup('env', 'INFINIBAND_INTERFACE_NAME') }}"
  interface: "{{ env_interface | ternary(env_interface, 'ib0') }}"
```

liangwen12year: I actually get a hint from here, https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#providing-default-values, and I tested it successfully locally when the env var `INFINIBAND_INTERFACE_NAME` is defined and not defined.

richm:
[citest]

richm (approved):
QE tests successful, please merge

PR #639, opened 2023-10-03
fix: Add dhcp client package dependency for initscripts provider

liangwen12year:
Enhancement:
   fix: Add dhcp client package dependency for initscripts provider
Reason:
   Initscripts provider requires `/sbin/dhclient` to obtain DHCP address, which is shipped in dhcp client package. But the managed hosts may not already install the dhcp client package, thus, add it as the dependency package for initscripts provider.

Result:


richm (approved):

richm:
[citest]

richm:
Also remove the `tests::expfail` tags from the bond tests

liangwen12year:
[citest]

PR #649, opened 2023-10-12
fix: Allow address 0.0.0.0/0 or ::/0 for 'from'/'to' in a routing rule

liangwen12year:
Enhancement:
fix: Allow address 0.0.0.0/0 or ::/0 for 'from'/'to' in a routing rule

Reason:
`from 0.0.0.0/0` means from all IPv4 addresses, `from ::/0` means from all IPv6 addresses. In NM, if `from` property is not specified in a routing rule, NM still appends `from 0.0.0.0/0` or `from ::/0` to the rule. NM also allows to specify `to 0.0.0.0/0` or `to ::/0` in a routing rule, but the connection profiles will only show the `from` setting for the rule.

Result:

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -779,15 +779,27 @@ class ArgValidatorIPRoutingRule(ArgValidatorDict):
                     name,
                     "missing 'table' for the routing rule",
                 )
-
-        if result["from"] is not None:
+        # `from 0.0.0.0/0` means from all IPv4 addresses
+        # `from ::/0` means from all IPv6 addresses
+        # In NM, if `from` property is not specified in a routing rule, NM
+        # still appends `from 0.0.0.0/0` or `from ::/0` to the rule
+        if result["from"] is not None and result["from"]["address"] not in [
+            "0.0.0.0",
+            "::",
+        ]:
             if result["from"]["prefix"] == 0:
                 raise ValidationError(
                     name,
                     "the prefix length for 'from' cannot be zero",
                 )
 
-        if result["to"] is not None:
+        # NM also allows to specify `to 0.0.0.0/0` or `to ::/0` in a routing
+        # rule, but the connection profiles will only show the `from` setting
+        # for the rule
+        if result["to"] is not None and result["to"]["address"] not in [
+            "0.0.0.0",
+            "::",
+        ]:
             if result["to"]["prefix"] == 0:
                 raise ValidationError(
                     name,
--- a/library/network_connections.py
+++ b/library/network_connections.py
@@ -1243,7 +1243,9 @@ class NMUtil:
                         routing_rule["dport"][0],
                         routing_rule["dport"][1],
                     )
-                if routing_rule["from"]:
+                # In NM, when user specifies `from 0.0.0.0/0`` or `from ::/0` in a
+                # routing rule, NM treats it as if the `from` setting is not specified.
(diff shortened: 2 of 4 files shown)


module_utils/network_lsr/argument_validator.py
```diff
@@ -779,15 +779,27 @@ def _validate_post(self, value, name, result):
...
                     "missing 'table' for the routing rule",
                 )
-
-        if result["from"] is not None:
+        # `from 0.0.0.0/0` means from all IPv4 addresses
+        # `from ::/0` means from all IPv6 addresses
+        # In NM, if `from` property is not specified in a routing rule, NM
+        # still appends `from 0.0.0.0/0` or `from ::/0` to the rule
+        if result["from"] is not None and result["from"]["address"] not in [
+            "0.0.0.0",
```
richm, line 787: Above you say "`from 0.0.0.0/0` means from all IPv4 addresses" but here the value is `"0.0.0.0"` - which one is correct?

richm: Is this true for all versions of NM in el7, el8, el9, and Fedora?

richm: Is there some corresponding change that will need to be made in the nmstate implementation?

liangwen12year: > Is there some corresponding change that will need to be made in the nmstate implementation?

There are some corresponding changes that will need to be made in NM, and very likely the update will only include in the NM newest release. And originally, the user requested to specify `from ::/0` in the routing rule, but the user can achieve the same effect by not specifying `from ::/0` in the routing rule (NM will automatically append `from ::/0` to the routing rule). Therefore, I will request that user to not specify `from ::/0` when using the  network role.

richm:
do we need an integration test which configures a routing rule without a `from` or `to`, then checks to see if the from all and/or to all has been added?

liangwen12year:
> do we need an integration test which configures a routing rule without a `from` or `to`, then checks to see if the from all and/or to all has been added?

There is a bug in the current libnm (NM), which will not end up appending `from all` to the routing rule, but raise a validation error instead, I am fixing that bug in NM now.

liangwen12year:
[citest]

richm:
@liangwen12year There are a few ci test failures which should be fixed.

* anything which installs from epel - not sure if we can increase the number of retries, wait interval, or timeouts - might help with the epel repo failures
* fedora 38 - looks like there is an issue with the copr liangwen12year/hostapd-owe ?

richm:
[citest bad]

liangwen12year:
[citest]

richm:
I think this is ready to merge - the failures are not related (and should be addressed in separate PRs)

thom311:
the patch for NetworkManager is not yet merged, and it's not yet exactly clear what the result will be.

The validation in the role really must mimic what NetworkManager does (which will aim to be compatible with `ip rule add`). As such, this needs to wait and then re-implement it.

For example, the check
```python
        if result["to"] is not None and result["to"]["address"] not in [
            "0.0.0.0",
            "::",
        ]:
            if result["to"]["prefix"] == 0:
                raise ValidationError(
```
is not what ip-rule-add does, and it's likely not what NetworkManager will end up doing.

richm:
any updates?  Is the NM change available in any of the composes?

liangwen12year:
> any updates? Is the NM change available in any of the composes?

Maybe instead of waiting for the changes in NM, we can merge this PR, I will file another one if needed afterwards.

PR #650, opened 2023-10-19
feat: support for ostree systems

richm:
Feature: Allow running and testing the role with ostree managed nodes.

Reason: We have users who want to use the role to manage ostree
systems.

Result: Users can use the role to manage ostree managed nodes.

NOTE: The .ostree/packages-*.txt are generated by running the
script https://github.com/linux-system-roles/auto-maintenance/blob/main/check-logs-for-packages.sh
which is used with an integration test CI run with the
https://github.com/linux-system-roles/auto-maintenance/blob/main/callback_plugins/dump_packages.py
plugin.  An example of such a CI run is
https://github.com/linux-system-roles/network/pull/647


richm:
[citest]

liangwen12year:
[citest]

tests/ensure_provider_tests.py
```diff
@@ -86,7 +81,7 @@
     "playbooks/tests_provider.yml": {
         MINIMUM_VERSION: "'1.20.0'",
         "comment": "# NetworKmanager 1.20.0 added support for forgetting profiles",
-        EXTRA_RUN_CONDITION: "ansible_distribution != 'RedHat' or\n      ansible_distr\
+        EXTRA_RUN_CONDITION: "not ansible_distribution in ['RedHat', 'CentOS'] or\n      ansible_distr\
```
liangwen12year, line 84: Is this change necessary? `playbooks/tests_802_1x_updated.yml` and `playbooks/tests_802_1x.yml` also have the the same EXTRA_RUN_CONDITION, but the condition is not changed.

richm: It is not related to this PR and I have removed it.  However, this test fails on centos-9 and should be fixed, probably in the manner I have above (basically, anything that should be done on `RedHat` should also be done on `CentOS`), in a subsequent PR.

.ostree/packages-runtime-CentOS-7.txt
```diff
@@ -0,0 +1,4 @@
+bridge-utils
```
liangwen12year, line 1: If these .txt files are generated by `dump_packages.py` ? maybe it is better to add a line in the commit message explain that how these txt files are  generated.

tests/tests_regression_nm.yml
```diff
@@ -36,5 +33,5 @@
     - ansible_distribution_major_version != '6'
     # The test depends on behavior that is only visible with newer NM
     - networkmanager_version is version('1.22.0', '>=')
-    - ansible_distribution != 'RedHat' or
+    - not ansible_distribution in ['RedHat', 'CentOS'] or
```
liangwen12year, line 36: Drop this change for now or put it into another PR ?

.ostree/packages-runtime-CentOS-7.txt
richm: Done.

tests/tests_regression_nm.yml
richm: Done.

.ostree/get_ostree_data.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
liangwen12year, line 1: I assume this file should be consistent as how it is used in other roles.

tests/ensure_provider_tests.py
```diff
@@ -21,17 +21,12 @@
...
+        - name: Get package info
+          package_facts:
         - name: Get NetworkManager version
           set_fact:
-            networkmanager_version: "{{ networkmanager_info.results |
-              selectattr('yumstate', 'match', '^installed$') |
-              map(attribute='version') | list | first }}"
-          when: true
-          changed_when: false
+            networkmanager_version: "{{
```
liangwen12year, line 28: What is the problem with the previous way of getting the networkmanager_version ?

tests/playbooks/tests_802_1x.yml
```diff
@@ -38,6 +38,9 @@
                     - none
                   client_cert: /etc/pki/tls/client.pem
                   ca_cert: /etc/pki/tls/cacert.pem
+        - name: Ensure ping command is present
+          package:
+            name: iputils
```
liangwen12year, line 43: `state: present` ?

tests/tasks/test_802.1x_capath.yml
```diff
@@ -73,6 +73,9 @@
               private_key_password_flags:
                 - not-required
               ca_path: /etc/pki/tls/my_ca_certs
+    - name: Ensure ping command is present
+      package:
+        name: iputils
```
liangwen12year, line 78: `state_present` ?

.ostree/get_ostree_data.sh
richm: Yes.  It will be managed by https://github.com/linux-system-roles/.github

tests/ensure_provider_tests.py
richm: The use of the `yum` module.  This conflicts with the package manager used by ostree systems.  Using the `package` and `package_facts` modules works correctly on ostree systems.

Why was the `yum` module used?  Is this a performance issue?  Using `yum` directly is much, much faster than using `package_facts`?

tests/playbooks/tests_802_1x.yml
richm: Done.

tests/tasks/test_802.1x_capath.yml
richm: Done.

tests/ensure_provider_tests.py
liangwen12year: The time when I was implementing it, I was not aware of `package` and `package_facts` module can get the NetworkManager version so easily.

tests/tasks/enable_epel.yml
```diff
@@ -4,6 +4,7 @@
   when:
     - ansible_distribution in ['RedHat', 'CentOS']
     - ansible_distribution_major_version in ['7', '8']
+    - ansible_facts["pkg_mgr"] != "ansible.posix.rhel_rpm_ostree"
```
liangwen12year, line 7: This seems redundant, because we already defined the condition `when: ansible_facts["pkg_mgr"] != "ansible.posix.rhel_rpm_ostree"` in some other files.

tests/tests_default.yml
```diff
@@ -6,8 +6,6 @@
   roles:
     - linux-system-roles.network
   tasks:
-    - name: Include the task 'el_repo_setup.yml'
```
liangwen12year, line 9: What is the reason for removing this task ?

tasks/set_facts.yml
```diff
@@ -7,6 +7,24 @@
     difference(ansible_facts.keys() | list) | length > 0
   no_log: true
 
+- name: Ensure correct package manager for ostree systems
```
liangwen12year, line 10: I can not understand what this task is intended for, maybe add a comment explaining in detail about the purpose of this task ?

richm: Well, the Ansible guidelines state that you should use the `name:` of a task to adequately describe what the task does.  So how about this:

`- name: Ensure the use of the ostree package manager on ostree systems`

If that is still not clear enough, I guess I would have to ask what is unclear.

richm: I could add a section to the README-ostree.md which describes in detail things like
- why isn't this functionality in ansible-core?
- what determines if a system is an ostree system?
- why does an ostree system require an "external" package manager for use by the `package` and `package_facts` modules?
- other faqs?

richm: Is this good now, as-is?  Further changes required?

tests/tasks/enable_epel.yml
richm: Fixed.

tests/tests_default.yml
richm: I put it back

tasks/set_facts.yml
liangwen12year: I have doubts about why an ostree system require an "external" package manager for use. Now I understand that none of the built-in package managers in Ansible can be used on ostree system, because that we can not install packages in an ostree system. Thanks for explaining it.

liangwen12year:
All the container tests failed in
```
TASK [Include the task 'el_repo_setup.yml'] ************************************
fatal: [localhost]: FAILED! => {"reason": "couldn't resolve module/action 'ansible.utils.update_fact'. This often indicates a misspelling, missing collection, or incorrect module path.\n\nThe error appears to be in '/network-role/tests/tasks/el_repo_setup.yml': line 27, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Set package manager to use for ostree\n      ^ here\n"}
```

liangwen12year:
[citest]

liangwen12year (approved):

PR #655, opened 2023-11-20
refactor: improve support for ostree systems

richm:
The dependency on `ansible.utils.update_fact` is causing issue with
some users who now must install that collection in order to run
the role, even if they do not care about ostree.

The fix is to stop trying to set `ansible_facts.pkg_mgr`, and instead
force the use of the ostree package manager with the `package:` module
`use:` option.  The strategy is - on ostree systems, set the flag
`__$ROLENAME_is_ostree` if the system is an ostree system.  The flag
will either be undefined or `false` on non-ostree systems.
Then, change every invocation of the `package:` module like this:

```yaml
- name: Ensure required packages are present
  package:
    name: "{{ __$ROLENAME_packages }}"
    state: present
    use: "{{ (__$ROLENAME_is_ostree | d(false)) |
      ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
```

This should ensure that the `use:` parameter is not used if the system
is non-ostree.  The goal is to make the ostree support as unobtrusive
as possible for non-ostree systems.
The user can also set `__$ROLENAME_is_ostree: true` in the inventory or play
if the user knows that ostree is being used and wants to skip the check.
Or, the user is concerned about the performance hit for ostree detection
on non-ostree systems, and sets `__$ROLENAME_is_ostree: false` to skip
the check.
The flag `__$ROLENAME_is_ostree` can also be used in the role or tests to
include or exclude tasks from being run on ostree systems.

This fix also improves error reporting in the `get_ostree_data.sh` script
when included roles cannot be found.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -25,5 +25,3 @@ exclude_paths:
   - examples/roles/
 mock_roles:
   - linux-system-roles.network
-mock_modules:
-  - ansible.utils.update_fact
--- a/.ostree/get_ostree_data.sh
+++ b/.ostree/get_ostree_data.sh
@@ -2,7 +2,6 @@
 
 set -euo pipefail
 
-role_collection_dir="${ROLE_COLLECTION_DIR:-fedora/linux_system_roles}"
 ostree_dir="${OSTREE_DIR:-"$(dirname "$(realpath "$0")")"}"
 
 if [ -z "${4:-}" ] || [ "${1:-}" = help ] || [ "${1:-}" = -h ]; then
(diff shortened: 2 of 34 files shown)


richm:
[citest]

richm:
@liangwen12year I've noticed that tests_network_state.yml is failing with recent rhel8 and rhel9 - the task `name: "** TEST check resolvectl"` fails - `example.com` is not in the `resolvectl` output - not sure if something has changed with recent rhel8/rhel9.

richm:
[citest]

spetrosi (approved):
lgtm

PR #659, opened 2023-12-06
refactor: get_ostree_data.sh use env shebang - remove from .sanity*

richm:
Use the `#!/usr/bin/env bash` shebang which is ansible-test friendly.
This means we can remove get_ostree_data.sh from the .sanity* files.
This also means we can remove the .sanity* files if we do not need
them otherwise.  Fix other shell scripts to use the friendly shebang
and remove from the .sanity* files.

Rename `pth` to `path` in honor of nscott

### DIFF
--- a/.github/run_test.sh
+++ b/.github/run_test.sh
@@ -1,6 +1,6 @@
-#!/bin/bash -x
+#!/usr/bin/env bash
 
-set -euo pipefail
+set -euxo pipefail
 
 TEST_SOURCE_DIR="/network-role"
 C8S_CONTAINER_IMAGE="quay.io/linux-system-roles/c8s-network-role"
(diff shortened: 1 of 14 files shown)


richm:
> There are other shell files using the `#!/bin/bash`, why the changes only happen in `get_ostree_data.sh` ?

Is it ok with you if this PR includes changes to tests/ - covstats ensure_provider_tests.py get_coverage.sh etc.?

liangwen12year:
> > There are other shell files using the `#!/bin/bash`, why the changes only happen in `get_ostree_data.sh` ?
> 
> Is it ok with you if this PR includes changes to tests/ - covstats ensure_provider_tests.py get_coverage.sh etc.?

Since ansible-test is checking other shell scripts,  let us replace them all with `#!/usr/bin/env bash`.

liangwen12year:
And the commit message should also be updated correspondingly.

PR #661, opened 2023-12-14
ci: Use supported ansible-lint action; run ansible-lint against the collection

richm:
ci: Use supported ansible-lint action; run ansible-lint against the collection

The old ansible-community ansible-lint is deprecated.  There is a
new ansible-lint github action.  The new ansible-lint has several checks
related to ansible-test and the ignore files.  Many of our ignore settings
are not allowed any more and are required to be fixed or addressed in the
Ansible preferred way.

The python imports have to be wrapped in a try/except ImportError, and
where possible, an error must be returned from the module explaining
what was not able to be imported.

The module documentation must comply with the Ansible standards.  One
aspect of this is the `version_added` must be a valid ansible-core
version in X.Y.Z format.  Note that this version isn't really used
anywhere, so it doesn't matter for users of the role, it is purely
an `ansible-test` and import gating issue.

The result of this is that the .sanity files can be reduced to the
bare minimum which will greatly reduce the maintenance burden of
those files, make it easier to support newer versions of Ansible,
and make it easier to import the system roles collection into Galaxy
and Automation Hub.

The latest Ansible repo gating tests run ansible-lint against
the collection format instead of against individual roles.
We have to convert the role to collection format before running
ansible-test.

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

Add `---` doc start to .markdownlint.yaml

The file `examples/down_profile+delete_interface.yml`
was not used and was causing ansible-lint errors.

ansible-lint enforces the order of keywords in plays - `name`,
then `hosts`, then `vars`, then `tasks`.


richm:
[citest]

library/network_state.py
```diff
@@ -22,15 +22,38 @@
 author: "Wen Liang (@liangwen12year)"
 """
 
+EXAMPLES = r"""
```
liangwen12year, line 25: To make the format consistent, should we also add the `options: desired_state:` for network_state ?

richm: > To make the format consistent, should we also add the `options: desired_state:` for network_state ?

It was already there - https://github.com/linux-system-roles/network/blob/1a689f7eedbdd9b1adbd93552700bb817aa419df/library/network_state.py#L17

One thing to note is that this internal module documentation is intended for end users to call these modules directly, without using the role.  This use case does not work and is unsupported.  For us, this internal role documentation and formatting is really only so that we can pass ansible-lint and ansible-test gating tests.  For us, the main documentation is the README.md for the role, which describes the working and supported interface.  Users are not supposed to know or care that there are modules used internally in the role to implement the functionality of the role.

tests/integration/conftest.py
```diff
@@ -1,8 +0,0 @@
-# -*- coding: utf-8 -*
-# SPDX-License-Identifier: BSD-3-Clause
-
-
-def pytest_addoption(parser):
-    parser.addoption(
-        "--provider", action="store", default="nm", help="Network provider"
```
liangwen12year, line 7: I have checked the commit c29a2bf70813c035fca1be1fdd7a06833a17e47d, is not the hook used with `tests/integration/test_ethernet.py` ?

richm: > I have checked the commit [c29a2bf](https://github.com/linux-system-roles/network/commit/c29a2bf70813c035fca1be1fdd7a06833a17e47d), is not the hook used with `tests/integration/test_ethernet.py` ?

I see it in that commit - but where/how is it used?  doing `git grep conftest` in the `network` role local git clone shows nothing

liangwen12year:
[citest]

tests/integration/conftest.py
liangwen12year: It seems to me that the `tests_integration_pytest.yml`  failure is related to the changes, I think that the initialization hooks are the common usage of `conftest.py`.

1. https://docs.pytest.org/en/7.1.x/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options
2. https://docs.pytest.org/en/6.2.x/reference.html#initialization-hooks

richm: ok - I see - conftest.py is used internally by pytest - so no explicit reference - I added it back

richm:
[citest]

richm:
[citest bad]

module_utils/network_lsr/nm/__init__.py
```diff
@@ -1,11 +0,0 @@
...
-# pylint: disable=import-error, no-name-in-module
-from __future__ import absolute_import, division, print_function
-
-__metaclass__ = type
-
-from ansible.module_utils.network_lsr.nm import provider  # noqa:E501
-
-# pylint: enable=import-error, no-name-in-module
-
-provider.NetworkManagerProvider
```
liangwen12year, line 11: What is the reason to remove this file ?

richm: > What is the reason to remove this file ?

https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/empty-init.html

It appears that this isn't being used directly - but perhaps it is needed to make the unit tests work - not sure - but for sure we need to make this an empty file, and figure out some other way to make the unit tests work if needed

richm: I don't see any unit test failures though - so IMO it isn't needed

The only test failures I see are

* tests_provider_nm and tests_regression_nm on CentOS 9 - I don't think we need to run these tests on platforms where initscripts is not supported
* tests_bond_options_nm on CentOS 8 - maybe the test is disabling network access?  cannot connect to the EPEL repo

liangwen12year: For skipping initscripts tests, I have the PR for review, https://github.com/linux-system-roles/network/pull/663

for the tests_bond_options_nm on CentOS 8, the test is not fully started yet, it fails in "INIT Prepare setup" where dnsmasq needs to be installed, maybe the EPEL repo can not be accessed itself sometime ?  Also I noticed that some other bond tests (e.g. tests_bond_nm.yml) may fail for the same reason, and tests_bond_options.yml can also pass sometimes. see:

https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-663-70e4146_CentOS-Stream-8_20231220-205206/artifacts/summary.html
https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_network-663-70e4146_CentOS-Stream-8_20231220-205218/artifacts/summary.html

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest bad]

liangwen12year (approved):

liangwen12year:
[citest bad]

PR #668, opened 2024-01-06
ci: Use new ansible-lint action, which requires collection format and 2.16

richm:
The old ansible-community ansible-lint github action is deprecated.
There is a new ansible-lint github action called `ansible/ansible-lint`.
However, this version requires a collection, so we have to convert
the role to collection format before running the ansible-lint action.

tox-lsr has been updated to use ansible-core 2.16 for ansible-lint and
ansible-test.

This also requires tox-lsr 3.2.1 - bump other actions to use 3.2.1

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -11,6 +11,9 @@ on:  # yamllint disable-line rule:truthy
     branches:
       - main
   workflow_dispatch:
+env:
+  LSR_ROLE2COLL_NAMESPACE: fedora
+  LSR_ROLE2COLL_NAME: linux_system_roles
 permissions:
   contents: read
 jobs:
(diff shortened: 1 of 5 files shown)


liangwen12year (commented):
I run the test locally with `tox -e collection,ansible-lint-collection`, and did not find the problem. But I notice that the commit has the same subject as the the commit in   https://github.com/linux-system-roles/network/pull/661.

liangwen12year:
> I run the test locally with `tox -e collection,ansible-lint-collection`, and did not find the problem. But I notice that the commit has the same subject as the the commit in #661.

Maybe just add a line to the commit message to highlight that this PR is about updating the github action to use ansible role and ansible-test 2.16.

liangwen12year (approved):

PR #675, opened 2024-02-18
fix: Allow network to restart when wireless or team connection is specified

liangwen12year:
Enhancement: Ask user's consent to restart NM due to wireless or team interfaces when the updates for network packages are available.

Reason: When wireless or team connections are specified and the updates for network packages are available, NetworkManager must be restarted, the role requires user's consent to restart NetworkManager. Otherwise, there
might be property conflicts between NetworkManager daemon and plugin, or
NetworkManager plugin is not taking effect.

Result:  When wireless or team connections are specified and the updates for network packages are available, NetworkManager must be restarted, the role will ask user's explicit consent to restart NetworkManager.

Resolves: https://issues.redhat.com/browse/NMT-1039


liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

richm (approved):

tyll:
This seems to be wrong. The role needs to handle this scenario properly (or the NM packages need to be changed).

tyll (changes requested):
The role should handle the scenario instead of doctoring the tests.

liangwen12year:
[citest]

defaults/main.yml
```diff
@@ -50,6 +48,11 @@ __network_packages_default_team: ["{%
       if __network_team_connections_defined
       %}NetworkManager-team{% endif %}"]
 
+# if wireless or team connection is specified, NetworkManager should be
+# allowed to restart by default
+network_allow_restart: "{{ true if __network_wireless_connections_defined
```
richm, line 53:
```yaml
network_allow_restart: "{{ __network_wireless_connections_defined or
  __network_team_connections_defined }}"
```
Ansible style says not to use boolean `true` and `false` if you can just use the result of a boolean context evaluation

liangwen12year:
[citest]

defaults/main.yml
```diff
@@ -50,6 +48,11 @@ __network_packages_default_team: ["{%
       if __network_team_connections_defined
       %}NetworkManager-team{% endif %}"]
 
+# if wireless or team connection is specified, NetworkManager should be
+# allowed to restart by default
+network_allow_restart: "{{ __network_wireless_connections_defined or
+      __network_team_connections_defined }}"
+
```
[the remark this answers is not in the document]
liangwen12year: I got the hint for setting the default value for `network_allow_restart` from the comment here, https://github.com/linux-system-roles/network/blob/4b12e154a69af7fb53b8f409eca3b94a23bbd1f1/tasks/main.yml#L72, I think that the comment should also be updated to reflect that the role needs user's consent to restart NM. I can update the README and the comment.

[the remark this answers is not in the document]
tyll: Failing the role the

```
    - __network_wireless_connections_defined
       or __network_team_connections_defined
    - network_provider == "nm"

    # ansible-lint wants this to be a handler, but this is not appropriate as
    # NetworkManager must be restarted prior to the connections being created.
    # see (https://docs.ansible.com/ansible-lint/rules/default_rules.html)
    - __network_package_install.changed  # noqa no-handler
```

is true but

```
    - network_allow_restart
```

is false with a descriptive error message would be good replacement for the comment IMHO. It is better to make the code clear to understand than having to rely on comments.

liangwen12year:
[citest]

tasks/main.yml
```diff
@@ -69,8 +69,22 @@
...
+  fail:
+    msg: The user needs to consent to restart NetworKManager explicitly
+      by setting `network_allow_restart` to `true` when network packages
+      are changed and wireless or team connections are specified
+  when:
+    - __network_wireless_connections_defined
+       or __network_team_connections_defined
+    - network_provider == "nm"
+    - not network_allow_restart
+    - __network_package_install.changed  # noqa no-handler
```
richm, line 83: What does package install have to do with this?

liangwen12year: Restarting NM is required only when network packages are changed on the system.

richm:
```suggestion
    - __network_package_install is changed  # noqa no-handler
```
Ansible style prefers the `is` test - https://docs.ansible.com/ansible/2.9/user_guide/playbooks_tests.html#task-results - but if other adjacent code doesn't, then leave it as-is

README.md
```diff
@@ -92,11 +92,13 @@ the name prefix. List of variables:
 - `network_connections` - The connection profiles are configured as
   `network_connections`, which is a list of dictionaries that include specific
   options.
-- `network_allow_restart` - Certain configurations require the role to restart
-  network services. For example, if a wireless connection is configured and
-  NetworkManager-wifi is not installed, NetworkManager must be restarted prior
-  to the connection being configured. Setting this to `false` will prevent the
-  role from restarting network service.
+- `network_allow_restart` - It defaults to `false`. Certain configurations like
+  wireless or team connections require the role to restart network services.
```
tyll, line 96: It is not the configuration but the need to install plugins that require the restart.

README.md
```diff
@@ -92,11 +92,13 @@ the name prefix. List of variables:
...
+  For example, if a wireless connection is configured and NetworkManager-wifi
+  is not installed, NetworkManager must be restarted prior to the connection
+  being configured, and the user needs to explicitly consent to it by setting
```
tyll, line 99: Please mention that the restart can result in connectivity loss and therefore the role does not do this without the consent.

tasks/main.yml
```diff
@@ -69,8 +69,22 @@
       ansible_distribution != 'Fedora' and
       ansible_distribution_major_version | int > 8
 
+- name: Ask user's consent to restart NetworkManager due to wireless or team
+    interfaces
+  fail:
+    msg: The user needs to consent to restart NetworKManager explicitly
+      by setting `network_allow_restart` to `true` when network packages
+      are changed and wireless or team connections are specified
```
tyll, line 77: Please change this to imperative:

"NetworkManager needs to be restarted to be able to proceed. This might disturb the connectivity of the managed system. Please set `network_allow_restart` to `true` if you are prepared for this."

Writing this makes me realize that this probably won't work because when the user set the variable to true, then the role won't need to install the packages anymore, therefore it won't restart NM and it will fail afterwards. So it seems that NM needs to be probed to check if it needs to be restarted regardless of whether the packages were just installed.

tyll: This also highlights why it is important to have a test case that covers the scenario.

ffmancera: IMHO, this should be done before installing the packages if possible. At that point we can detect if `network_allow_restart` is set to `true`. Then, the user gets the error and can decide to do it or not.

ffmancera (changes requested):

tasks/main.yml
liangwen12year: A tentative approach could be using `update_cache`, but then we need to write a task  for different package managers (`yum`, `dnf`), like:

```
    - name: Check if updates are available for network packages 
      dnf:
        update_cache: yes
        name: "{{ network_packages }}"
        state: latest
      register: package_info
      check_mode: yes 
      when: 
         -  ansible_distribution == 'Fedora' or ansible_distribution_major_version | int > 7  
      
    - name: Check if updates are available for network packages 
      yum:
        update_cache: yes
        name: "{{ network_packages }}"
        state: latest
      register: package_info
      check_mode: yes 
      when: 
         -   ansible_distribution != 'Fedora' and  ansible_distribution_major_version | int < 8      
```

liangwen12year:
[citest]

liangwen12year:
Since team is deprecated in RHEL-9, I think that it is not appropriate to test team connection in this PR, I will switch to wifi instead tomorrow.

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
@richm , about the test failure in c9s container test, my original idea is to remove the conflicting NM package, and  then check again if the updates for the `network_package` are available, then install the new NM package, but that would necessarily need to restart NM regardless of user's consent.  It is bad because restarting NM  can result in connectivity loss and therefore the role does not allow it without the consent. The problem here is that the local system has the NetworkManager package version `NetworkManager-1:1.46.0-1.el9.x86_64`  installed, but the baseos has the version `NetworkManager-team-1:1.45.91-1.el9.x86_64`  and `NetworkManager-1:1.45.91-1.el9.x86_64`. I honestly do not know why this situation would happen, I can imagine a way to do about it:

1. Add a rescue task in `main.yml` explaining better the error message about the package conflicts 



 
```
TASK [linux-system-roles.network : Check if updates for network packages are available through the DNF package manager due to wireless or team interfaces] ***
fatal: [localhost]: FAILED! => {"changed": false, "failures": [], "msg": "Depsolve Error occured: \n Problem: cannot install both NetworkManager-1:1.45.91-1.el9.x86_64 from baseos and NetworkManager-1:1.46.0-1.el9.x86_64 from @System\n  - package NetworkManager-team-1:1.45.91-1.el9.x86_64 from baseos requires NetworkManager(x86-64) = 1:1.45.91-1.el9, but none of the providers can be installed\n  - cannot install the best update candidate for package NetworkManager-1:1.46.0-1.el9.x86_64\n  - cannot install the best candidate for the job", "rc": 1, "results": []}
```

liangwen12year:
> @richm , about the test failure in c9s container test, my original idea is to remove the conflicting NM package, and then check again if the updates for the `network_package` are available, then install the new NM package, but that would necessarily need to restart NM regardless of user's consent. It is bad because restarting NM can result in connectivity loss and therefore the role does not allow it without the consent. The problem here is that the local system has the NetworkManager package version `NetworkManager-1:1.46.0-1.el9.x86_64` installed, but the baseos repo has the version `NetworkManager-team-1:1.45.91-1.el9.x86_64` and `NetworkManager-1:1.45.91-1.el9.x86_64`. I honestly do not know why this situation would happen, I can imagine a way to do about it:
> 
> 1. Add a rescue task in `main.yml` explaining better the error message about the package conflicts
> 
> ```
> TASK [linux-system-roles.network : Check if updates for network packages are available through the DNF package manager due to wireless or team interfaces] ***
> fatal: [localhost]: FAILED! => {"changed": false, "failures": [], "msg": "Depsolve Error occured: \n Problem: cannot install both NetworkManager-1:1.45.91-1.el9.x86_64 from baseos and NetworkManager-1:1.46.0-1.el9.x86_64 from @System\n  - package NetworkManager-team-1:1.45.91-1.el9.x86_64 from baseos requires NetworkManager(x86-64) = 1:1.45.91-1.el9, but none of the providers can be installed\n  - cannot install the best update candidate for package NetworkManager-1:1.46.0-1.el9.x86_64\n  - cannot install the best candidate for the job", "rc": 1, "results": []}
> ```

Ok, after rebuilding the c9s container image in quay, the github c9s container integration test passed.

liangwen12year:
[citest]

README.md
```diff
@@ -92,11 +92,15 @@ the name prefix. List of variables:
...
-  network services. For example, if a wireless connection is configured and
-  NetworkManager-wifi is not installed, NetworkManager must be restarted prior
-  to the connection being configured. Setting this to `false` will prevent the
-  role from restarting network service.
+- `network_allow_restart` - It defaults to `false`. In order to install and make
+  certain NetworkManager plugin (e.g. NetworkManager-wifi or NetworkManager-team)
+  takes effect, NetworkManager requires to be restarted. For example, if a wireless
+  connection is configured and NetworkManager-wifi is not installed, NetworkManager
+  must be restarted prior to the connection being configured. The restart can result in
+  connectivity loss and therefore the role does not allow it without the consent. The
```
tyll, line 100:
```suggestion
  connectivity loss and therefore the role does not allow it without explicit consent. The
```

README.md
```diff
@@ -92,11 +92,15 @@ the name prefix. List of variables:
...
+  user can consent to it by setting `network_allow_restart` to `true`. Setting
+  `network_allow_restart` to `false` will prevent the role from restarting network
+  service.
```
tyll, line 103:
```suggestion
  `network_allow_restart` to `false` will prevent the role from NetworkManager.
```

tasks/main.yml
```diff
@@ -24,6 +24,48 @@
...
+    - ansible_distribution_major_version | int < 8
+    - __network_wireless_connections_defined
+      or __network_team_connections_defined
+
+- name: Ask user's consent to restart NetworkManager due to wireless or team
+    interfaces
+  fail:
+    msg: NetworkManager needs to be restarted to be able to proceed. This might
+      disturb the connectivity of the managed system. Please set
+      `network_allow_restart` to `true` if you are prepared for this.
```
tyll, line 60: Please explain to the user that this is because wireless or team interfaces are defined. Also, the full necessary action is to install the package and to restart NM. It seems best to be explicit about this, so users can also install the packages/restart themselves if necessary.

tyll (approved):

tyll:
The PR summary is missing the text for Enhancement/Reason/Result.

tasks/main.yml
```diff
@@ -24,6 +24,48 @@
...
+  check_mode: true
+  when:
+    - ansible_distribution == 'Fedora' or
+      ansible_distribution_major_version | int > 7
+    - __network_wireless_connections_defined
+      or __network_team_connections_defined
+
+- name: Check if updates for network packages are available through the YUM
+    package manager due to wireless or team interfaces
+  yum:
```
tyll, line 44: Can https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html be used instead of having two tasks for dnf/yum? What's the reason that `update_cache` is enabled?

tasks/main.yml
```diff
@@ -24,6 +24,48 @@
   when:
     - network_state is defined
     - ansible_distribution_major_version | int < 8
+
+- name: Check if updates for network packages are available through the DNF
+    package manager due to wireless or team interfaces
+  dnf:
+    update_cache: true
+    name: "{{ network_packages }}"
+    state: latest  # noqa package-latest
```
tyll, line 33: There needs to be an explanation why this needs to be latest. Ideally in the commit message IMHO.

tyll (changes requested):
see my detailed comments for requested changes

tests/tasks/assert_network_connections_failed.yml
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Assert that configuring network connections failed
+  assert:
+    that:
+      - __network_service_restart_requested is failed
+      - __network_service_restart_requested is search(
+        network_connections_result_error )
```
richm, line 8:
```suggestion
        network_connections_result_error)
```

tests/tasks/assert_package_installed.yml
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: "Get the rpm package facts"
+  package_facts:
+    manager: "auto"
+
+- name: "Assert installed package '{{ package }}'"
+  assert:
+    that:
+      - "'{{ package }}' in ansible_facts.packages"
```
richm, line 10:
```suggestion
      - package in ansible_facts.packages
```

tests/tasks/create_wireless_profile_restart_network.yml
```diff
@@ -0,0 +1,21 @@
...
+            type: wireless
+            wireless:
+              ssid: "My WPA2-PSK Network"
+              key_mgmt: "wpa-psk"
+              password: "p@55w0rD"
+
+  rescue:
+    - name: Show rescue result
+      debug:
+        var: __network_service_restart_requested
```
richm, line 21: If you want this error to be raised to the caller, you'll need to have a task

```yaml
    - name: Raise error
      fail:
        msg: "{{ ansible_failed_result }}"
```

otherwise, the `rescue` will "eat" the error and the play will proceed normally - not sure if this is the desired behavior

tests/tasks/remove_and_assert_package_absent.yml
```diff
@@ -0,0 +1,16 @@
...
+    state: absent
+
+- name: "Get the rpm package facts"
+  package_facts:
+    manager: "auto"
+
+- name: "Assert removed package '{{ package }}'"
+  assert:
+    that:
+      - "package not in ansible_facts.packages"
```
richm, line 15:
```suggestion
      - package not in ansible_facts.packages
```

tasks/main.yml
liangwen12year: https://github.com/linux-system-roles/network/pull/675#discussion_r1497609081, as suggested by Fernando, I want to check if any network package can be updated, and this is the reason that `update_cache` is used here. 

The package module can not check if a list of packages can be updated.

liangwen12year: I updated the commit message.

tasks/main.yml
liangwen12year: I explained in the commit message.

tests/tasks/create_wireless_profile_restart_network.yml
liangwen12year: I just want to show the error message when the following task fails rather than stopping running the playbook : 

```
- name: Ask user's consent to restart NetworkManager due to wireless or team
    interfaces
  fail:
    msg: NetworkManager needs to be restarted to be able to proceed
      because wireless and team interfaces are defined. This might
      disturb the connectivity of the managed system. Please set
      `network_allow_restart` to `true` if you are prepared for this.
      Notice that the necessary action is to install NetworkManager-wifi or
      NetworkManager-team plugin and to restart NetworkManager.
  register: __network_service_restart_requested
  when:
    - __network_wireless_connections_defined
       or __network_team_connections_defined
    - network_provider == "nm"
    - not network_allow_restart
    - dnf_package_update_info is changed or yum_package_update_info is changed
    ```

Comment (richm) [reply tests/tasks/create_wireless_profile_restart_network.yml:21]: ok - then what you have is fine

**Comment by liangwen12year** [comment]
[citest]

**Comment by liangwen12year** [comment]
[citest]

**Comment by liangwen12year** [comment]
> The PR description is missing the text for Enhancement/Reason/Result.

I added it.

**Comment by liangwen12year** [comment]
[citest]

File: tasks/main.yml [hunk 16]
```diff
@@ -24,6 +24,51 @@
...
+  register: dnf_package_update_info
+  check_mode: true
+  when:
+    - ansible_distribution == 'Fedora' or
+      ansible_distribution_major_version | int > 7
+    - __network_wireless_connections_defined
+      or __network_team_connections_defined
```
richm, line 40:
```suggestion
      or __network_team_connections_defined
    - not __network_is_ostree
```

tasks/main.yml
```diff
@@ -24,6 +24,51 @@
...
+  yum:
+    update_cache: true
+    name: "{{ network_packages }}"  # noqa package-latest
+    state: latest
+  register: yum_package_update_info
+  check_mode: true
+  when:
+    - ansible_distribution_major_version | int < 8
+    - __network_wireless_connections_defined
+      or __network_team_connections_defined
```
richm, line 53:
```suggestion
      or __network_team_connections_defined
    - not __network_is_ostree
```

tasks/main.yml
liangwen12year: It is fine to add it, but I think that update cache itself will only change the `/var/cache/dnf` directory but not the `/usr` directory.

tasks/main.yml
```diff
@@ -24,6 +24,53 @@
   when:
     - network_state is defined
     - ansible_distribution_major_version | int < 8
+
+- name: Check if updates for network packages are available through the DNF
+    package manager due to wireless or team interfaces
+  dnf:
```
tyll, line 30: Please explain in the commit message why we have both dnf and yum tasks.

tests/ensure_provider_tests.py
```diff
@@ -105,6 +105,9 @@
     "playbooks/tests_team.yml": {
         EXTRA_RUN_CONDITION: "ansible_distribution != 'Fedora'",
     },
+    "playbooks/tests_wireless_and_network_restart.yml": {
+        EXTRA_RUN_CONDITION: "ansible_distribution != 'Fedora'",
```
tyll, line 109: What's the reason for this?

tests/playbooks/tests_wireless_and_network_restart.yml
```diff
@@ -0,0 +1,79 @@
...
+  tasks:
+    - name: Show playbook name
+      debug:
+        msg: "this is: playbooks/tests_wireless_and_network_restart.yml"
+      tags:
+        - always
+
+    - name: Check if rpm ostree system - cannot test
+      meta: end_host
+      when: __network_is_ostree | d(false)
```
tyll, line 28: What's the reason this cannot be tested on an rpm ostree system?

tests/playbooks/tests_wireless_and_network_restart.yml
```diff
@@ -0,0 +1,79 @@
...
+            lsr_setup:
+              - tasks/delete_interface.yml
+              - tasks/assert_device_absent.yml
+              - tasks/remove_and_assert_package_absent.yml
+            lsr_test:
+              - tasks/create_wireless_profile_restart_network.yml
+            lsr_assert:
+              - tasks/assert_network_connections_failed.yml
+            lsr_cleanup:
+              - tasks/check_network_dns.yml
```
tyll, line 51: :+1:

tests/tasks/remove_and_assert_package_absent.yml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: "Get the rpm package facts"
+  package_facts:
+    manager: "auto"
+
+- name: "Assert removed package '{{ package }}'"
+  assert:
+    that:
+      - package not in ansible_facts.packages
+    msg: "'{{ package }}' is not removed"
```
tyll, line 16: IMHO it is fine to trust that the package tasks will remove the package and the assert is not necessary.

tests/playbooks/tests_wireless_and_network_restart.yml
liangwen12year: https://github.com/linux-system-roles/network/blob/main/README-ostree.md#rpm-ostree, " the /usr filesystem is read-only, and the role cannot install packages".

richm: The problem is that we don't have a good way currently to pass in the custom yum repos/copr repos that are used by these tests.  If you could build an image with all of the repos and packages, then I think the tests should work.

tests/ensure_provider_tests.py
liangwen12year: Nice catch.

tests/playbooks/tests_wireless_and_network_restart.yml
tyll: > [`main`/README-ostree.md#rpm-ostree](https://github.com/linux-system-roles/network/blob/main/README-ostree.md?rgh-link-date=2024-03-13T13%3A32%3A38Z#rpm-ostree), " the /usr filesystem is read-only, and the role cannot install packages".

This is not a sufficient explanation. Because this still leaves the question what's the reason that this cannot be tested? The role needs to have a defined behavior when running against ostree systems.



> The problem is that we don't have a good way currently to pass in the custom yum repos/copr repos that are used by these tests. If you could build an image with all of the repos and packages, then I think the tests should work.

Maybe I am missing something. If an ostree system is missing the wifi or team subpackage (or even some NM or Nmstate package), then the role needs to handle this with a proper error message (if we intend to support ostree targets). If the role handles this correctly.

tasks/main.yml
liangwen12year: I explained it.

tests/playbooks/tests_wireless_and_network_restart.yml
richm: The problem is that the test wants to remove the package `NetworkManager-wifi` then reinstall it later - packages cannot be removed from an ostree system - the only thing package management can do at runtime is to verify that the given package exists (or is absent).

If the test needs to verify the package `NetworkManager-wifi` is absent, then call the role to install it, then verify that the package `NetworkManager-wifi` is present, then at least this part of the test is not applicable for ostree systems.  If we really want this test to be run on ostree systems, then we need to modify the test such that the testing of the package absence and presence is skipped on ostree systems.  I don't know if that will substantially change what this test is doing.

tyll: From what I understand so far, for the ostree case, if the package is installed, the test should be skipped. If the package is missing, the test should assert that the role produces a descriptive error message.

richm: > From what I understand so far, for the ostree case, if the package is installed, the test should be skipped.

Then the test should always be skipped in our testing frameworks, because we always build the image with all of the packages needed for every runtime and testing configuration: https://github.com/linux-system-roles/network/blob/main/.ostree/packages-runtime.txt

> If the package is missing, the test should assert that the role produces a descriptive error message.

The package should never be missing in our ostree testing.

tyll: Which testing frameworks remain to test package installation for the role? Which systems would still run the test?

richm: > Which testing frameworks remain to test package installation for the role?

Our CI system, local developer testing via `tox -e qemu...`, and all of our downstream testing.

> Which systems would still run the test?

All of the above.

That is - we are still running all tests, in a few different ways, with "regular" systems that can install and remove packages at runtime.

tyll: I do not follow, previously you wrote "Then the test should always be skipped in our testing frameworks," - when exactly will it be skipped and when not? Just considering automated CI.

richm: > I do not follow, previously you wrote "Then the test should always be skipped in our testing frameworks," - when exactly will it be skipped and when not? Just considering automated CI.

We currently do not have automated CI for ostree.

Because of the way we build images for ostree testing, the `NetworkManager-wifi` package will always be installed in the image.

If a customer builds an ostree image, then runs the network role against it, and they specify some parameter that requires an additional package that is not built into the ostree image, then the `package:` task will print a message like "The package XXX is not found".

tyll: ok, good enough.

tyll: What's missing is to explain this in the code.

tests/ensure_provider_tests.py
```diff
@@ -101,10 +101,7 @@
 blackhole, prohibit and unreachable",
     },
     "playbooks/tests_routing_rules.yml": {},
-    # team interface is not supported on Fedora
```
tyll, line 104: Please don't add more commits to this PR that are not needed to fix the Jira issue. It only makes the review more complicated. This can be merged just fine in its own PR.

tests/playbooks/tests_wireless_and_network_restart.yml
richm: > What's missing is to explain this in the code.

How?  As a code comment in tasks/main.yml for the various `package:` tasks?

tyll: I was thinking of changing `name: Check if rpm ostree system - cannot test` to something like `name: Cannot test ostree system because ... `

tyll: If there are multiple places, maybe it is also worth to write this in one file and include it in all places. Doing this in a follow-up PR is fine.

tests/ensure_provider_tests.py
liangwen12year: I moved the commit to its own PR.

liangwen12year:
[citest]

liangwen12year:
[citest]

tests/playbooks/tests_wireless_and_network_restart.yml
liangwen12year: tracked by https://github.com/linux-system-roles/network/issues/681

tyll: I meant doing this at other places and/or using a task file for all those places is something to do in a different PR. The task in this PR needs a good name, still.

liangwen12year:
@tyll , if you agree, then I will proceed to merge this PR.

tests/playbooks/tests_wireless_and_network_restart.yml
```diff
@@ -0,0 +1,80 @@
...
+      - __network_connections_result
+  tasks:
+    - name: Show playbook name
+      debug:
+        msg: "this is: playbooks/tests_wireless_and_network_restart.yml"
+      tags:
+        - always
+
+    - name: Cannot test ostree systems because the package can not be removed
+        or installed - 'NetworkManager-wifi'
```
liangwen12year, line 27: If we are using the jinja2 syntax for the '{{ package }}', the {{ package }}' should be appeared at the end of the task name.

tests/playbooks/tests_wireless_and_network_restart.yml
liangwen12year: I renamed the task into following.

```
    - name: Cannot test ostree systems because the package can not be removed
        or installed - 'NetworkManager-wifi'
```

liangwen12year:
[citest]

tests/playbooks/tests_wireless_and_network_restart.yml
tyll: FWIW, `cannot` is more common than `can not`.

tyll (approved):
Thank you, LGTM

tests/playbooks/tests_wireless_and_network_restart.yml
liangwen12year: Right, thanks for pointing it out, I will address it in another PR.

richm (approved):

PR #676, opened 2024-02-21
test: improve bond test failure debugging

richm:
Improve bond test failure debugging
* put test setup into block/always so that cleanup happens for setup failures
* trace shell commands so that we can determine exactly which command failed
* add error reporting so that when certain commands fail, we can determine the error code

### DIFF
--- a/tests/playbooks/tests_bond.yml
+++ b/tests/playbooks/tests_bond.yml
@@ -10,26 +10,26 @@
     port2_profile: bond0.1
     dhcp_interface2: test2
   tasks:
-    - name: "INIT Prepare setup"
-      debug:
-        msg: "##################################################"
-    - name: Import the task 'create_test_interfaces_with_dhcp.yml'
-      import_tasks: tasks/create_test_interfaces_with_dhcp.yml
-    - name: Import the task 'assert_device_present.yml'
-      import_tasks: tasks/assert_device_present.yml
-      vars:
-        interface: "{{ dhcp_interface1 }}"
-    - name: Import the task 'assert_device_present.yml'
-      import_tasks: tasks/assert_device_present.yml
-      vars:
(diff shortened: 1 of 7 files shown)


richm:
[citest]

richm:
@liangwen12year before this patch, I could reproduce bond test errors about 50% of the time.  With this patch, I cannot reproduce the bond test errors at all.  Note that my patch doesn't actually fix anything, it just adds better debugging.  Perhaps the bond test issue is timing related, and the additional debugging makes the problem go away? Anyway, if we see the problem again, it will be easier to debug.

liangwen12year:
> @liangwen12year before this patch, I could reproduce bond test errors about 50% of the time. With this patch, I cannot reproduce the bond test errors at all. Note that my patch doesn't actually fix anything, it just adds better debugging. Perhaps the bond test issue is timing related, and the additional debugging makes the problem go away? Anyway, if we see the problem again, it will be easier to debug.

But I do not understand how additional debugging would change the timing problem in this case. It is true that, it would be easier to debug next time.

liangwen12year (approved):

richm:
> > @liangwen12year before this patch, I could reproduce bond test errors about 50% of the time. With this patch, I cannot reproduce the bond test errors at all. Note that my patch doesn't actually fix anything, it just adds better debugging. Perhaps the bond test issue is timing related, and the additional debugging makes the problem go away? Anyway, if we see the problem again, it will be easier to debug.
> 
> But I do not understand how additional debugging would change the timing problem in this case.

I haven't measured it, but the additional checking and debug output takes time, and perhaps enough time elapses that the network is in the correct state.  Or maybe it is one of those "Heisenberg" issues that the act of measuring alters the outcome.

> It is true that, it would be easier to debug next time.

+1

richm:
Can this be merged?  Any objections?

PR #684, opened 2024-03-17
test: improve name text for skipped ostree tests

richm:
Improve the name text for skipped ostree tests to explain
why the test is skipped.

### DIFF
--- a/tests/playbooks/tests_team_plugin_installation.yml
+++ b/tests/playbooks/tests_team_plugin_installation.yml
@@ -3,9 +3,10 @@
 - name: Play for testing team plugin installation
   hosts: all
   tasks:
-    - name: Check if rpm ostree system - cannot test
-      meta: end_host
-      when: __network_is_ostree | d(false)
+    - name: Check if test can run on ostree systems
+      include_tasks: tasks/ostree_systems_check.yml
+      vars:
+        package: NetworkManager-team
 
     - name: Remove the NetworkManager-team package
       package:
--- a/tests/playbooks/tests_wireless_plugin_installation.yml
+++ b/tests/playbooks/tests_wireless_plugin_installation.yml
@@ -3,9 +3,10 @@
 - name: Play for testing wireless plugin installation
   hosts: all
   tasks:
-    - name: Check if rpm ostree system - cannot test
-      meta: end_host
-      when: __network_is_ostree | d(false)
+    - name: Check if test can run on ostree systems
+      include_tasks: tasks/ostree_systems_check.yml
+      vars:
+        package: NetworkManager-wifi
 
     - name: Remove the NetworkManager-wifi package
       package:
--- a/tests/tests_wireless_and_network_restart_nm.yml
+++ b/tests/tests_wireless_and_network_restart_nm.yml
@@ -21,4 +21,3 @@
   import_playbook: playbooks/tests_wireless_and_network_restart.yml
   when:
     - ansible_distribution_major_version != '6'
-    - ansible_distribution != 'Fedora'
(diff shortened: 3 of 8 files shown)


tests/playbooks/tests_team_plugin_installation.yml
```diff
@@ -3,7 +3,11 @@
 - name: Play for testing team plugin installation
   hosts: all
   tasks:
-    - name: Check if rpm ostree system - cannot test
+    - name: >-
+        This is an rpm-ostree system.  This test cannot be run on such a system.
```
spetrosi, line 7: There are some double spaces in the text.

spetrosi: Wow, never heard of it. I am not against you can keep it if so. Otherwise, lgtm.

tests/playbooks/tests_wireless_plugin_installation.yml
```diff
@@ -3,7 +3,11 @@
 - name: Play for testing wireless plugin installation
   hosts: all
   tasks:
-    - name: Check if rpm ostree system - cannot test
+    - name: >-
+        This is an rpm-ostree system.  This test cannot be run on such a system.
+        The test wants to test removing and reinstalling the
+        NetworkManager-wifi package.  Package removal is not allowed on
+        rpm-ostree systems.
```
liangwen12year, line 10: maybe we can include the task file here with `var: {{package}}`, then in the task file itself, we can write it as :

```
    - name: Cannot test ostree systems because the package can not be removed
        or installed - '{{ package}}'
      meta: end_host
      when: __network_is_ostree | d(false) 
  ```

File: tests/tests_wireless_and_network_restart_nm.yml [hunk 3]
```diff
@@ -21,4 +21,3 @@
   import_playbook: playbooks/tests_wireless_and_network_restart.yml
   when:
     - ansible_distribution_major_version != '6'
-    - ansible_distribution != 'Fedora'
```
liangwen12year, line 24: Nice catch!!

tests/playbooks/tests_wireless_plugin_installation.yml
richm: > maybe we can include the task file here

which task file?  I'm not sure what you mean.

> with `var: {{package}}`, then in the task file itself, we can write it as :
> 
> ```
>     - name: Cannot test ostree systems because the package can not be removed
>         or installed - '{{ package}}'
>       meta: end_host
>       when: __network_is_ostree | d(false) 
> ```

liangwen12year: Put the following task into a separate task file, for example, `tasks/ostree_systems_check.yml`
```
    - name: Cannot test ostree systems because the package can not be removed
        or installed - '{{ package}}'
      meta: end_host
      when: __network_is_ostree | d(false) 
```

Then in the playbook (tests/playbooks/*.yml), we can include the task file as: 
```
    - name: Include the task 'ostree_systems_check.yml'
      include_tasks: tasks/ostree_systems_check.yml
```

liangwen12year: If using the task file `tasks/ostree_systems_check.yml`, we can also include the task file in  `tests_wireless_and_network_restart.yml`.

liangwen12year:
Can you also use the `include_tasks: ostree_systems_check.yml` in tests_wireless_and_network_restart.yml ?

liangwen12year:
[citest]

richm:
[citest]

richm:
fedora 39 test hung so I killed it - every test passed except the one that hung - the hang is unrelated to this change

PR #702, opened 2024-07-03
fix: network_state must be defined in defaults/main.yml

richm:
As part of the public API, `network_state` must be defined in
defaults/main.yml, and it must be defined with the correct
type `dict`, so the correct default value must be `{}` the
empty dict.

All checking for `network_state` must check for a value of
`{}` to mean "network_state not set or empty".

Fix the test which looks for teaming configuration in EL10
to correctly look for the value in `network_state`.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,6 +2,8 @@
 ---
 network_connections: []
 
+network_state: {}
+
 network_allow_restart: false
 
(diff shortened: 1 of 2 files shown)


liangwen12year:
[citest]

richm:
[citest]

liangwen12year (approved):

PR #704, opened 2024-07-24
test: fix some Ansible warnings not caught by lint

richm:
Do not use templating in `when:`, `that:`, `until:`.  These
are evaluated as Jinja statements.  In cases where the string
used is long or awkward to generate in-line, use an
intermediate var for the value.

Use a unique loop var instead of `item` in cases where a loop
may be called in a nested context.

Fix some formatting.


richm:
[citest]

richm:
[citest]

spetrosi (dismissed):

liangwen12year (approved):

richm:
[citest]

PR #706, opened 2024-07-24
test: team plugin test does not clean up properly

richm:
The team plugin test does not clean up properly causing the following team
test to fail.  The fix is to use the network standard run_test.yml interface
to ensure proper preconditions and cleanup for the team plugin test.


richm:
[citest]

PR #529, opened 2022-10-05
added optional route src parameter

h3po:
works with nm, untested with initscript
TODO: test with initscript
TODO: unit tests

Signed-off-by: h3po <mail@h3po.de>

PR #714, opened 2024-08-04
feat: Add the support for the optional route source parameter in nm provider

liangwen12year:
Enhancement: Add the optional route source parameter for the nm provider

Reason: In a scenario where you have a machine with multiple public IP
addresses, typically due to a multi-WAN setup, the src parameter in the
context of routes allows you to specify which source IP address should
be used when sending packets via a specific route.  This is crucial when
you want to ensure that outbound traffic uses a specific IP address tied
to a particular network interface, especially when dealing with multiple
WAN connections.

Result: Adding support for the src parameter in routes results in a
more powerful and flexible network configuration capability, especially
important in environments with multiple network interfaces or multiple
IP addresses, it provides better control over traffic routing.

Resolves: https://issues.redhat.com/browse/RHEL-3252

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,10 +596,11 @@ The IP configuration supports the following options:
 
   Static route configuration can be specified via a list of routes given in the
   `route` option. The default value is an empty list. Each route is a dictionary with
-  the following entries: `gateway`, `metric`, `network`, `prefix`, `table` and `type`.
-  `network` and `prefix` specify the destination network. `table` supports both the
-  numeric table and named table. In order to specify the named table, the users have to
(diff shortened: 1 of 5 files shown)


liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
This PR takes part of the code from https://github.com/linux-system-roles/network/pull/529.

liangwen12year:
[citest]

cathay4t (approved):

richm (approved):

PR #720, opened 2024-08-14
test: skip integration pytest on fedora 39 and later

richm:
Something has changed in python, similar to the change between
el8 and el9, that causes the test to fail on f39 and later, so
skip it.


richm:
[citest]

richm:
[citest]

PR #721, opened 2024-08-15
ci: fix and improve integration container testing

richm:
Use the vault for centos 7 and centos stream 8

Exclude tests/tests_team_plugin_installation_nm.yml since
tests/tests_team_nm.yml is excluded.

Use grouping to group log lines for better readability.
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines


richm:
I have to merge this to trigger a rebuild of the containers

PR #724, opened 2024-08-17
test: allow dhcp service if firewall is active

richm:
If firewall is active, the dhcp services do not work.  Change the test
to add and remove the dhcp services if firewall is active.


richm:
[citest]

PR #726, opened 2024-08-20
test: use is-active instead of is-enabled to check for firewalld

richm:
Need to test if firewalld is running - the is-active test is
for that, not is-enabled


richm:
[citest]

PR #731, opened 2024-08-27
docs: Remove invalid network state example

liangwen12year:
For an ethernet device which contains the kernel link, we should not and cannot delete such a device using `network_state` variable.

### DIFF
--- a/examples/network_state_example.yml
+++ b/examples/network_state_example.yml
@@ -19,16 +19,6 @@
       ansible.builtin.include_role:
         name: linux-system-roles.network
 
-    - name: Delete the ethernet device eth1
-      vars:
-        network_state:
-          interfaces:
-            - name: eth1
-              type: ethernet
-              state: absent
-      ansible.builtin.include_role:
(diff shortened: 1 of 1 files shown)


richm (approved):

spetrosi (approved):

examples/network_state_example.yml
```diff
@@ -19,16 +19,6 @@
...
-    - name: Delete the ethernet device eth1
-      vars:
-        network_state:
-          interfaces:
-            - name: eth1
-              type: ethernet
-              state: absent
-      ansible.builtin.include_role:
-        name: linux-system-roles.network
-
```
liangwen12year, line 31: @tyll , it is worth explaining...

For state apply, nmstate unifies the veth and ethernet into `type: ethernet`, the veth type is only meaningful for querying the state.  We can only use nmstate to delete the virtual NIC that is created by NM/Nmstate, we can not use nmstate to delete the veth device that is created by iproute2.

PR #751, opened 2025-01-18
tests: Prefer permanent MAC for finding link info, fallback to current MAC

liangwen12year:
Add the integration test `tests_mac_address_match.yml` to verify that the commit "Prioritize find link info by permanent MAC address, with fallback to current address" (c3416831) can properly resolve the scenarios where multiple network interfaces share the same current MAC address, leading to potential ambiguity in link matching (for example, Virtual interfaces or VLANs, which often lack a valid perm-address and rely on the parent interface's address).

Notice that the test `tests_mac_address_match.yml` will be skipped in upstream testing and it will only be manually tested downstream since it requires to know the mac address to match when configuring vlan's parent.

Enhancement:

Reason:

Result:

Resolves: https://issues.redhat.com/browse/RHEL-74211


tests/playbooks/tests_mac_address_match.yml
```diff
@@ -0,0 +1,43 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Play for testing MAC address match on device
+  hosts: all
+  vars_prompt:
+    - name: "interface"
+      prompt: "Please provide the interface name (e.g., eno2)"
+      private: no
```
richm, line 8:
```suggestion
      private: false
```

tests/playbooks/tests_mac_address_match.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: "mac"
+      prompt: "Please provide the MAC address to match (e.g., ec:f4:BB:d3:EC:92)"
+      private: no
```
richm, line 11:
```suggestion
      private: false
```

tests/playbooks/tests_mac_address_match.yml
```diff
@@ -0,0 +1,43 @@
...
+  vars:
+    profile: "{{ interface }}"
+    vlan_profile: "{{ interface }}.3732"
+    lsr_fail_debug:
+      - __network_connections_result
+  tags:
+    - "tests::mac"
```
richm, line 18: @spetrosi I think we'll need to exclude this test from automated testing - we'll need to add `tests::mac` to the `--skip-tags` list  @liangwen12year I think this test is meant to be run interactively

spetrosi: Created a PR for that https://github.com/linux-system-roles/tft-tests/pull/65

tyll: I don't follow why the test cannot determine the MAC address itself instead of asking for it in a prompt.

tests/playbooks/tests_mac_address_match.yml
```diff
@@ -0,0 +1,43 @@
...
+        - tests::mac:match
+      block:
+        - name: Include the task 'run_test.yml'
+          include_tasks: tasks/run_test.yml
+          vars:
+            lsr_description: Test MAC address match on device
+            lsr_setup:
+              - tasks/assert_profile_absent.yml
+            lsr_test:
+              - tasks/create_mac_address_match.yml
```
tyll, line 37: Does this file already exist? I don't see it in the PR.

liangwen12year: Thanks for pointing it out, I did not add the file in the commit. Now I added it back.

tests/playbooks/tests_mac_address_match.yml
liangwen12year: The test is supposed to operate on an interface which has the permanent address, I do not make the test accidentally configured on the interface that is used to bridge the ssh connection between the control node and managed host.  Possibly, I can search through all the links in `/sys/class/net/` and use the first unconfigured ethernet interface. But such an ethernet interface (unconfigured and also have a permanent mac address) may not exist in the system environment of upstream testing.

liangwen12year: You are right. We can use `ethtool -P $iface` to determine the mac address of the interface.

tyll: We controll the upstream test envirionment, so it should be clear whether or not such an interface exists and if it does not, there might be an option to create such an interface.

Nevertheless, the prompt also does not specify the requirements for the interface to ensure that it is eligible/the test does not ensure that it actually tests what it is supposed to test.

liangwen12year: Honestly, it turns out that the upstream testing environment does not have the interface with the permanent mac address.  I also tried `iproute2` , `systemd-networkd`, and NetworkManager, but I can not create such an interface with the permanent mac address (the permanent mac address can be queried by calling `ethtool -P $iface` ), I can only create an interface with the current mac address (the current mac address can be queried by calling `cat /sys/class/net/$iface/address` ).

liangwen12year: I specified the requirements for the interface to ensure that it is eligible.

tyll: How does the upstream environment work?

tests/playbooks/tests_mac_address_match.yml
```diff
@@ -0,0 +1,65 @@
+# SPDX-License-Identifier: BSD-3-Clause
+---
+- name: Play for testing MAC address match on device
+  hosts: all
+  vars_prompt:
+    - name: "interface"
+      prompt: "The test requires matching an Ethernet interface to its
```
tyll, line 7: From what I understand, the current MAC address of this interface needs to be different than the permanent MAC address for the new code to actually be tested. And after further discussion, there should probably be a second interface (maybe a virtual one) that has the same MAC address as the permanent MAC address as this was causing an error condition. @liangwen12year told me about this in a meeting, AFAIU, this is not mentioned in the original PR.

liangwen12year: The original issue https://issues.redhat.com/browse/RHEL-73442 describes the scenario that when providing the interface name and mac address at the same time in the network connections for configuring the parent and the vlan connection, and applying the same network connections multiple times, the network role will raise an error - “profile specifies interface_name 'eno2' and mac 'ec:f4:bb:d3:ec:92' but no such interface exists”. The reason that the error was raised is that `SysUtil._link_infos_fetch()` will create a dictionary containing all the available link infos in the `/sys/class/net/`, each dictionary element has the structure - {“ifname”: {“ifindex”, “ifname”, “address”, “perm-address”}}.  Then, in the `SysUtil.link_info_find()` function, the code iterates through all the link info in the aforementioned dictionary, and tries to compare if the mac address that the user specified in the network connections appeared either as the “perm-address” or “address” (current address) in the link info, return such link info if appeared. However, if the mac address that the user specified matches only the “address” but not the “perm-address”, the function will still return. In this case, the returned link info will contain a different “ifname” than the interface name that the user specified in the network connections, as a result, the error is raised. If the physical interface has the permanent mac address A and current address B, and the user specifies the mac address A in the network connections, when applying the same network connections to configure the parent and the vlan multiple times, the error is not raised, because the vlan interface will have the same current mac address as the parent’s current mac address, but the vlan’s perm-address will not be set, therefore, the vlan’s link info is never returned for comparing the interface name in the link info against the interface name that user specified. 

From my understanding, the current mac address of this interface should be the same as the permanent mac, then the created vlan interface will have the same current mac address as the permanent mac address of its parent, then without the prior fix (https://github.com/linux-system-roles/network/commit/c34168318c3982bb5b88793dd84c5f95c5be6f13) on the issue, I can reproduce the problem by applying the same network connections multiple times for configuring the parent and the vlan connections.

tests/playbooks/tests_mac_address_match.yml
liangwen12year: Summarized from the slack discussion, the upstream testing uses Testing Farm, which uses AWS machines, we do not run kvm/qemu virtual machines on the AWS machines. According to the doc https://tmt.readthedocs.io/en/stable/spec/hardware.html#network, requesting a machine with extra NICs should be already supported via `network.type`, we should be able to provision a machine with 2 NICs by specifying the following in the network system role test configuration. 

```
hardware:
  network:
    - type: eth
    - type: eth
```

library/network_connections.py
```diff
@@ -173,7 +173,7 @@ def _link_read_permaddress(ifname):
     @staticmethod
     def _link_infos_fetch():
         links = {}
-        for ifname in os.listdir("/sys/class/net/"):
+        for ifname in sorted(os.listdir("/sys/class/net/")):
```
richm, line 176: `sorted` will sort in alphanum order, but the order of the parent and virtual interfaces here is not guaranteed to be in that order - so how does sorting help in this case?

library/network_connections.py
```diff
@@ -222,31 +222,29 @@ def link_infos(cls, refresh=False):
         return linkinfos
 
     @classmethod
-    def link_info_find(cls, refresh=False, mac=None, ifname=None):
-        if mac is not None:
+    def link_info_find(cls, mac=None, ifname=None):
+        if mac:
```
richm, line 226: note that `""`, `[]`, and `{}` all evaluate to `False` in a boolean context - so if `mac == ""` the `if mac:` will evaluate to `False` - not sure if that is allowed

library/network_connections.py
liangwen12year: Yes, the list will be sorted in alphanum order, the  VLAN's device name can be arbitrary and independent of parent's device name or the VLAN id. 

Honestly, we had a hard time reproducing the issue because of the arbitrarily ordered list returned by `os.listdir("/sys/class/net/")`, sorting the list can make the behavior consistent and more predictable given the VLAN's device name. For example, we can have two sets of tests, the VLAN device's link info is added before the parent's link info to the link infos dictionary in one test, in the other test, the VLAN device's link info is added after the parent's link info to the link infos dictionary. By following this testing plan, we can validate how effective the patch https://github.com/linux-system-roles/network/commit/c34168318c3982bb5b88793dd84c5f95c5be6f13 addresses the problem. 

It is beneficial for the user too, the user will consistently reproduce the problem given a fixed VLAN device name (when the VLAN device's link info is added before the parent's link info to the link infos dictionary) before the patch https://github.com/linux-system-roles/network/commit/c34168318c3982bb5b88793dd84c5f95c5be6f13. If some users run the network role module code directly instead of running the network role installed from the ansible-galaxy, an ordered link infos list will have deterministic behavior or easily nail isolate the root cause when link info related error happens.

library/network_connections.py
liangwen12year: `mac` should be in string format, the `ArgValidatorMac` already rules out the possibility of the empty string in this case.

library/network_connections.py
richm: good idea

tyll, line 176: Please reduce the commit message to the essence of why  this change is useful.

library/network_connections.py
```diff
@@ -222,31 +222,29 @@ def link_infos(cls, refresh=False):
         return linkinfos
 
     @classmethod
-    def link_info_find(cls, refresh=False, mac=None, ifname=None):
```
tyll, line 225: The first line of the commit message is unnecessarily long, please make it shorter.

library/network_connections.py
```diff
@@ -222,31 +222,29 @@ def link_infos(cls, refresh=False):
...
-        for linkinfo in cls.link_infos(refresh).values():
-            perm_address = linkinfo.get("perm-address", None)
-            current_address = linkinfo.get("address", None)
 
-            # Match by perm-address (prioritized)
-            if mac is not None and perm_address not in [None, "00:00:00:00:00:00"]:
-                if mac == perm_address:
-                    return linkinfo
+        result = None
+        null_mac = "00:00:00:00:00:00"
```
tyll, line 230: please make this a constant (all caps), define it outside the method.

library/network_connections.py
```diff
@@ -222,31 +222,29 @@ def link_infos(cls, refresh=False):
...
-                return linkinfo
+            if ifname and ifname == linkinfo["ifname"]:
+                result = linkinfo
+                break
 
-        # Return fallback match by address if no perm-address match found
-        if "matched_by_address" in locals():
-            return matched_by_address
+            if mac:
+                if perm_address != null_mac and mac == perm_address:
```
tyll, line 241: What's the reason to check that the permanent mac address is not null here?

liangwen12year: null mac is an indication of the absence of the mac address or uninitialized mac address,  checking the permanent mac address is not null here can guarantee that the permanent mac address is valid and initialized.
As alluded in https://standards-support.ieee.org/hc/en-us/articles/4888705676564-Guidelines-for-Use-of-Extended-Unique-Identifier-EUI-Organizationally-Unique-Identifier-OUI-and-Company-ID-CID,  `EUI-48 and EUI-64 identifiers ( Extended Unique Identifier) are most commonly used as globally unique network addresses (sometimes called MAC addresses)`,  and `00-00-00-00-00-00 shall not be used as identifiers`.

library/network_connections.py
liangwen12year: I rewrote it as `refactor: Make SysUtil.link_info_find() more pythonic`.  Thanks.

library/network_connections.py
liangwen12year: I rephrased it to capture the essence of the reason for the change.

library/network_connections.py
liangwen12year: The function `get_perm_addr()` effectively converts the unset/uninitialized MAC into `00:00:00:00:00:00` 
https://github.com/linux-system-roles/network/blob/main/module_utils/network_lsr/ethtool.py#L39.

library/network_connections.py
liangwen12year: I defined it as a constant outside the method. Thanks.

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
Close this PR as the patches in this PR have been split into multiple separate PRs.

https://github.com/linux-system-roles/network/pull/768
https://github.com/linux-system-roles/network/pull/769

PR #762, opened 2025-02-06
ci: Add an additional NIC for test purposes

spetrosi:


.github/workflows/tft.yml
```diff
@@ -180,7 +180,7 @@ jobs:
           tf_scope: private
           api_key: ${{ secrets.TF_API_KEY_RH }}
           update_pull_request_status: false
-          tmt_hardware: '{ "memory": ">= ${{ needs.prepare_vars.outputs.memory }} MB" }'
+          tmt_hardware: '{ "memory": ">= ${{ needs.prepare_vars.outputs.memory }} MB", "network": [{"type": "eth"},{"type": "eth"}] }'
```
liangwen12year, line 183: what is the NIC name or how can we tell from the config  ?

.github/workflows/tft.yml
```diff
@@ -180,7 +180,7 @@ jobs:
           tf_scope: private
           api_key: ${{ secrets.TF_API_KEY_RH }}
           update_pull_request_status: false
-          tmt_hardware: '{ "memory": ">= ${{ needs.prepare_vars.outputs.memory }} MB" }'
+          tmt_hardware: '{ "memory": ">= ${{ needs.prepare_vars.outputs.memory }} MB", "network": [{"type": "eth", "device-name": "eth0"}, {"type": "eth", "device-name": "eth1"}] }'
```
richm, line 183: @liangwen12year does this work?

liangwen12year, line 183: The commit message is a bit confusing, it is not adding two NICs , you only added 1 additional NIC, `eth0` is still the default device for bridging the ssh connection between the control node and the managed host. Can you change the commit message to  `ci: Configure two NICs for test purposes` ?

spetrosi: We've tested it with Wen in TF and it seems to be working. Will see how TF action likes this syntax.

liangwen12year: I pair-programmed with @spetrosi on a testing farm machine using the following command, and it works as expected.  

```
ip a
nmcli c
nmcli d
ethtool -P eth0
cat  /sys/class/net/eth0/address
ethtool -P eth1
cat  /sys/class/net/eth1/address
```

liangwen12year (approved):

PR #763, opened 2025-02-10
ci: Add test plan in .ci dir to be able to customize it per each role

spetrosi:


spetrosi:
@liangwen12year @richm I realized that TF does not support providing hardware requirements with `tmt_hardware` when using multi-host. So the solution would be adding a test plan within roles' repos, like I do in this PR. That way we can specify reqs in the plan directly and it works. 
device-name property is not supported (https://gitlab.com/testing-farm/artemis/-/blob/main/server/src/tft/artemis/schema/environment-v0.0.74.yml?ref_type=heads#L296), tmt documentation is wrong about it. But I once again checked that device names are `eth0`, `eth1`.

liangwen12year:
Please sign off the commit.

.ci/.fmf/version
```diff
@@ -0,0 +1 @@
+1
```
liangwen12year, line 1: what is the purpose of this file ? is this required to add ?

spetrosi: Yes, it's is required so that tmt reads this dir. This file exists in the parent dir too, but tmt doesn't read the plan if it is stored in a hidden file. And I want it to be stored as a hidden file so that we do not need to remove it manually when building RPMs (we remove all hidden files automatically.

spetrosi: i.e. creating `./.general.fmf` doesn't work. As ell as creating `./.github/general.fmf`

.ci/general.fmf
```diff
@@ -0,0 +1,108 @@
+summary: A general test for a system role
+tag: general
+provision:
+  - name: control-node1
+    role: control_node
+    # TF uses `how: artemis`, tmt try uses `how: virtual`. No need to define `how`
+    # `connection: system` is for `how: virtual` to make VMs get a real IP to configure ssh easily
+    # This setting is ignored on artemis so we can keep it
+    connection: system
+  - name: managed-node1
```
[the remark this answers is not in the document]
richm: We configure multiple hosts so that we can run multiple tests at the same time.

.ci/general.fmf
```diff
@@ -0,0 +1,108 @@
+summary: A general test for a system role
+tag: general
+provision:
+  - name: control-node1
+    role: control_node
+    # TF uses `how: artemis`, tmt try uses `how: virtual`. No need to define `how`
+    # `connection: system` is for `how: virtual` to make VMs get a real IP to configure ssh easily
+    # This setting is ignored on artemis so we can keep it
```
liangwen12year, line 8: Honestly, I can not understand the comment here, can you rephrase it so that it would be easier for the people without much knowledge in TF or tmt to understand it? Thanks.

liangwen12year: You are missing the period after "No need to define `how`".

how about 
```
# TF uses `how: artemis`, while `tmt try` uses `how: virtual`, so there's no need to define `how`.
# `connection: system` is required for `how: virtual` to assign VMs a real IP, making SSH configuration easier.
# This setting is ignored in `artemis`, so we can leave it as is.
```

liangwen12year (commented):
Can you also add the reason of the change to the commit message? Thanks!

.ci/.fmf/version
spetrosi: Ok I am storing it in the `plans` directory to be consistent with what tmt usually does.
Also, changed the name to be more descriptive.

.ci/general.fmf
spetrosi: Thanks, done

plans/test_playbooks_parallel.fmf
```diff
@@ -0,0 +1,110 @@
...
+    connection: system
+    hardware:
+      network:
+        - type: eth
+        - type: eth
+environment:
+  ANSIBLE_VER: 2.17
+  REPO_NAME: network
+  PYTHON_VERSION: 3.12
+  SYSTEM_ROLES_ONLY_TESTS: "tests_default.yml"
```
richm, line 58: @spetrosi is this just for testing this PR?

spetrosi: A leftover from my local tests, removed it. Testing a PR is not possible unfortunately.

richm (approved):

liangwen12year (approved):

richm:
The last question is - does this work?  If so, I think it is ready to merge.

PR #768, opened 2025-03-12
fix: Refine MAC validation using interface name

liangwen12year:
When a user provides both an interface name and a MAC address, the current validation process retrieves sysfs link info separately using the interface name and the MAC address, then compares the results. If the information doesn't match, an error is raised.

However, this approach may trigger false alarms because retrieving the link info by MAC might return data that only matches the current MAC instead of the permanent MAC. Since the interface name is unique within the kernel, a more robust validation method is to fetch the MAC address using the interface name and then compare it directly with the user-provided MAC address.

Cause:  When a user provides both an interface name and a MAC address, the current validation process retrieves sysfs link info separately using the interface name and the MAC address, then compares the results. If the information doesn't match, an error is raised.

Consequence:  However, this approach may trigger false alarms because retrieving the link info by MAC might return the link info that only matches the current MAC instead of the permanent MAC.

Fix:  Since the interface name is unique within the kernel, a more robust validation method is to fetch the MAC address using the interface name and then compare it directly with the user-provided MAC address.

Result:  With the fix, the role will never trigger false alarms when validating the user-provided MAC address and the interface name.

Resolves: https://issues.redhat.com/browse/RHEL-84362


liangwen12year:
[citest]

liangwen12year:
[citest]

library/network_connections.py
```diff
@@ -222,31 +244,29 @@ def link_infos(cls, refresh=False):
         return linkinfos
 
     @classmethod
-    def link_info_find(cls, refresh=False, mac=None, ifname=None):
-        if mac is not None:
+    def link_info_find(cls, mac=None, ifname=None):
```
cathay4t, line 247: You should remove searching link by mac address because that will cause ambiguity. This function should accept interface name only and all user of this function should adopt to it. Otherwise, you will always raise false alarm somewhere.

cathay4t (changes requested):

library/network_connections.py
liangwen12year: I adopted your suggestion and updated the code, thanks!

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

richm:
@liangwen12year I can create the jira issue from this PR - if this is a fix, please change the description to the
```
Cause: 

Consequence: 

Fix: 

Result: 
```
format

@xjezda00 please review for automated test coverage - it lgtm - also - for the check files database - I think adding some line from the files tests/playbooks/tests_bond_port_match_by_mac.yml and tests/playbooks/tests_mac_address_match.yml will be sufficient - wdyt?

liangwen12year:
> @liangwen12year I can create the jira issue from this PR - if this is a fix, please change the description to the
> 
I linked the issue for the PR and changed the description to the format as you suggested.

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
> @liangwen12year is this ready to be merged?

@cathay4t , if no objection from you, can we merge it?

cathay4t (approved):

PR #770, opened 2025-03-13
test: do not need to install from epel or pip

richm:
The tests should not install anything from outside of the distribution
unless absolutely necessary, like the copr repos.

All of the EPEL dependencies have been removed or replaced
with coprs.

We do not need to install pytest from pip since it is available
as `pytest-3` from `python3-pytest`.

We do not need `git` or `rsync` in the tests.


richm:
[citest]

PR #774, opened 2025-04-02
fix: Correct attribute checks for routing rule validation

liangwen12year:
The validation was incorrectly checking for routing rule attributes at the top-level NM module instead of the NM.IPRoutingRule class. This was causing validation failures because:

libnm's API has two core aspects:
1. NMConnection/NMSetting types for handling connection profiles
2. NMClient as a cache of D-Bus objects

The suppress_prefixlength and uid_range attributes are not part of the top-level NM module but belong to NM.IPRoutingRule. Updated the validation to properly check for:
- set_suppress_prefixlength instead of NM_IP_ROUTING_RULE_ATTR_SUPPRESS_PREFIXLENGTH
- set_uid_range instead of NM_IP_ROUTING_RULE_ATTR_UID_RANGE_START

This aligns with the correct API usage and fixes the validation errors.

Enhancement:

Reason:

Result:

Resolves: https://issues.redhat.com/browse/RHEL-85872

### DIFF
--- a/tests/playbooks/tests_routing_rules.yml
+++ b/tests/playbooks/tests_routing_rules.yml
@@ -28,89 +28,134 @@
         mode: "0644"
         create: true
 
-    - name: Configure connection profile and specify the numeric table in
-        static routes
-      import_role:
-        name: linux-system-roles.network
-      vars:
-        network_connections:
-          - name: "{{ interface }}"
-            interface_name: "{{ interface }}"
-            state: up
-            type: ethernet
-            autoconnect: true
-            ip:
-              dhcp4: false
-              address:
-                - 198.51.100.3/26
-                - 2001:db8::2/32
-              route:
-                - network: 198.51.100.64
-                  prefix: 26
-                  gateway: 198.51.100.6
-                  metric: 4
-                  table: 30200
-                - network: 198.51.100.128
-                  prefix: 26
-                  gateway: 198.51.100.1
-                  metric: 2
-                  table: 30400
-                - network: 2001:db8::4
-                  prefix: 32
-                  gateway: 2001:db8::1
-                  metric: 2
-                  table: 30600
-              routing_rule:
-                - priority: 30200
-                  from: 198.51.100.58/26
-                  table: 30200
-                - priority: 30201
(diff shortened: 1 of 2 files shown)


richm:
lgtm - but I'm not familiar with the nm api

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
[citest]

liangwen12year:
Thanks to @richm 's help, https://github.com/liangwen12year/network/pull/3.

tests/playbooks/tests_routing_rules.yml
```diff
@@ -255,18 +316,17 @@
...
   hosts: all
   tasks:
+    - name: Include the task 'remove+down_profile.yml'
+      include_tasks: tasks/remove+down_profile.yml
+      vars:
+        profile: "{{ interface }}"
+      when: not __network_end_play | d(false)
+    - name: Include the task 'delete_interface.yml'
+      include_tasks: tasks/delete_interface.yml
+      when: __network_end_play| d(false)
```
richm, line 329:
```suggestion
```
I think we want to delete the interface every time

tests/playbooks/tests_routing_rules.yml
```diff
@@ -28,89 +28,134 @@
...
+              failure is not raised when the distro's major version is 7
+          when: ansible_distribution_major_version == "7"
+
+        - name: Clear errors
+          meta: clear_host_errors
+
+        - name: Reset __network_end_play variable
+          set_fact:
+            __network_end_play: true
+
```
richm, line 155: need to delete the routing table file here or it won't be removed (at least, if it is, I can't tell where it is removed)  path: /etc/iproute2/rt_tables.d/table.conf

liangwen12year: I reorganized the route table cleanup task, so that it is guaranteed to run.

liangwen12year:
[citest]

richm (approved):

liangwen12year:
[citest_bad]

PR #775, opened 2025-04-08
ci: Drop Python 3.8 from python-unit-test.yml, add Python 3.12

martinpitt:
Tested in https://github.com/linux-system-roles/sudo/pull/47 , and I'll submit it to .github.git in a minute, but I first want a second project to succeed here before we generalize it.


martinpitt:
Uh, some unrelated TODOs here (ubuntu-20.04 *cough*), but the unit tests are happy :heavy_check_mark:

PR #779, opened 2025-04-14
test: find second interface to use for mac address match

richm:
Some systems do not use the `ethN` interface naming scheme, and
use `ensN` instead.  The test wants to use `eth1` as the second
interface.  If this does not exist, try `ens4` instead.


richm:
fedora 42 tests are hanging on tests_mac_address_match - can reproduce locally - debugging

richm:
Will fix one of the f42 issues in a separate commit.

PR #794, opened 2025-06-26
refactor: support Ansible 2.19

richm:
The big problem was trying to use `vars` with `import_playbook`.

We do not need to use `import_playbook` when `include_tasks` will
work.  Perhaps the original author of these tests thought that
the play `roles` keyword was the only way to invoke roles, so
that had to be "called" using an `import_playbook`?

Use `include_tasks` instead of `import_playbook`, and move some
of those "tasks" playbooks to be tasks files in tests/tasks.

Use `include_role` instead of `import_role`.

Do not set variables using `set_fact` if they have already been
set at the appropriate scope using `vars`.

"Modernize" the code somewhat.

Improve formatting.

Work around an Ansible bug https://github.com/ansible/ansible/issues/85394

## Summary by Sourcery

Refactor the test suite to support Ansible 2.19 by replacing deprecated import_* directives with include_tasks/include_role, removing redundant set_fact usage, reorganizing task files, and filtering expected plugin warnings to work around Ansible issue #85394.

Bug Fixes:
- Adjust assertion logic to filter out known plugin-load warnings, working around Ansible bug #85394

Enhancements:
- Replace import_playbook calls with include_tasks throughout test playbooks
- Replace import_role usage with include_role in all tests
- Remove unnecessary set_fact statements and consolidate variable definitions at the play or role level
- Move standalone playbook fragments into dedicated files under tests/tasks and update references accordingly

Tests:
- Revise test playbooks to leverage include_tasks/include_role patterns and updated task structure


richm:
[citest]

PR #801, opened 2025-08-28
test: ensure NetworkManager, ensure eth1 is active

richm:
This is mainly needed on el7 - NetworkManager is installed by default, or
somewhere else, on el8 and later.

Additionally, if the NetworkManager-server-config package is installed, then
the secondary interface will not be active, so ensure it is active.

## Summary by Sourcery

Ensure NetworkManager is installed, running, and at a sufficient version for the MAC address match tests and update test configurations accordingly

Tests:
- Install NetworkManager, gather its version, and start its service in the MAC address match test playbook
- Add a condition to require NetworkManager version >= 1.18.0 before running the MAC address match test
- Update the minimum supported NetworkManager version to 1.18.0 in the provider tests configuration


richm:
[citest]

richm:
[citest]

richm:
[citest]

PR #804, opened 2025-09-03
fix: allow use of built-in routing tables

richm:
Cause: The user is trying to specify the routing table to use by the name of
a built-in routing table defined in /usr/share/iproute2/rt_tables such as `main`.

Consequence: The network role gives an error:
"cannot find route table main in `/etc/iproute2/rt_tables` or `/etc/iproute2/rt_tables.d/`"
The workaround is that the user must specify the table by number instead of name e.g
`table: 254` instead of `table: main`

Fix: Look for table mappings in /usr/share/iproute2/rt_tables as well as the other
paths.

Result: The user can use built-in route table names.

## Summary by Sourcery

Enable specifying built-in route tables by name by loading mappings from the system rt_tables file and validate the behavior with new tests

Bug Fixes:
- Allow using built-in routing tables by name by parsing /usr/share/iproute2/rt_tables

Enhancements:
- Extend route table mapping lookup order to include system-defined tables before user-defined ones

Tests:
- Add tests to verify route addition and retrieval for built-in and custom routing tables

### DIFF
--- a/module_utils/network_lsr/argument_validator.py
+++ b/module_utils/network_lsr/argument_validator.py
@@ -2771,6 +2771,11 @@ class IPRouteUtils(object):
     def get_route_tables_mapping(cls):
         if not hasattr(cls, "_cached_rt_tables"):
             mapping = {}
+            # look in static built-in route tables file first
+            cls._parse_route_tables_mapping_from_file(
+                "/usr/share/iproute2/rt_tables", mapping
+            )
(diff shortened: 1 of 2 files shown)


richm:
https://issues.redhat.com/browse/RHEL-110865

richm:
[citest]

richm:
[citest_bad]

richm:
[citest]

PR #861, opened 2026-04-27
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.

## Summary by Sourcery

Add syslog fingerprinting for the network system role and verify it via tests.

New Features:
- Introduce an sr_fingerprint Ansible module to write timestamped fingerprint messages to syslog for diagnostic use.
- Emit begin and success fingerprint messages from the network role including Ansible version and distribution information.

Tests:
- Extend the default role test playbook to capture a journal start time and assert that the expected begin and success fingerprint messages are written to the system journal.


richm:
[citest]

PR #865, opened 2026-04-30
test: ensure integration pytest works in collection mode

richm:
The tests/ directory has a library sub-directory with a symlink to the
network_connections.py module.  This was causing problems because the logic
in get_modules_and_utils_paths.yml really wants the real files and
directories, not the symlinks, and for some reason the bash `-f` test
is returning `true` for the symlink.  Instead, ensure that we do not use
the link by using `-L` to screen it out.

Also added the collection paths to the search.

Make the path arrays have unique values.

Use the `realpath` filter to see the full actual path.

## Summary by Sourcery

Adjust test collection path discovery to ignore symlinked modules, include additional collection plugin paths, and normalize discovered paths for integration pytest.

Bug Fixes:
- Prevent symlinked network_connections.py in the tests library directory from being incorrectly treated as the real module during path discovery.

Enhancements:
- Deduplicate module and module_utils search paths when constructing collection path lists.
- Extend module and module_utils search paths to include collection plugin directories under plugins/modules and plugins/module_utils.
- Normalize the resolved tests directory path in the integration pytest playbook using a realpath filter.


richm:
[citest]

richm:
[citest]

PR #868, opened 2026-05-07
feat: Use verbosity-based logging in network role

spetrosi:
This PR replaces hardcoded `no_log: true` with verbosity-based logging `no_log: "{{ ansible_verbosity < 3 }}"` throughout the network role.

## Changes

- Replace `no_log: true` with `no_log: "{{ ansible_verbosity < 3 }}"` in:
  - `tasks/set_facts.yml`: setup, service_facts, package_facts tasks
  - `tasks/main.yml`: NetworkManager and initscripts service tasks

## Rationale

In the network role, `no_log: true` was used for verbosity control to reduce verbose output from service and package fact gathering tasks, not to protect sensitive information. 

Using `ansible_verbosity < 3` allows users to see the output when debugging with increased verbosity flags (`-vvv` or higher), while keeping output concise by default.

This differs from roles where `no_log` is used for security purposes (e.g., protecting credentials or secrets), where a `*_secure_logging` parameter would be appropriate.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -157,7 +157,7 @@
     enabled: true
   when:
     - network_provider == "nm" or network_state != {}
-  no_log: true
+  no_log: "{{ ansible_verbosity < 3 }}"
 
 # If any 802.1x connections are used, the wpa_supplicant
 # service is required to be running
@@ -176,7 +176,7 @@
     enabled: true
   when:
     - network_provider == "initscripts"
-  no_log: true
+  no_log: "{{ ansible_verbosity < 3 }}"
 
 - name: Ensure initscripts network file dependency is present
   copy:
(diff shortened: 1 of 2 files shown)


tasks/main.yml
```diff
@@ -157,7 +157,7 @@
     enabled: true
   when:
     - network_provider == "nm" or network_state != {}
-  no_log: true
+  no_log: "{{ network_secure_logging }}"
```
richm:
```suggestion
  no_log: "{{ ansible_verbosity < 2 }}"
```
Adding `no_log: true` was just for verbosity, not security - same with other places

so I don't think we need `network_secure_logging` for now

richm (approved):

richm:
@spetrosi you'll need to sign your commit for the `DCO` check e.g. `git commit -s --amend`

richm (approved):

PR #880, opened 2026-06-25
test: fix kernel url path for the mac80211_hwsim kernel module

richm:
The method for determining the kernel url path for the module did not
work on all arches.

Skip the test if the module is not found.  Some architectures do not provide
the module.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Wi‑Fi test setup on CentOS/Red Hat by installing `mac80211_hwsim` dependencies with more robust kernel/version handling.
  * Added stricter validation of the kernel release string to prevent incorrect package URL construction.
  * The setup now stops early if `mac80211_hwsim` fails to install, avoiding misleading test outcomes.
* **Tests**
  * Adjusted the installation order for NetworkManager and related Wi‑Fi services (`wpa_supplicant`, `hostapd`) to follow the latest test flow.

### DIFF
--- a/tests/tasks/setup_mock_wifi.yml
+++ b/tests/tasks/setup_mock_wifi.yml
@@ -9,31 +9,6 @@
     __mock_wifi_server_cert_dir: "{{ '/etc/pki/tls/hostapd_test'
       if __mock_wifi_mode == '802_1x' else '/etc/pki/tls' }}"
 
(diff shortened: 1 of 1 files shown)


richm:
[citest]

richm:
[citest]

PR #912, opened 2026-09-08
fix: reject network_state dns-resolver when NM global-dns is configured

pfeifferj:
Enhancement: 

The network_state module checks NetworkManager's configuration files (NetworkManager.conf and the conf.d directories) for a [global-dns] or [global-dns-domain-*] section before applying a state that contains dns-resolver, and fails with an error naming the file and the workarounds.

Reason: 

nmstate stores and purges dns-resolver through NetworkManager's D-Bus global DNS API. NetworkManager refuses that call while global DNS comes from a configuration file, and nmstate surfaces the refusal as an "Global DNS configuration already set via configuration file" error, which the role forwarded as a module traceback. Neither NetworkManager nor nmstate plan to change this, so the role reports the configuration as unsupported, as agreed on in RHEL-4450.

Result: 

Users with a [global-dns] section get an actionable error telling them to remove the section and reload NetworkManager, or to configure DNS on the connection profiles instead of dns-resolver. Global DNS set over D-Bus (what nmstate itself writes) lives in [.intern.*] groups and does not trigger the check.

RHEL-4450, RHEL-3257, NMT-1833

Signed-off-by: Josephine Pfeiffer josie@redhat.com

## Summary by CodeRabbit

* **Bug Fixes**
  * `dns-resolver` configurations now fail with a clear explanation when NetworkManager global DNS settings are present, instead of producing an internal error.
  * DNS resolver configuration continues to apply when no conflicting global DNS settings are detected.

* **Documentation**
  * Added guidance on the limitation, failure behavior, and available workarounds.

* **Tests**
  * Added coverage for global DNS detection, configuration precedence, unsupported scenarios, and successful DNS resolver configuration.


spetrosi:
[citest_all]

spetrosi:
[citest_all]

library/network_state.py
```diff
@@ -55,6 +58,33 @@
     NETWORK_HAS_NMSTATE = True
     NETWORK_NMSTATE_IMPORT_ERROR = None
 
+# NetworkManager.conf(5) load order: a later dir shadows a same-named file.
+# D-Bus-set global DNS lives in [.intern.*] groups and must not match.
+NM_CONFIG_FILE = "/etc/NetworkManager/NetworkManager.conf"
```
richm, line 63: @pfeifferj is there some way we can get the values for the config file and dirs from `NM`?

library/network_state.py
```diff
@@ -55,6 +58,33 @@
...
+]
+NM_GLOBAL_DNS_SECTION_RE = re.compile(b"^\\s*\\[global-dns(-domain-[^\\]]*)?\\]")
+
+
+def find_nm_global_dns_config():
+    """Return the NetworkManager config files that define [global-dns*] sections."""
+    snippets = {}
+    for config_dir in NM_CONFIG_DIRS:
+        for path in glob.glob(os.path.join(config_dir, "*.conf")):
+            snippets[os.path.basename(path)] = path
```
richm, line 77: What if you have `/usr/lib/NetworkManager/conf.d/50-conf.conf` which has a `[global-dns-something]` section and `/etc/NetworkManager/conf.d/50-conf.conf` which does not have a `[global-dns-something]` section?  From this code, it looks as though the second loop below will only look at `/etc/NetworkManager/conf.d/50-conf.conf` and will miss the global dns section in  `/usr/lib/NetworkManager/conf.d/50-conf.conf`

richm: maybe that's fine - maybe the way that NM/nmstate works is that if you have both `/usr/lib/NetworkManager/conf.d/50-conf.conf` and `/etc/NetworkManager/conf.d/50-conf.conf`, the former file will be completely ignored.

library/network_state.py
pfeifferj: NM doesn't export config paths, and GlobalDnsConfiguration on d-bus has no origin flag. the paths I defined are the NetworkManager.conf(5) defaults

library/network_state.py
pfeifferj: yea `/usr/lib` being skipped when `/etc` or `/run` has the same name is documented in NetworkManager.conf(5)

richm (approved):

htetarkarpmn-design (approved):

